PR #26, opened 2024-10-01
PyPI upload workflow uses thrusted publisher instead of PyPI API token.

rafa-be:
Uses trusted PyPI publisher credentials instead of PyPI API token when releasing the package.

@JamieSlome Please go to the ["Publishing" tab from PyPI dashboard for Scaler](https://pypi.org/manage/project/scaler/settings/publishing/) and add Github as a trusted publisher, with the following values:

- Owner: Citi
- Repository name: scaler
- Workflow name: pypi.yml
- Environment name: release

I couldn't test it, as I didn't have access to the PyPI package anymore.


JamieSlome (approved):
LGTM! 👍

PR #31, opened 2024-10-10
Adds a GitHub Action that publishes the Scaler docs.

rafa-be:


docs/requirements_docs.txt
```diff
@@ -1,4 +1,3 @@
 sphinx
-sphinx-jsonschema
```
rafa-be, line 2: I had to remove this package, as it's under GPL 2.0, which is not compatible with Apache 2, even as an imported library.

In addition, it does not seem to be required for generating the documentation.

JamieSlome: Cool, if not required, we should remove dependencies 👍

JamieSlome:
@rafa-be - can we look into why the unit tests are taking so long to run? Looks like a runaway issue?

rafa-be:
> @rafa-be - can we look into why the unit tests are taking so long to run? Looks like a runaway issue?

Not sure what happened, I don't see any error.

I restarted it.

JamieSlome:
Thanks @rafa-be - we might want to investigate though as we've had this issue before where the tests just runaway and keep running on 👍 Not major, has occurred twice since we pushed out the repo, maybe 3 times max.

sharpener6 (approved):

sharpener6 (approved):

PR #34, opened 2024-10-12
Get rid of copy memory when receive and send message

sharpener6:
Hopefully this will use zero copy, but it's rely on pyZMQ's implementation, if this still doesn't solve copy problem (should use zero copy), we will need implement c/c++ version of client agent + scheduler + worker agent.


rafa-be (dismissed):
LGTM. Have you noticed any improvement in performances?

rafa-be (dismissed):

rafa-be (approved):

PR #38, opened 2024-10-24
Fix serialized big data issue

sharpener6:
- now single object can be more than 500MB because of cap'n protocol limitation, according to author, it should be List(List(Data)) then concatenate them should support the bytes more than 500MB, but pycapnp is not supporting memoryview now, so when splitting and concatenation, it will create multiple copy for sure, but it should be done right after serialization and deserialization stage and release the memory,

- later when we moved from pyzmq to pynng, the memory copy issue should get easied because pynng will expose more raw operations, and we can save threads


scaler/io/utility.py
```diff
@@ -20,3 +20,15 @@ def deserialize(data: bytes) -> Optional[Message]:
...
+def chunk_to_list_of_bytes(data: bytes) -> List[bytes]:
+    # TODO: change to list of memoryview when capnp can support memoryview
+    return [data[i : i + CAPNP_DATA_SIZE_LIMIT] for i in range(0, len(data), CAPNP_DATA_SIZE_LIMIT)]
+
+
+def concat_list_of_bytes(data: List[bytes]) -> bytes:
+    one_object_bytes = bytearray()
+    for chunk in data:
+        one_object_bytes.extend(chunk)
+    return one_object_bytes
```
rafa-be, line 34: Looks like it could be implemented with `b"".join([..])`:

```python
In [4]: b"".join(b"hello" for i in range(0, 4))
Out[4]: b'hellohellohellohello'
```

rafa-be (approved):

PR #46, opened 2024-11-20
Allows client to discard object data

rafa-be:
Allows the client to discard object data without disconnecting.

```python
with Client("tcp://127.0.0.1:60338") as client:
    client.submit(math.sqrt, 16)
    client.clear()
```


scaler/client/agent/object_manager.py
```diff
@@ -48,6 +69,13 @@ async def __send_object_creation(self, instruction: ObjectInstruction):
         if not new_object_ids:
             return
 
+        if b"serializer" in instruction.object_content.object_names:
```
sharpener6, line 72: @rafa-be I think we should put this `b"serializer"` into a capnproto constant, and all worker, client should use it

scaler/about.py
```diff
@@ -1 +1 @@
-__version__ = "1.8.14"
+__version__ = "1.8.16"
```
sharpener6, line 1: Unfortunately, this need be bumped to 1.8.17

sharpener6 (approved):

Issue #45, opened 2024-11-20
Scheduler incorrectly returns a task not found error


rafa-be:
When a client immediately cancels a task it just submitted, the scheduler will sometime return a `TaskStatus.NotFound` message, and log a error message such as:

```python
[EROR]2024-11-20 22:47:58+0100: cannot find task_id=1ad3d9caa78911ef99be965046454255 in task workers
```

## How to reproduce

The bug **will systematically occur if the scheduler does not have any worker registered**. It can also occur with registered workers, but with a lower probability.

```python
with Client("tcp://127.0.0.1:60338") as client:
    client.submit(math.sqrt, 16).cancel()
```

## Cause

The bug is in the implementation of the scheduler's task manager routine:

https://github.com/Citi/scaler/blob/f4b4daaa9aac218f710cb03ea7fd09fdd67660d7/scaler/scheduler/task_manager.py#L55-L67

The scheduler maintains two collections for tasks, `_unassigned` and `_running`.

As the `assign_task_to_worker()` might be blocking (if there is no worker), or yield to another async task, **there is a possibility that the task being scheduled is removed from `_unassigned` but not yet added to `_running`**. A `TaskCancel` message will thus raise a task not found error as the task is in neither of these collections.

## Possible fix

Adding the task to `_running` before assigning it to a worker does not work, as it breaks task cancellation: the scheduler does not know to which worker to route the `TaskCancel` message.

Instead, the scheduler could in this order:

1. acquire an available worker;
2. wait on the task queue for the next task;
3. immediately add the task to `_running`;
4. send the `Task` object to the acquired worker.

This requires some refactoring of the scheduler.

rafa-be:
The problem is due to the scheduler waiting on two blocking queues: the task queue and the worker queue.

@sharpener6 As I said, I'd like to make the worker queue non-blocking while implementing task tagging (#32). This will remove this bug, and will simplify task scheduling.

This requires a small behavior change: when no worker is connected to the scheduler, the client will receive a `TaskStatus.NoWorker` error, while currently the task is queued until a worker connects.

sharpener6:
This will be address and fix in https://github.com/finos/opengris-scaler/pull/174

Issue #32, opened 2024-10-10
Support Tagging when scheduling the worker load


sharpener6:
- Worker should send supported tag when they connected
- Client will send task along with tags
- Scheduler will only schedule tasks to workers that supports tags

rafa-be:
Tags makes balancing tasks slightly more complex, as these add constraints on how tasks can be balanced from one worker to another.

**My favored way to implement tags would be in a way similar to Dask's resources, by allowing workers to announce support for multiple string-based tags (e.g. "high-memory", "linux" or "GPU"). Tasks would then be declared with a set of tags that are required to run them**.  Tags would thus represent features provided by workers, and required by tasks.

A worker would be allowed to run a task if `task.tags.issubset(worker.tags)`.

**Balancing tasks like these is more complex than without tags**.

For example, consider this cluster's state:

- Worker 1 
    - Supported tags: {Linux, GPU}
    - Tasks: 
        - Task 1: {Linux}

- Worker 2 
    - Supported tags: {Linux}
    - Tasks: None

- Worker 3:
    - Supported tags: {GPU}
    - Tasks:
        - Task 2: {GPU}
        - Task 3: {GPU}
          
Here the optimal way of balancing the cluster would be to move Task 1 to Worker 2 and then Task 3 to Worker 1.

By adding tags, balancing cannot be done only considering two individual workers (Worker 3 and Worker 2 in the example). The whole cluster (including Worker 1) must be considered in the balancing computation.

Balancing algorithms exist that can find this optimal balancing (see [assignment problem](https://en.wikipedia.org/wiki/Assignment_problem)), but these are complex and slow. These might also cause a lot of messages to be propagated through the cluster.

Instead, I propose we use a simpler but less optimal algorithm:

1. We select the worker with the highest number of tasks;
2. We try to remove one task from this worker by moving it to a worker that has less than the average number of tasks;
    - If we cannot find a compatible worker, we stop considering this worker in step 1.
3. Repeat step 1. until either:
    - All workers reached the average number of tasks;
    - We cannot find a worker to move task from.

Worst-case for this algorithm seems to be *O(|workers| • |tasks|)*, while the average should be closer to *O(|tasks|)*.
If no tag is used in the cluster, worst-case will always be *O(|tasks|)*, like the current balancing algorithm.

[...]

PR #47, opened 2024-12-19
Issue #32: Initial implementation of the non-blocking tagging task allocator.

rafa-be:
Adds a first implementation of a task allocator that supports tags. Scheduler is still using the old `QueuedAllocator`.

Once this replaces the old `QueuedAllocator`, it will also fix issue #45, as it removes the blocking worker queue.


requirements.txt
```diff
@@ -4,4 +4,5 @@ graphlib-backport; python_version < '3.9'
 psutil
 pycapnp
 pyzmq
+sortedcontainers
```
rafa-be, line 7: This does not really add a new dependency, as pycapnp already has it as a dependency.

scaler/protocol/capnp/message.capnp
```diff
@@ -6,9 +6,10 @@ using Status = import "status.capnp";
 struct Task {
     taskId @0 :Data;
     source @1 :Data;
-    metadata @2 :Data;
-    funcObjectId @3 :Data;
-    functionArgs @4 :List(Argument);
+    tags @2 :List(Text);
```
rafa-be, line 9: Tags are using capnp's text strings. Maybe we would want to use UTF-8 encoded bytes instead.

sharpener6, line 9: add `tags` at last as `@5`

scaler/protocol/capnp/message.capnp
```diff
@@ -58,12 +59,13 @@ struct ClientHeartbeatEcho {
...
-    latencyUS @3 :UInt32;
-    taskLock @4 :Bool;
-    processors @5 :List(Status.ProcessorStatus);
+    tags @0 :List(Text);
+    agent @1 :Status.Resource;
+    rssFree @2 :UInt64;
+    queuedTasks @3 :UInt32;
+    latencyUS @4 :UInt32;
+    taskLock @5 :Bool;
+    processors @6 :List(Status.ProcessorStatus);
```
sharpener6, line 68: no need shifting fields, put them to the last

tests/test_tagged_allocator.py
```diff
@@ -0,0 +1,147 @@
...
+
+from typing import Dict, Set
+
+from scaler.utility.logging.utility import setup_logger
+from scaler.scheduler.allocators.tagged_allocator import TaggedAllocator
+from scaler.protocol.python.message import Task
+
+from tests.utility import logging_test_name
+
+MAX_TASKS_PER_WORKER = 10
```
sharpener6, line 12: move this to default value file

PR #50, opened 2025-01-17
Add Symphony worker implementation

1597463007:
# Preliminary Symphony worker implementation

## Motivation

IBM Spectrum Symphony is a grid compute framework that allows for large-scale CPU-intensive workloads. We want to adapt Scaler to use Symphony as if it is a normal Scaler worker.

## Usage

```bash
python run_symphony_worker.py tcp://127.0.0.1:8000 SymphonyService
```

A Symphony service needs to be deployed beforehand, here is an example service that accepts the message payload sent from Scaler:

```python
class Message(soamapi.Message):
    def __init__(self, payload: bytes = b""):
        self.__payload = payload

    def set_payload(self, payload: bytes):
        self.__payload = payload

    def get_payload(self) -> bytes:
        return self.__payload

    def on_serialize(self, stream):
        payload_array = array.array("b", self.get_payload())
        stream.write_byte_array(payload_array, 0, len(payload_array))

    def on_deserialize(self, stream):
        self.set_payload(stream.read_byte_array("b"))

class ServiceContainer(soamapi.ServiceContainer):
    def on_create_service(self, service_context):
        return

    def on_session_enter(self, session_context):
        return

    def on_invoke(self, task_context):
        input_message = Message()
        task_context.populate_task_input(input_message)

        fn, *args = cloudpickle.loads(input_message.get_payload())
        output_payload = cloudpickle.dumps(fn(*args))

        output_message = Message(output_payload)
        task_context.set_task_output(output_message)

    def on_session_leave(self):
        return

    def on_destroy_service(self):
        return
```

##  Rough edges

* Object cache is implemented to not linger objects after a task is submitted which may impact performance
* Processor suspend is not possible in Symphony hence it is not implemented
* Task cancelling is not implemented and cancelled tasks are simply marked as cancelled in the task manager for bookkeeping
* Profiling is not implemented


scaler/entry_points/symphony_worker.py
```diff
@@ -0,0 +1,106 @@
...
+    DEFAULT_IO_THREADS,
+    DEFAULT_NUMBER_OF_WORKER,
+    DEFAULT_WORKER_DEATH_TIMEOUT,
+)
+from scaler.utility.event_loop import EventLoopType, register_event_loop
+from scaler.utility.logging.utility import setup_logger
+from scaler.utility.zmq_config import ZMQConfig
+from scaler.worker.symphony.worker import SymphonyWorker
+
+
```
bcncpp, line 18: Nitpick: i wonder if here it might be nicer having a typer app.

scaler/worker/symphony/task_manager.py
```diff
@@ -0,0 +1,301 @@
...
+        """
+        soamapi.initialize()
+
+        self._session_callback = SessionCallback()
+
+        self._ibm_soam_connection = soamapi.connect(
+            self._service_name, soamapi.DefaultSecurityCallback("Guest", "Guest")
+        )
+        logging.info(f"established IBM Spectrum Symphony connection {self._ibm_soam_connection.get_id()}")
+
```
bcncpp, line 78: Nitpick:  IBM SOAM context can be delegated to an external object / context making less coupled this class, i.e. SOAM bulider.

scaler/worker/symphony/task_manager.py
```diff
@@ -0,0 +1,301 @@
...
+            if task_id in self._acquiring_task_ids:
+                self._acquiring_task_ids.remove(task_id)
+                self._executor_semaphore.release()
+
+            self._task_id_to_task.pop(task_id)
+
+    async def process_task(self):
+        await self._executor_semaphore.acquire()
+
+        _, task_id = await self._queued_task_id_queue.get()
```
bcncpp, line 233: is there any possibility here to have None value?

rafa-be: Please correct me if I'm wrong, but the call should be blocking until there is an item to consume in the queue (this is a blocking queue, not a list/deque/other).

scaler/worker/symphony/task_manager.py
```diff
@@ -0,0 +1,301 @@
...
+            serializer = self._serializers[serializer_id]
+
+            object_ids = (task.func_object_id, *(arg.data for arg in task.function_args))
+            for object_id in object_ids:
+                lock = asyncio.Lock()
+                await lock.acquire()
+                self._object_cache_lock[object_id] = lock
+
+            await self._connector_external.send(ObjectRequest.new_msg(ObjectRequest.ObjectRequestType.Get, object_ids))
+            await asyncio.gather(*[self._object_cache_lock[object_id].acquire() for object_id in object_ids])
```
rafa-be, line 267: Could it be made slightly faster by sending a single object request for the serializer and the arguments?

rafa-be (commented):
Looks great to me, although I've 2 remarks:

1) Please consider adding a documentation/README entry on how to use this Symphony implementation;

2) Looks like a task with an high-priority could "bypass" the semaphore mechanism. If this happens, could it cause the number of concurrent tasks to exceed the `max_concurrency` value?

scaler/about.py
rafa-be, line 1: Please skip to 1.8.16 as @sharpener6 requested.

scaler/entry_points/symphony_worker.py
1597463007: Good suggestion, I will look into it

scaler/worker/symphony/task_manager.py
1597463007: It is blocking, so None shouldn't occur

scaler/worker/symphony/task_manager.py
1597463007: I can batch the two object request calls

1597463007:
1. I will update the readme
2. max_concurrency is probably a bad name. It should be more accurately named base_concurrency as the semaphore can be bypassed for higher priority tasks. The behavior should be well documented because the Symphony scheduler can lock up due to bypassing.

rafa-be (approved):

rafa-be (approved):

sharpener6 (approved):

Issue #52, opened 2025-01-29
Make queue size configurable by the worker


1597463007:
Currently queue size is configured by the scheduler is applied globally across all connected workers. This was deemed not a problem up until the introduction of the IBM Spectrum Symphony worker which features workers with alternative implementations. Some workers running on higher specced hosts or workers that can execute multiple tasks concurrently should be allowed to accept more tasks into its queue.

This issue will track the progress of moving the queue size configuration into the worker side.

The worker will send the queue size info using the heartbeat payload and the existing `--per-worker-queue-size` scheduler flag will become a noop and will be removed in the future.

gxuu:
I've rolled a basic solution according to the requirements outlined in issue #52. I welcome any comments and advice to further improve the program.

This pull request doesn't fully comply with the guidelines. Here's a list of the missing items:

- I didn't increment the version number. While this change breaks compatibility, it's not a significant feature. It might be best to address this when rolling out major updates.
- I didn't write tests. The implementation passed all existing tests, and I've also manually tested the code. I'm willing and eager to write tests after receiving feedback.
- I didn't craft the code, and the naming was bad.

I'm also confused about the following:

- It seems "servers" are organized into "Clusters," and each cluster consists of several workers. Since all workers within a cluster run on the same machine, why not specify the queue size at the cluster level?

Thanks,
gxu

1597463007:
I agree the terminology can be confusing. In Dask, Dask workers is analogous to Scaler clusters as each Dask worker can hold more than one process to execute tasks.

In Scaler's context, "Cluster" means a group of workers running under the same parent PID. It's mainly used to make clean up the workers easier.

The term has diverged a bit from the original meaning as more worker implementations are created. E.g. The IBM Spectrum Symphony Worker is for all intents and purposes a "Worker" but it behaves more like a "Cluster" as communicates with the Symphony Grid Scheduler (and by extension Symphony Grid Workers) and can run tasks concurrently.

The cluster level doesn't have the ability to handle messages, messages are sent directly between the Scheduler and Workers.

gxuu:
I have made commits since last round of discussion. I think that pull request is ready to merge. I welcome review.

gxu

PR #53, opened 2025-01-31
Basic Implementation Making Queue Size Configurable by Workers

gxuu:
I've rolled a basic solution according to the requirements outlined in issue #52. I welcome any comments and advice to further improve the program.


gxuu:
More detail in the discussion of issue #51. This is not ready to merge and I plan to make improvement after receiving feedback.

scaler/scheduler/allocators/queued.py
1597463007, line 1: Since the queue size is now configurable, the balancing implementation should also be updated

The old balancing algorithm ensures all workers have the same amount of tasks queued. The new balancing algorithm should be based on queue occupancy ratio. Since this change ties queue size to the relative computing performance of each worker, the change should be communicated inside the README but that can come later.

Example:

worker 1 (100 queue size)
worker 2 (100 queue size)
worker 3 (1000 queue size)

If there are 600 tasks:

Old algorithm (all workers should have the same amount of tasks):

worker 1: 200 tasks
worker 2: 200 tasks
worker 3: 200 tasks

New algorithm (since 600/1200=50%, each worker should have 50% of the queue size occupied by tasks):

worker 1: 50 tasks
worker 2: 50 tasks
worker 3: 500 tasks

gxuu: I forgot to ask this but yes I was wondering how `balance` in `queued.py` should be implemented. 

Initially I was thinking about the ratio method you proposed, and worried whether this causes too much task moving and starvation. The reasoning is this:

- Workers that are faster finishing tasks have shorter queue. (That was my assumption).
- Workers that have shorter queue triggers balance(meaning gets task from longer queue) more frequently.
- Workers that have shorter queue frequently starve between two `balance`s.

IMHO the balance doesn't need to be done, unless there is at least one "short enough" (naively this is empty, if we worry about starvation then this can be a threshold)queue. The reasoning is this:

- Some tasks might block. Therefore a worker with 100 tasks queued is not necessarily faster to complete than a worker with 1 task.
- If all the queues are not short enough, moving from A to B doesn't make tasks complete faster. And moving too frequently might be a waste.

My idea was then to implement an algorithm that does roughly the following:

- Check for short enough queue in each balance call. This should be an absolute value (based on queue size).
- Balance if there is a short enough queue. Don't balance if there isn't one.
- Fill up shorter queue first if we assume that "workers that are faster finishing tasks have shorter queues".

With that said, I am happy to implement either one of these two methods. 

gxu

gxuu:
UPD: 

This commit refactors the worker queue handling and balance algorithm.  The changes include:

*   Modified the original function signature instead of introducing new APIs.
*   Updated the `entry_point/` implementation to use the new functions.
*   Removed references to `per_worker_queue_size`.
*   Allowed worker queue sizes to be initialized externally, both as command-line arguments and function parameters.
*   Revised the implementation for symphony workers.
*   Modified the balance algorithm.
*   Updated all test implementations to use the new functions.

Note on design:

* `per_worker_queue_size` is probably worth keeping. It can be used as default when queue sizes are not provided. My opinion is that the name is confusing, and therefore I removed it for now.
* Symphony CLUSTER seems to have only a single worker instead of a cluster. I am not sure whether this is intended or a work in progress. For now, the worker in symphony cluster only have one queue size.
* worker queue sizes doesn't have a default value for now, but it might be worth adding one. Otherwise, the code looks nasty.

I would like to receive one more round of feedback before cleaning up the mess. That said, I think all the changes that needs to be done are done.

gxu

PR #66, opened 2025-03-27
Make SchedulerClusterCombo's address arg defaulted to None

gxuu:
Internally, `SchedulerClusterCombo` will get an available port and use that. User will then get address through `get_address` method.

### DIFF
--- a/scaler/utility/network_util.py
+++ b/scaler/utility/network_util.py
@@ -1,0 +1,8 @@
+import socket
+
+def get_available_tcp_port(hostname: str = "127.0.0.1") -> int:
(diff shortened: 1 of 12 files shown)


yzard (commented):
if address is None:
    self._address = ZMQConfig.from_string(...)
else:
    self._address = ZMQConfig.from_string(address)


def get_address(self) -> str:
    return self._address.to_address()

scaler/utility/network_util.py
```diff
@@ -0,0 +1,8 @@
+import socket
+
+def get_available_tcp_port(hostname: str = "127.0.0.1") -> int:
```
yzard, line 3: if we move this to library, remove the one used in tests

PR #76, opened 2025-04-03
Make project version appears on sphinx doc

gxuu:
As the title suggests

### DIFF
--- a/docs/source/_templates/layout.html
+++ b/docs/source/_templates/layout.html
@@ -2,3 +2,13 @@
 {% block extrahead %}
   <link href="{{ pathto("_static/style.css", True) }}" rel="stylesheet" type="text/css">
 {% endblock %}
+
+{% block sidebartitle %}
+  <div class="switch-menus">
+    <div class="version">
+      <strong>Version:</strong> {{ version }}
+    </div>
+  </div>
+
+  {{ super() }}
+{% endblock %}


gxuu:
An example was added, and so does its docs.

examples/applications/implied_volatility.py
```diff
@@ -0,0 +1,74 @@
...
+    for _ in range(0, 3):
+        size = 10000
+        S = np.random.randint(100, 200, size)
+        K = S * 1.25
+        T = np.ones(size)
+        R = np.random.randint(0, 3, size) / 100
+        vols = np.random.randint(15, 50, size) / 100
+        prices = bs_call(S, K, T, R, vols)
+        params = np.vstack((prices, S, K, T, R))
+        futs.append(client.submit(wrapper, find_vol, params))
```
rafa-be, line 67: I don't understand why `map()` does not work.

From your comment, I've the feeling that using `map()` was slower than `submit()`, is that the reason?

I would not expect `map()` to be slower. However, there is a slight difference in how lazily the two functions execute/queue the computations, which might create a kind of latency difference.

examples/applications/implied_volatility.py
```diff
@@ -0,0 +1,74 @@
...
+
+    for fut in futs:
+        fut.result()
```
rafa-be, line 70: Would it make sense to print the results? 

If you only wish to wait on the futures, you can use [`concurrent.futures.wait()`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait) instead.

examples/applications/implied_volatility.py
```diff
@@ -0,0 +1,74 @@
...
+        sigma = sigma + diff / vega  # f(x) / f'(x)
+    return sigma  # value wasn't found, return best guess so far
+
+
+def wrapper(find_vol_function, params):
+    return list(map(find_vol_function, *params))
+
+
+def main():
+    cluster = SchedulerClusterCombo(n_workers=3)
```
rafa-be, line 40: Is `n_workers=3` an optimal value? Or else it might be better to use `psutils.cpu_count()`.

docs/source/tutorials/applications.rst
```diff
@@ -0,0 +1,32 @@
...
+
+.. literalinclude:: ../../../examples/applications/implied_volatility.py
+   :language: python
+
+
+Get Option Closing Price
+------------------------
+
+This example gets option closing price for a specified ticket and the start date. We use `client.map` to reduce the
+overhead introduced by slow IO speed.
```
rafa-be, line 26: Shouldn't it be "ticker" and not "ticket"? I might be wrong.

sharpener6 (dismissed):

sharpener6:
@gxuu, see above linter

examples/applications/implied_volatility.py
```diff
@@ -0,0 +1,80 @@
...
+    #
+    # V_market = bs_call(S, K, T, r, vol)
+    # implied_vol = find_vol(V_market, S, K, T, r)
+    #
+    # print ('Implied vol: %.2f%%' % (implied_vol * 100))
+    # print ('Market price = %.2f' % V_market)
+    # print ('Model price = %.2f' % bs_call(S, K, T, r, implied_vol))
+
+    futs = []
+    for _ in range(0, 3):
```
rafa-be, line 65: I'm not sure I why you do these 3 iterations.

From what I understand, to benefit from Scaler's parallelism, `find_vol()` should be called N times (3 ?) with `MAX_ITERATIONS / N`, and then somehow combine the N estimates. Or else, I don't see the benefit of using Scaler.

examples/applications/yfinance_historical_price.py
```diff
@@ -0,0 +1,65 @@
...
+        option_history = get_option_history_data(contract_symbol, days_before_expiration)
+        first_option_history = option_history.iloc[0]
+        first_option_history_date = option_history.index[0]
+        first_option_history_close = first_option_history["Close"]
+        res.append((contract_symbol, first_option_history_close, first_option_history_date))
+    return res
+
+
+def main():
+    cluster = SchedulerClusterCombo(n_workers=2)
```
rafa-be, line 49: I got faster results with `n_workers=psutils.cpu_count()`.

More workers take a little bit more take time on startup, but the actual computation is `n_workers` times faster.

examples/applications/yfinance_historical_price.py
```diff
@@ -0,0 +1,65 @@
...
+    client = Client(address=cluster.get_address())
```
rafa-be, line 50: I personally prefer to showcase the use of the client with `with Client(...) as client:`, but that's minor.

docs/source/tutorials/applications.rst
```diff
@@ -0,0 +1,33 @@
...
+Quickstart section. Those examples are better starting point than examples in this page. 
+
+
+Calculate Implied Volatility
+----------------------------
+
+This example calculate implied volatility for many given market price. We use `client.submit` to achieve such goal.
+Alternatively, one may use `client.map` if their computation function is heavier than what's shown in the example.
+Remember, `client.map` makes sense only in two cases: A. The body of the function you are submitting is quite large; 
+B. The function is very computation heavy or is possible to block.
```
rafa-be, line 17: This paragraph has to be rewritten as implied_volatility.py now uses `map()`.

rafa-be (changes requested):
This looks almost ready for merge to me. Way better than the latest version 💪

rafa-be (approved):

examples/applications/implied_volatility.py
```diff
@@ -0,0 +1,93 @@
+"""
+This program computes the implied Black-Scholes volatility given market price and model price.
+
+This program is revised based on
+https://stackoverflow.com/questions/61289020/fast-implied-volatility-calculation-in-python
+
+Usage:
+
+    $ python -m examples.applications.implied_volatility
```
sharpener6, line 9: is this will work under any directory? if we distribute scaler examples, then we should put examples under scaler directory in site-packages, put this into root directory of site-packages is not good idea

sharpener6 (dismissed):

sharpener6 (approved):

PR #82, opened 2025-04-11
Initial scaler c++ object storage sever

gxuu:
This PR aims to provide object storage functionality in C++. Below is the protocol I will be using to achieve such goal.

```
@0xc2a14174aa42a12a;

struct ObjectProtocol {
    objectId @0: Data; # 128 bits
    objectName @1: Data; # Undetermined
    payload @2: UInt64;
    instruction @3: ObjectInstructionType;

    enum ObjectInstructionType {
        createObjectByNameAndId @0;
        getObjectByName @1;
        getObjectById @2;
        deleteObjectByName @3;
        deleteObjectById @4;
        getObjectNameById @5;
        setObjectNameById @6;
    }
}
```
* `objectId` will contain unique object id. This is the source of truth.
* `objectName` will contain the name of a specific object. This name is not necessarily unique.
* `payload` will contain the payload's length, as an unsigned 64 bits integer.
* `instruction` is one value of the `enum`.
* For `enum ObjectInstructionType`,
* * `createObjectByNameAndId` will create object by given name and id. All three arguments are expected to be filled. Upon success, `objectId` that's passed in will be return; Upon failure, Null `objectId` will be returned. *Content will not be returned*.
* * `getObjectByXX` will first lookup whether the object exist in the system. If exists, all arguments will be filled in, and payload will be after the header. If does not exist, Null `objectId` will be returned.
* * The rest of the APIs follow similar pattern.


gxuu:
UPD on protocol: `payload` is now renamed to `payload_length`.

gxuu:
UPD:
- Requirements: compiler with C++20 coroutine support
- Dependencies: boost.asio, GTest, Capnp
- Typical cmake - make workflow should build everything
- Only 'success' cases are implemented.
- Roadmap:
	- error handling
	- add logging
	- add config to port
	- write examples
WARNINGS:
 - Tests are expected to be runned after the program start separately
 - *Only* the first round of test will succeed when the program outlives
   the tests. The second round of test will fail due to record mismatch.
   This is expected behaviour.
 - Client <-> Server connections are long connections. However SO_LINGER
   and other good options are not set. One may encounter the classic
   port used error. This is expected behaviour.

All issues listed in the warning will be fix by the next patch.

Status quo: It is usable, if you use it correctly.

scaler/scheduler/object_usage/scaler_object/scaler_object/client.cpp
```diff
@@ -0,0 +1,71 @@
+
```
sharpener6, line 1: @gxuu, please rename this file to `server.cpp`, also move all those cpp code to root directory, under the new directory called object_storage,

so it should have directory called `scaler/object_storage/protocol/header.capnp`, `scaler/object_storage/server.cpp`

gxuu: resovled.

gxuu:
UPD2:
 - Now can specify port. E.g. ./server 11451. Default to 55555
 - Added hard limit for memory usage to prevent OOM killing the program
 - Error handling are done, object_id 0 is now reserved.

Roadmap:
 - Replace err messages from printf to actual log with timestamp etc.
 - Write some easy-to-follow tests.
 - Write some examples

object_storage/.clang-format
rafa-be, line 1: Shoudln't this be shared for the whole project, not only `object_storage/`?

object_storage/server.cpp
```diff
@@ -0,0 +1,266 @@
...
+    header.object_id = {
+        object_id.getField0(),
+        object_id.getField1(),
+        object_id.getField2(),
+        object_id.getField3(),
+    };
+  } catch(boost::system::system_error& e) {
+    // TODO: make this a log, since eof is not really an err.
+    printf("exception throwned, read error e.what() = %s\n", e.what());
+    throw e;
```
rafa-be, line 90: What is the point of printing the exception if you re-throw it?

object_storage/server.cpp
```diff
@@ -0,0 +1,266 @@
...
+    case type::SET_OBJECT_CONTENT_BY_ID:
+      object_id_to_meta[header.object_id].object = std::make_shared<object_t>(std::move(payload));
+      break;
+
+    case type::SET_OBJECT_NAME_BY_ID:
+      object_id_to_meta[header.object_id].name = std::move(payload);
+      break;
+
+    case type::DEL_OBJECT_BY_NAME:
+      for(const auto& [obj_id, meta] : object_id_to_meta) {
```
rafa-be, line 144: Wouldn't it be better to have an additional `std::map<object_name_t, object_id_t>` to avoid this `O(n)` loop?

sharpener6:
I have second thought for following stuff:

1. let's rename `object_header.capnp` to `object_instruction_header.capnp`
2. move the object_storage directory into `scaler/object_storage`, this is not related to `scaler/scheduler/object_storage`, please leave that alone
3. pip install scaler will include pre-compiled `object_storage_server` as .so or executable into python package
4. please apply proper clangd_format
5. we might need change the object storage protocol to following:

```
@0xc2a14174aa42a12a;

struct ObjectInstructionHeader {
    objectId @0: ObjectId; # 32 bytes
    payloadLength @1: UInt64; # 8 bytes
    instruction @2: ObjectInstructionType; # 2 bytes

    enum ObjectInstructionType {
        setObjectById @0;
        getObjectById @1;
        getObjectHeaderById @2;
        delObjectById @3;
    }

    struct ObjectId {
        field0 @0: UInt64;
        field1 @1: UInt64;
        field2 @2: UInt64;
        field3 @3: UInt64;
    }
}
```

object_storage/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+else()
+  set(CMAKE_BUILD_TYPE "Debug")
+  message(STATUS "${PROJECT_NAME} Debug build")
+endif()
+
+
+find_package(CapnProto CONFIG REQUIRED)
+message(STATUS "Found CapnProto")
+
+find_package(Boost REQUIRED)
```
sharpener6, line 30: find_package(Boost CONFIG REQUIRED)

object_storage/.clang-format
```diff
@@ -0,0 +1,234 @@
+
```
sharpener6, line 1: I will submit .clang_format on root directory of scaler project, please #83

gxuu: removed clang-format in this pr

object_storage/.clang-format
gxuu: resolved by new clang-format pr

object_storage/server.cpp
gxuu: rethrow here to stop the execution flow

scaler/object_storage/object_instruction_header.capnp
```diff
@@ -0,0 +1,30 @@
+@0xc2a14174aa42a12a;
```
sharpener6, line 1: As I put in previous comment, move this to `scaler/protocol`?

scaler/object_storage/object_instruction_header.capnp
```diff
@@ -0,0 +1,30 @@
...
+
+struct ObjectInstructionHeader {
+    objectId @0: ObjectId; # 32 bytes
+    payloadLength @1: UInt64; # 8 bytes
+    instruction @2: ObjectInstructionType; # 2 bytes
+
+    # enum ObjectInstructionType {
```
sharpener6, line 8: remove those comment out code

scaler/object_storage/object_instruction_header.capnp
```diff
@@ -0,0 +1,30 @@
...
+        delObjectByID @3;
+    }
+
+    struct ObjectId {
+        field0 @0: UInt64;
+        field1 @1: UInt64;
+        field2 @2: UInt64;
+        field3 @3: UInt64;
+    }
+}
```
sharpener6, line 30: we also need a response header:
```
struct ObjectResponseHeader {
    objectID @0: ObjectID;
    payloadLength @1: UInt64;
    objectResponseType @0: ResponseType;

    enum ObjectResponseType {
        setOK @0;
        setOKOverride @1;
        getOK @2;
        getNotExist @3;
        delOK @4;
        delNotExist @4;
    }
}
```

scaler/object_storage/object_instruction_header.capnp
```diff
@@ -0,0 +1,30 @@
...
+    # }
+
+    enum ObjectInstructionType {
+        setObjectByID @0;
+        getObjectByID @1;
+        getObjectHeadBytesByID @2;
+        delObjectByID @3;
+    }
+
+    struct ObjectId {
```
sharpener6, line 24: `struct ObjectID`

scaler/object_storage/object_instruction_header.capnp
gxuu: Current design is to reuse `ObjectInstructionHeader`.  My initial thought is to reserve `object_id = 0`, but we can look more into possible solutions.

scaler/object_storage/object_instruction_header.capnp
gxuu: This requires top level CMakeLists.txt.

object_storage/server.cpp
```diff
@@ -0,0 +1,266 @@
...
+using object_t        = std::vector<unsigned char>;
+using shared_object_t = std::shared_ptr<object_t>;
+using payload_t       = std::vector<unsigned char>;
+
+struct object_with_meta {
+  shared_object_t              object;
+  std::optional<object_name_t> name;
+};
+
+std::map<object_id_t, object_with_meta> object_id_to_meta;
```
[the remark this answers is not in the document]
gxuu: Once the design is fixed and the project needs it, I will do it.

scaler/object_storage/object_instruction_header.capnp
gxuu: will do.

scaler/object_storage/object_instruction_header.capnp
sharpener6: move the CMakeLists.txt to root of `scaler` folder then

scaler/object_storage/object_instruction_header.capnp
sharpener6: ?

scaler/object_storage/object_instruction_header.capnp
sharpener6: how about rename to `ObjectRequestHeader` and `ObjectResponseHeader`

also we need objectID in objectResponseHeader to identify what object_id it relates to, client will need that if they submit ObjectRequestHeader asynchronously

sharpener6:
We will need add a little enhancement here:

1. there will be a lot round trip from client to object storage manager, to save those round trips, please do following:
   1. for all incoming ObjectRequestHeader get request, let's have a per tcp peer queue for those requests, so the problem is object creation request might come after get requests, and if we respond with not exists message, it might take much more time, so the solution is, let's keep a requests queue for each tcp peer, if until we have data, then we process the request and send response back
   2. if tcp link is broken, we drop all the requests for that peer


For the request protocol:
```
@0xc2a14174aa42a12a;

struct ObjectInstructionHeader {
    objectId @0: ObjectId; # 32 bytes
    payloadLength @1: UInt64; # 8 bytes
    instruction @2: ObjectInstructionType; # 2 bytes

    enum ObjectInstructionType {
        setObject @0;
        getObject @1;  // this will block if no object
        hasObject @2;  
        getObjectHeader @3;
        delObject @4;
    }

    struct ObjectId {
        field0 @0: UInt64;
        field1 @1: UInt64;
        field2 @2: UInt64;
        field3 @3: UInt64;
    }
}
```

For the reponse protocol:
```
struct ObjectResponseHeader {
    objectID @0: ObjectID;
    payloadLength @1: UInt64;
    objectResponseType @0: ResponseType;

    enum ObjectResponseType {
        setOK @0;
        setOKOverride @1;
        getOK @2;   // if object not exists, it will hang
        hasOK @3;
        hasNotExists @4;
        delOK @5;
        delNotExist @6;
    }
}
```

gxuu:
ROADMAP:
// done - Test case for halting.
// done - Test case for behavior of the server.
- A decent async model that actually provide a good enough notion of continuation.

scaler/object_storage/server.cpp
```diff
@@ -0,0 +1,168 @@
...
+    auto executor = co_await this_coro::executor;
+    tcp::acceptor acceptor(executor, {tcp::v4(), port});
+    for (;;) {
+        tcp::socket socket = co_await acceptor.async_accept(use_awaitable);
+        co_spawn(executor, process_request(std::move(socket)), detached);
+    }
+}
+
+int main(int argc, char* argv[]) {
+    short unsigned port = 55555;
```
sharpener6, line 139: please make host name and port configurable, it should able to listen to `localhost` as well as ip `0.0.0.0`, default to `localhost`, with `--host` or `-H`, same for port `--port` or `-p`, with some default port

scaler/object_storage/server.cpp
```diff
@@ -0,0 +1,168 @@
...
+using boost::asio::detached;
+using boost::asio::use_awaitable;
+using boost::asio::ip::tcp;
+namespace this_coro = boost::asio::this_coro;
+
+#if defined(BOOST_ASIO_ENABLE_HANDLER_TRACKING)
+#define use_awaitable boost::asio::use_awaitable_t(__FILE__, __LINE__, __PRETTY_FUNCTION__)
+#endif
+
+std::map<scaler::object_storage::object_id_t, scaler::object_storage::object_with_meta> object_id_to_meta;
```
sharpener6, line 36: please create a class called `ObjectStorageServer` and wrap it up, so we can refactor this later to support cluster etc.

scaler/protocol/capnp/object_instruction_header.capnp
```diff
@@ -0,0 +1,35 @@
+@0xc2a14174aa42a12a;
```
sharpener6, line 1: please rename this file to `object_storage.capnp`

gxuu: done

scaler/object_storage/server.cpp
gxuu: For `host`, is it necessary to make it configurable? User can only run this program locally. Meaning `localhost` is always set. `asio` takes care for setting `0.0.0.0`.

gxuu: Done

scaler/object_storage/server.cpp
gxuu: Done

sharpener6 (approved):

sharpener6 (approved):

PR #90, opened 2025-04-23
Add Examples to Docs

magniloquency:


docs/source/tutorials/applications.rst
```diff
@@ -30,4 +30,31 @@ overhead introduced by slow IO speed.
...
+
+
+Web Server
+----------
+
+This example shows how Scaler can be used to parallelize a web server.
+It includes a sequential version and a parallelized version using Scaler for comparison, plus an implementation of a client that rapidly sends requests to test the server.
+
+.. literalinclude:: ../../../examples/applications/web_server.py
+   :language: python
```
gxuu, line 60: I might be wrong but I recall there needs to be two newlines. Otherwise sphinx has a bug: once you clicked on other directory in the menu, the last directory will not be rendered again.

magniloquency: There are two newlines. Just tested it and it renders fine 👍

PR #95, opened 2025-04-24
Use so as server

gxuu:
A python wrapper to object storage server

- `server.py` can be executed as a program. In this case, it depends on
  python-argparse. It can also be executed as a separate module. In this
  case, user call `run_object_storage_server` and the server will be
  started. Note, if the use case is always to call
  `run_object_storage_server`, then the program should remove argparse
  as its dependencies. The caller, is expected to fill in legit args.

- The original server_as_so is renamed as server. `server` is now a
  dynamic library, that can be accepted by server.py. The original
  `server` executable is now moved to tests/test_server_main. This
  program is meant to be used in testing.

- Dependency changes: For non-testing purposes, C++ argparse is not
  a dependency. pyargparse is not a dependency to server.py if it is
  directly executed. Otherwise, pyargparse is not a dependency.

### DIFF
--- a/scaler/object_storage/server.py
+++ b/scaler/object_storage/server.py
@@ -1,0 +1,42 @@
+import argparse
+import ctypes
+import sys
+
+def load_library(lib_path):
+    """
+    Load the shared dynamic library.
+    """
+    try:
+        return ctypes.CDLL(lib_path)
+    except OSError as e:
+        print(f"Failed to load library: {e}")
+        sys.exit(1)
+
+def run_object_storage_server(lib_path: str, name: str = "127.0.0.1", port: str = "55555"):
+    # Load the library
+    lib = load_library(lib_path=lib_path)
+    # Define the argument and return types for the main_entrance function
+    lib.run_object_storage_server.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
+    lib.run_object_storage_server.restype = None
+
+    # Call the function
+    name_bytes = name.encode('utf-8')
+    port_bytes = port.encode('utf-8')
+
+    lib.run_object_storage_server(name_bytes, port_bytes)
+
+def main():
+    parser = argparse.ArgumentParser(description='Call main_entrance from a dynamic library.')
+    parser.add_argument('--lib', required=True, help='Path to the foundational library')
+    parser.add_argument('--name', '-n', default='127.0.0.1', 
+                        help='Ip address or NS record. Default to 127.0.0.1')
+    parser.add_argument('--port', '-p', default='55555', 
+                        help='Port number to connect. Default to 55555')
+    args = parser.parse_args()
+
+    run_object_storage_server(args.lib, args.name, args.port)
+
+
+if __name__ == '__main__':
+    main()
+
--- a/scaler/object_storage/CMakeLists.txt
+++ b/scaler/object_storage/CMakeLists.txt
@@ -1,29 +1,9 @@
-include(FetchContent)
-set(ARGPARSE_GIT_REPOSITORY "https://github.com/p-ranav/argparse.git")
-set(ARGPARSE_GIT_TAG "d924b84eba1f0f0adf38b20b7b4829f6f65b6570")
-FetchContent_Declare(
-    argparse
-    GIT_REPOSITORY ${ARGPARSE_GIT_REPOSITORY}
-    GIT_TAG ${ARGPARSE_GIT_TAG}
-)
-FetchContent_MakeAvailable(argparse)
-message(STATUS "Fetched argparse from ${ARGPARSE_GIT_REPOSITORY} @ ${ARGPARSE_GIT_TAG}")
-
-add_executable(server)
-# add_library(server_as_so SHARED)
+add_library(server SHARED)
 
 target_sources(server PRIVATE
   server.cpp
   io_helper.cpp
 )
-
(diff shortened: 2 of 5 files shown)


scaler/object_storage/server.py
```diff
@@ -0,0 +1,42 @@
...
+    lib.run_object_storage_server.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
+    lib.run_object_storage_server.restype = None
+
+    # Call the function
+    name_bytes = name.encode('utf-8')
+    port_bytes = port.encode('utf-8')
+
+    lib.run_object_storage_server(name_bytes, port_bytes)
+
+def main():
```
sharpener6, line 28: create entry point just like `scaler_scheduler`, you need also add entry point in `pyproject.toml`

scaler/object_storage/server.py
```diff
@@ -0,0 +1,42 @@
...
+    parser = argparse.ArgumentParser(description='Call main_entrance from a dynamic library.')
+    parser.add_argument('--lib', required=True, help='Path to the foundational library')
+    parser.add_argument('--name', '-n', default='127.0.0.1',
```
sharpener6, line 31: you need use black to format it

scaler/object_storage/server.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # Call the function
+    name_bytes = name.encode('utf-8')
+    port_bytes = port.encode('utf-8')
+
+    lib.run_object_storage_server(name_bytes, port_bytes)
+
+def main():
+    parser = argparse.ArgumentParser(description='Call main_entrance from a dynamic library.')
+    parser.add_argument('--lib', required=True, help='Path to the foundational library')
```
sharpener6, line 30: if should automatically detect the lib, without configuration, the lib is always the same position relate to this server.py file

gxuu: Currently `lib` exists because the logic of packaging has not been done. It should be fixed. Please ignore this problem for now.

scaler/object_storage/server.py
gxuu: done

scaler/object_storage/server.py
gxuu: done

scaler/entry_points/object_storage_server.py
```diff
@@ -0,0 +1,25 @@
...
+    )
+    parser.add_argument(
+        "--port", "-p", type=str, default="55555", help="Specify port the server will be listening to. Default to 55555"
+    )
+    return parser.parse_args()
+
+
+def main():
+    args = get_args()
+    run_object_storage_server("../lib/libserver.so", args.name, args.port)
```
sharpener6, line 25: this will not work, it changes as your working directory get changed.

1. your build system need compile `libobjectserver.so` and move to `scaler/lib/`
2. you need include lib in the wheel
3. this "../lib/libobjectserver.so" will not work as your working directory keep changing, you will need do as I states in my previous comments: `os.join(os.path.dirname(__file__), "..", "lib", "libobjectserver.so")`

setup.py
```diff
@@ -0,0 +1,79 @@
+import os
```
sharpener6, line 1: setup.py is deprecated, we use pyproject.toml,

sharpener6: https://stackoverflow.com/questions/74409966/how-to-replace-setup-py-with-a-pyproject-toml-for-a-native-c-build-dependency

gxuu: `setup.py` is still used according to the building package python 3.10 guide.
Besides, all big projects that I checked (numpy, pandas, and pytorch) are still using it.

sharpener6: > setup.py is deprecated, we use pyproject.toml, please remove it

Hold on this, let me do some research

scaler/entry_points/object_storage_server.py
gxuu: The entry point program will never move location.

1. My build system does compile and move the binary to `scaler/lib`. 
2. I included lib in the wheel.

If you don't mind, please run `python -m build` and see the output.

gxuu: Changed, as per instructed.

gxuu:
Now the project uses `scikit-build-core>=0.5.0` as its building backend.
The `setup.py` is no longer needed, as the new backend integrates with
CMake nicely. Few things to discuss before further changes apply:

- The backend does not allow reading dependencies from requirements.txt.
  The project maintainer needs to decide whether we need to maintain two
  sets of requirements at the same time.
- There might be differences in package METAINFO and other information,
  due to the two backends are not compatible with each other. How to
  tackle this?

And this PR is therefore not complete. If the maintainer decides to
remove requirements.txt then other things such as Github workflow will
make changes accordingly. Combining those changes in one PR for easier
navigation in the future. If the maintainer decides to maintain all
information the original package maintains, then we might as rollback to
build script.

--------

Details on which method of packaging will be posted once the maintainer accepts the idea of removing requirements.txt.

sharpener6:
@gxuu , then
1. remove `requirements.txt`
2. change linter.yaml that does `uv pip install --system -r requirements.txt` to `uv pip install --system -r pyproject.toml`

sharpener6 (approved):

sharpener6 (approved):

PR #99, opened 2025-04-29
Object Storage Server Improvements

gxuu:
First commit:
-------------

Update CMake to work with new build backend

We no longer need to hard code (well, still a little bit) the version
number in CMakeLists.txt now. When invoked by the python build backend,
the backend automatically define some variables for us. We use them.

Also do not build tests when we are not manually running CMake. This is
temporary, and will be changed once we get real unit tests.

Second commit:
--------------

Create a run_object_storage_server.py on root directory

This is for debugging when engineering on git version (not install
package).

Third commit:
-------------

Update naming convention following C++ Coding Guidance

- Non-free functions are now renamed to camelCase;
- Variables are now renamed to camelCase as well;
- Tests and free functions are not yet renamed, reasons as follows:
  - we will get rid of most tests soon enough as we will write UT;
  - free functions have a long tradition of being snake style, we should
  discuss this.


sharpener6 (approved):

gxuu:
Fourth commit:
--------------

Decouple updateRecord and network logic

So that we don't have to deal with asio when calling updateRecord, 
helps with testing

Fifth Commit:
-------------

Add unit test that doesn't need standalone server running

Update CMake script to build this new example. Don't remove old examples
aka test_scaler_object for now. Planning to remove it after the code is
redesigned.

Update the examples. The new example file is much shorter, cleaner,
doesn't require users to look at asio details, and most importantly,
doesn't need standalone server to run.

sharpener6 (approved):

sharpener6:
@gxuu , seems like the updateRecord is not public method

gxuu:
> @gxuu , seems like the updateRecord is not public method

Root cause: `updateRecord` and internal map were defined to be `public` when it is a debug build, and private when not a debug build. `linter.yml` is building the project in a release build.

Fix: Change build command in `linter.yml` to be a debug build.
Rationale: Building and running test should not happen in `release`. Building a release build is also slower due to optimizations the compiler performs. 

It should work for now. I've also updated the `linter.yml` such that it runs through the unit test automatically.

sharpener6 (dismissed):

examples/applications/yfinance_historical_price.py
```diff
@@ -11,12 +11,15 @@
 import datetime
 from typing import Optional
 
+# import numpy as np
```
sharpener6, line 14: move those 2 commented out imports into the function along with commented out code

examples/applications/yfinance_historical_price.py
```diff
@@ -37,20 +40,34 @@ def get_option_history_data(contract_symbol, days_before_expiration: int = 30):
...
+    #         first_option_history_close = first_option_history["Close"]
+    #         res.append((contract_symbol, first_option_history_close, first_option_history_date))
+    # except YFRateLimitError:
+    #     print("Early return due to requesting too many requests.")
+    # return res
+
+    # NOTE: Here, we are mocking data that we will be receiving, and pass them back. This is to avoid creating network
+    # traffic to a third party. If you wish to get data from Yahoo, comment out below section, and uncomment everything
+    # that's above.
+    df = pd.read_csv("mock_data.csv")
```
sharpener6, line 65: rename `mock_data.csv` to `downloaded_data.csv`

gxuu: done

examples/applications/yfinance_historical_price.py
gxuu: done

sharpener6 (dismissed):

sharpener6 (approved):

PR #100, opened 2025-05-01
Move object storage server tests out of tree

gxuu:
The title of this PR is misleading. See comment for what it does. It is basically the second round of enhancement to the object storage server that is less important than the last one. 

- object storage server tests were in the source tree, while the correct place of putting tests is tests/ directory. This changes the tests location.

- test_scaler_object test was not self-contained. This changes it, and make the test self-contained, so we can treat it as a unit test.

- Dependency CapnProto is now a project dependency instead of a component dependency. Rationale: Dependencies of a necessary component of a project should be treated as a project dependency; Dependencies of an optional component of a project (such as tests) should be placed in sub-directory. Since scheduler will also use CapnProto, it is considered as a necessary dependency.

- Project directory is now an include directory to every source file. C++ sources should always write "scaler/..." to get the header files they need.

TODO:

 - Fix linter.yml behavior for tests as tests are now moved to another position.

 - Remove redundant tests in scaler/object_storage.

 - Logging support to object storage server, currently the error message is raw.

 - Fix the packaging logic to the following: all build should not contain tests and examples; sdist build should not contain any binary; and wheel build should contain dependencies' binaries. Currently the packaging logic does not conform to these rules.

 - Further simplify the test_scaler_object, the connecting logic should also be moved to global setup.

 - Potentially change cmake script to get SO from system directory.


gxuu:
Done:

- object storage server tests were in the source tree, while the correct
  place of putting tests is tests/ directory. This changes the tests
  location.

- test_scaler_object test was not self-contained. This changes it, and
  make the test self-contained, so we can treat it as a unit test.

- Dependency CapnProto is now a project dependency instead of a
  component dependency. Rationale: Dependencies of a necessary component
  of a project should be treated as a project dependency; Dependencies
  of an optional component of a project (such as tests) should be placed
  in sub-directory. Since scheduler will also use CapnProto, it is
  considered as a necessary dependency.

- Project directory is now an include directory to every source file.
  C++ sources should always write "scaler/..." to get the header files
  they need.

- Fix linter.yml behavior for tests as tests are now moved to another
   position.

- Remove redundant tests in scaler/object_storage.

- Further simplify the test_scaler_object, the connecting logic should
   also be moved to global setup. (CANCELLED. Test cannot access
   variable defined in global setup).

TODO:

 - Logging support to object storage server, currently the error message
   is raw.

 - Fix the packaging logic to the following: all build should not
   contain tests and examples; sdist build should not contain any
   binary; and wheel build should contain dependencies' binaries.
   Currently the packaging logic does not conform to these rules.

 - Potentially change cmake script to get SO from system directory.

gxuu:
Move object storage server tests out of tree

    - object storage server tests were in the source tree, while the correct
      place of putting tests is tests/ directory. This changes the tests
      location.

    - test_scaler_object test was not self-contained. This changes it, and
      make the test self-contained, so we can treat it as a unit test.

    - Dependency CapnProto is now a project dependency instead of a
      component dependency. Rationale: Dependencies of a necessary component
      of a project should be treated as a project dependency; Dependencies
      of an optional component of a project (such as tests) should be placed
      in sub-directory. Since scheduler will also use CapnProto, it is
      considered as a necessary dependency.

    - Project directory is now an include directory to every source file.
      C++ sources should always write "scaler/..." to get the header files
      they need.

    - Fix linter.yml behavior for tests as tests are now moved to another
       position.

    - Remove redundant tests in scaler/object_storage.

    - Further simplify the test_scaler_object, the connecting logic should
      also be moved to global setup. (CANCELLED. Test cannot access
      variable defined in global setup).

    - Fix the packaging logic to the following: all build should not
      contain tests and examples; sdist build should not contain any
      binary; and wheel build should contain dependencies' binaries.
      Currently the packaging logic does not conform to these rules.

    - Potentially change cmake script to get SO from system directory.
      (CANCELLED, we should let tools to handle it. Copy SO will not work
      because SO will also have dependencies).

    // TODO:
    //  - Logging support to object storage server, currently the error message
    //    is raw. (Logging support will be move to another PR and will be
    //    discussed)

gxuu:
ready for review

sharpener6 (approved):

PR #102, opened 2025-05-06
Change protocol

gxuu:
Please review this protocol change. Please be noted that the `requestID` should be way, way, way beyond how much object you would actually want to send - they are independent in each connection. A tempting idea is to reserve the top/bottom 32 bits. But that is up for discussion.

Beside this, current implementation lacks the ability to "know how much to read". That being said, if we would like to not using `read/write` process request model then we will have to "try read x bytes" before we process. Should we also add a mechanism for user to tell the server how much messages are there?

### DIFF
--- a/scaler/object_storage/io_helper.cpp
+++ b/scaler/object_storage/io_helper.cpp
@@ -6,6 +6,7 @@
 #include <boost/asio/awaitable.hpp>
 #include <boost/asio/co_spawn.hpp>
 #include <boost/asio/detached.hpp>
+#include <boost/asio/error.hpp>
 #include <boost/asio/io_context.hpp>
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/asio/read.hpp>
@@ -38,6 +39,7 @@ awaitable<void> read_request_header(tcp::socket& socket, ObjectRequestHeader& he
 
         header.reqType       = requestRoot.getRequestType();
         header.payloadLength = requestRoot.getPayloadLength();
+        header.requestID     = requestRoot.getRequestID();
 
         auto objectID   = requestRoot.getObjectID();
         header.objectID = {
@@ -48,7 +50,11 @@ awaitable<void> read_request_header(tcp::socket& socket, ObjectRequestHeader& he
         };
     } catch (boost::system::system_error& e) {
         // TODO: make this a log, since eof is not really an err.
-        printf("exception throwned, read error e.what() = %s\n", e.what());
+        if (e.code() == boost::asio::error::eof) {
+            printf("Remote end closed, nothing to read.\n");
+        } else {
+            printf("exception throwned, read error e.what() = %s\n", e.what());
+        }
         throw e;
     } catch (std::exception& e) {
         // TODO: make this a log, capnp header corruption is an err.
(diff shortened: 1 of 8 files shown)


sharpener6 (approved):

rafa-be (dismissed):

gxuu:
Implementation of the object storage server protocol v5

For now, blocking happens on the level of requests instead of the level
of connections. This means every requests gets an id, and every response
contains an id, where these two ids are the same.

Get request when no object presents will be blocked. However, client is
able to send other requests (or even another blocking get request), and
get corresponding response back from the same connection channel.

Notice that the order of responses is not guaranteed anymore; Should
user wish to have an order, they need to sort responses based on the
responseId, with the same relation they have in the requestId.

scaler/object_storage/io_helper.cpp
```diff
@@ -48,7 +50,11 @@ awaitable<void> read_request_header(tcp::socket& socket, ObjectRequestHeader& he
         };
     } catch (boost::system::system_error& e) {
         // TODO: make this a log, since eof is not really an err.
-        printf("exception throwned, read error e.what() = %s\n", e.what());
+        if (e.code() == boost::asio::error::eof) {
+            printf("Remote end closed, nothing to read.\n");
```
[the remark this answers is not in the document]
gxuu: done

scaler/object_storage/object_storage_server.h
```diff
@@ -22,7 +22,7 @@ using boost::asio::ip::tcp;
 
 class ObjectStorageServer {
     struct meta {
-        boost::asio::ip::tcp::socket socket;
+        boost::asio::ip::tcp::socket& socket;
```
rafa-be, line 25: I'm a little bit surprise by this reference to the socket.

Is it because the field is always accessed within the `for(;;)` loop of `process_request()`, which ensure that the socket is always alive? (before you didn't had that guarantee as `process_request()` could return).

rafa-be (approved):

scaler/object_storage/object_storage_server.h
rafa-be: Related to this, what if the connection making the blocking GET request closes/finishes before the SET request arrives? Will the server try to send the GET-OK response to the closed socket? Or are you cleaning up the `meta` values somewhere (I couldn't find it).

gxuu: `socket` is always accessed in `process_request`'s scope, because we no longer halt the whole connection anymore. (It was a value because the connection needs to be halt, which is equivalent to exiting `process_request` and that's why the value has to be moved).

To the second comment - 
When this socket `read/write` and the remote end is closed, it throws exception. But that's fine, since we catch it outside and handle them there.

I think there is a bug somewhere that I need to look into further. Please do not merge the code for now.

gxuu: What you said is perfectly true, and this is a small problem to fix. There is a larger problem that I just fixed and will hopefully push in the next twenty mins or so. I will exp it in more detail later

gxuu: The bigger issue:

The `socket&` is holding a socket from *some* connections. This is needed because we want proactive async execution flow. Our implementation says that if a socket has its remote end closed, then we close the connection as well. That has an implication that this socket will be destroyed as we exited `process_request`. That ref became a dangling ref.

Furthermore, if undesired disconnect happens, then we also have dang ref.


--------------------------------------------
on why we cannot have a queue for each connections (aka why we have to write this ugly code):

1. One queue per connection essentially means polling, and this polling is slow. 
2. The pre-condition of the polling is that there are new requests from the same connection, that's not always the case.
   Consider `C_a` sending only one get object request that turns out to be blocking and wait, and `C_b` sends a set request to the object that `C_a` want to get.
  Now if we maintain one queue for each connection, then `C_a` will never get the result back. Because `C_a` execution flow will be blocked on `read` messages!

  The workaround to this problem is of course, setting a timeout to the read request call. But that just compicates the design

gxuu: And to why it complicated design is simple: How much time should we set the expiration to be? The length of this expiration essentially means how long the remote end needs to wait before they can possibly get the response.

sharpener6 (approved):

PR #106, opened 2025-05-10
Client, Scheduler and Worker use the new Object Manager server

rafa-be:
‼️ **Work In Progress** ‼️

This PR removes the object's data management from the scheduler, moving it to the new C++ Object Manager.

Here are is a diagram with the new data flows:

scaler_object_storage

Here are the major changes:

- Two new network connectors that link directly to the object storage server (`AsyncObjectStorageConnector` and `SyncObjectStorageConnector`);
- The client and its futures directly connect to the object storage. Object's data do not get transmitted to the client's agent, only the `ObjectInstruction` messages do;
- Only the scheduler sends `Delete` messages to the storage server, except for the task result objects, that are also immediately deleted by the client when received. The scheduler might `Delete` these object a second time, but that's OK;
- There is an unexpected async connector that runs inside the worker's agent. Its only purpose is to create the exception objects when a processor fails;
- The Symphony worker has been greatly simplified, as the async object connector removes the need for the `asyncio` lock logic from the previous implementation @1597463007.


scaler/entry_points/scheduler.py
```diff
@@ -23,6 +23,14 @@
 def get_args():
     parser = argparse.ArgumentParser("scaler scheduler", formatter_class=argparse.ArgumentDefaultsHelpFormatter)
     parser.add_argument("--io-threads", type=int, default=DEFAULT_IO_THREADS, help="number of io threads for zmq")
+    parser.add_argument(
+        "--object-storage-address",
+        "-osa",
```
gxuu, line 28: Consider single letter arg name when prefixed with `-`, if there isn't a good enough letter, remove short option.
This is because lots of people is accustomed to the bash format, where each letter is parsed as a separate arguments to the program. For example `ls -la` is equivalent to `ls -l -a`. They would be surprise when seeing `-osa` as a single arg.

scaler/entry_points/scheduler.py
```diff
@@ -23,6 +23,14 @@
...
+        type=str,
+        default=None,
+        help="specify object storage address, the object storage address will be send to client and workers so they "
+        "can talk to object storage server directly, e.g.: localhost:6379",
```
gxuu, line 32: Be noted that when user typed in `localhost:port`, it will be resolved to `ipv6` address. E.g., if you pass in the address as `name="localhost"`, and `port="6379"` to `run_object_storage_server`, connecting to `127.0.0.1:6379` will not work. This is because `dns` util in the server side is doing so, and there is no good way to change it.
I am not sure whether this will affect the client side though, just put the note here.

scaler/io/sync_object_storage_connector.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def __ensure_response_type(
+        self,
+        header: ObjectResponseHeader,
+        valid_response_types: Iterable[ObjectResponseHeader.ObjectResponseType],
+    ):
+        if header.response_type not in valid_response_types:
+            raise RuntimeError(f"unexpected object storage response_type={header.response_type}.")
+
+    def __ensure_empty_payload(self, payload: bytearray):
```
gxuu, line 95: Depending on when you call this function, the following may be true.
If you call this after a non-blocking call (for example `delete_object`), and check whether the payload is exactly 0, and in the meantime another client sends a `set_object` and causes the server to reply to your pending `get_object` request (suppose there is one, of course), then a response will be sent to you when you are calling this function, lead to undesired error.

scaler/io/async_object_storage_connector.py
```diff
@@ -0,0 +1,163 @@
...
+        object_id: bytes,
+        payload_length: int,
+        request_type: ObjectRequestHeader.ObjectRequestType,
+        payload: Optional[bytes],
+    ):
+        self.__ensure_is_connected()
+        assert self._writer is not None
+
+        request_id = self._next_request_id
+        self._next_request_id += 1
```
gxuu, line 119: Though this will probably never happen: consider adding a test that reports error when the `_next_request_id` is too large. When overflow occurs on the server side, the server will mod this number automatically by `UINT64_MAX`, which means you will never get the corresponding `responseID` that's the same as your `requestID`.

scaler/object_storage/object_storage_server.h
```diff
@@ -119,13 +119,21 @@ class ObjectStorageServer {
 public:
     awaitable<void> process_request(std::shared_ptr<tcp::socket> socket) {
         try {
+            std::cerr << "New connection" << std::endl;
```
gxuu, line 122: These debug messages should probably be removed.

scaler/utility/object_utility.py
```diff
@@ -9,11 +9,14 @@
 def generate_object_id(identity: bytes, object_bytes: bytes) -> bytes:
     identity_hash = hashlib.md5(identity).digest()
     object_hash = hashlib.md5(object_bytes).digest()
-    return identity_hash[:8] + object_hash
+    object_id = identity_hash + object_hash
+    assert len(object_id) == 32
```
gxuu, line 13: Is this needed?

gxuu: The `assert`, I meant.

scaler/entry_points/scheduler.py
sharpener6: it should be `tcp://localhost:6379`, in case if we will use other kind protocol

scaler/entry_points/scheduler.py
gxuu: > @gxuu, this is convention for shortcut so many options, please take a look at scaler source code first

@sharpener6 My bad. Diff was showing that one change, and I honestly didn't realize that we are using multi-letter short option in `scaler`.

scaler/entry_points/scheduler.py
rafa-be: > Be noted that when user typed in `localhost:port`, it will be resolved to `ipv6` address. E.g., if you pass in the address as `name="localhost"`, and `port="6379"` to `run_object_storage_server`, connecting to `127.0.0.1:6379` will not work. This is because `dns` util in the server side is doing so, and there is no good way to change it. I am not sure whether this will affect the client side though, just put the note here.

That's a good point. I guess everything should work with IPv6, but I haven't tried it yet.

> it should be `tcp://localhost:6379`, in case if we will use other kind protocol

It's, but the comment it out-to-date. I'll change it.

scaler/io/sync_object_storage_connector.py
rafa-be: This should not happen, as for now, the synchronous connector always blocks until it receives the response (including for Set/Delete requests), so there is only a single message on the line at any given time:

```python
sc = SyncObjectStorageConnector("127.0.0.1", 5555)

sc.set_object(id, value)  # blocks until it receives a SetOK or SetOverride

value = sc.get_object(id)  # blocks until it receives a GetOK
```

The sync connector is used by the client and the processor process (not the worker agent).

The asynchronous connector, however, can send multiple concurrent messages, and does the request/response ID matching you're talking about:

```python
ac = ASyncObjectStorageConnector("127.0.0.1", 5555)
await ac.connect()

await ac.set_object(id, value)  # does not block for the SetOK or SetOverride message

value = await sc.get_object(id)  # blocks until it receives a GetOK
```

The async connector is used by the scheduler and the worker agent.

Maybe we will remove the sync connector altogether if it becomes a bottleneck, but it adds a little bit of complexity, as the data has to be transmitted somehow to the client agent or worker agent that run the asyncio loop.

scaler/io/async_object_storage_connector.py
rafa-be: Indeed. But as you said, extremely unlikely.

scaler/object_storage/object_storage_server.h
rafa-be: Yeah, I wrongfully added them to one commit. I'll remove them from the release.

scaler/utility/object_utility.py
rafa-be: Asserts are never needed :-P.

I cannot (easily) create a 32 bytes type in Python, so looking at this function alone (and not the capnp protocol), it might look like object IDs can be of any length. Think of it as a comment with some added safety. I'll add a more meaningful error message.

sharpener6:
@rafa-be @gxuu ,
There is one edge case that current object storage server protocol will not work:
That is for test case `TestClient.test_clear`
so if user use `send_object` api which actually send object to object storage, but then quickly deleted the object, so if we send a task that uses this staled object reference, before the change, it will raise MissingObject exception and propagate to the frontend. but since now object storage will block until it have data in the object storage server, this will cause this task hanging forever

gxuu:
> if user use `send_object` api which actually send object to object storage, but then quickly deleted the object
Should the user ever do that? And should we support this?

There isn't much to do in the server side. 
- Suppose server remember whether an object id appeared before in the system, when a new request comes in, there is no way for the server side to distinguish whether it should hang or sending "once appeared in the system but now removed" response).
- Suppose server adding timeout mechanism and postpone all delete request, this will still not work because how long should we set the timeout to be? 

The ideal way to do this is to somehow make the scheduler get request reach sooner than delete object request, which can be done if we use the same tcp connection.

rafa-be:
> @rafa-be @gxuu , There is one edge case that current object storage server protocol will not work: That is for test case `TestClient.test_clear` so if user use `send_object` api which actually send object to object storage, but then quickly deleted the object, so if we send a task that uses this staled object reference, before the change, it will raise MissingObject exception and propagate to the frontend. but since now object storage will block until it have data in the object storage server, this will cause this task hanging forever

It should be OK with the current version of the Scheduler's state-machine, as clearing the client's objects will cancel all the tasks first:

```python
class Client:
     ...
     def clear(self):
        self._future_manager.cancel_all_futures()
        self._object_buffer.clear()
```

Thus the `TaskCancel(force=True)` will `SIGTERM` the processor process that is waiting on the `GET` request. Once cleared, you cannot use the old object references for new tasks, as these got invalidated.

If we remove this `force=True` feature, I think we could somehow fix the problem by making the scheduler delay the `DELETE` requests for objects that are associated with a running task.

>   Suppose server adding timeout mechanism and postpone all delete request, this will still not work because how long should we set the timeout to be?

That might still be desirable, as a kind of safety trigger, and to easily detect these problems. But it does not have to be implemented on the server side.
 
> The ideal way to do this is to somehow make the scheduler get request reach sooner than delete object request, which can be done if we use the same tcp connection.

Unfortunately, these don't use the same TCP link, as the scheduler makes the `DELETE` requests, while the worker makes the `GET` requests. But I can prevent the scheduler to send the `DELETE` request if it's associated task is running.

rafa-be:
I pushed a major update to this PR.

**All unit tests are now passing successfully**.

The general architecture hasn't changed (i.e. the diagram above is still valid).

There are the major changes:

- the **worker and processor initialization have been re-written**, that's because of the connection to the object storage server that can only be made once the connection to the scheduler has been made;
- the **object IDs are now randomly generated** (instead of based on the md5sum of their content). That's is required because there might be synchronization issues between the client, scheduler and object storage server when a client do a `clear()` request;
- I **fixed a few issues in the `ScalerFuture` class**, related to how and when result objects are obtained from the storage server;
- all **IDs (Task ID, Object ID, Worker ID, Worker ID...) now have their respective types**, instead of the previous `bytes`type.
    - that helps avoiding mistakes when dealing with these values;
    - IDs were generated by calling the `uuid` module directly, sometimes inconsistently, now it's "standardized" in these classes.  
- the **C++ object storage server got wrapped into an `ObjectStorageServer` Python class**.
    - API is now similar to the one of `Scheduler` and `Worker`;
    - the server exposes an event file-descriptor that allows the Python code to be notified when the server is ready (i.e. when the TCP socket is open).

scaler/utility/identifiers.py
```diff
@@ -0,0 +1,98 @@
...
+
+        return ClientID(f"{os.getpid()}|Client|{name}".encode())
+
+
+class WorkerID(Identifier):
+    def __repr__(self) -> str:
+        return f"WorkerID({self.decode()})"
+
+    @staticmethod
+    def generate_unique_worker_id(name: Optional[str] = None) -> "WorkerID":
```
gxuu, line 31: IMHO just `generate_worker_id` is good enough. ID somewhat means unique.

scaler/client/agent/heartbeat_manager.py
```diff
@@ -46,6 +50,11 @@ async def on_heartbeat_echo(self, heartbeat: ClientHeartbeatEcho):
         self._latency_us = int(((time.time_ns() - self._start_timestamp_ns) / 2) // 1_000)
```
magniloquency, line 50: Off topic: Why do we divide by 2 here?

rafa-be:
> Great work!
> 
> There are a few things that I'm in doubt with, they are not necessarily associated with this PR though:
> 
>     * I was inspecting the `client_manager.py`'s code. I found that `client` (which in your PR is renamed to `client_id` ) is quite large (IIRC 46 Bytes). This is way longer than `task_id` which is only 16 bytes. If they are all IDs then why length differ so much?
> 
>     * Is now a good time to fix the length of various kinds of IDs? Fixed size ID's are much easier to deal with in C++ code, and is easier to bound check.
> 
> 
> And thanks for the improvement of object storage server! It was changing so drastically within a tight schedule, so I postpone lots of clean up.

I haven't changed how client, task and worker IDs are generated, so these keep the old format, which can basically be any size as it's an UTF-8 encoded string.
Object IDs have been changed to the 32-bytes format used by the object storage server.

But I agree, it would be nice to standardize all these IDs. What I was thinking was to have random 16-bytes IDs for worker and clients, and 32-bytes IDs for tasks and objects, with the first 16-bytes always referring to the owner of the object (i.e. the client).

For example, tasks and object IDS would always like this:

```
+--------------+-----------------------+
|   16-bytes   |       16-bytes        |
+--------------+-----------------------+
| Client ID    | Random task/object ID |
+--------------+-----------------------+
```

That would help debugging whom a task or object comes from.

rafa-be:
@sharpener6 Here are the benchmarks of the new object storage against the old one on some relevant unit tests we have:

| Test name                             | Old Real Time | Old User Time | New Real Time     | New User Time     |
|--------------------------------------|----------------|----------------|--------------------|--------------------|
| `TestClient.test_noop_submit`        | 13.061s        | 21.504s        | 14.567s (+12%)     | 27.949s (+30%)     |
| `TestClient.test_noop_map`           | 9.725s         | 20.125s        | 11.959s (+23%)     | 22.202s (+10%)     |
| `TestClient.test_heavy_function`     | 13.178s        | 23.425s        | 15.835s (+20%)     | 32.524s (+39%)     |
| `TestClient.test_very_large_payload` | 7.484s         | 3.984s         | 4.130s (−45%)      | 2.038s (−49%)      |
| `TestGraph.test_graph`               | 3.523s         | 1.620s         | 1.827s (−48%)      | 1.401s (−14%)      |

As expected, the new storage is faster on large payload, but slower on small tasks. That's kind of expected, as the client uses a synchronized connector to the storage server, which adds a little bit of latency.

gxuu:
WRT the linter test failure:
The test failed because previously the directory path coded in the load directory script
assumes user who call this script is running in the environment such that the scaler has
been packaged. This means that if a user download the package using pip and run scaler
entry points, the dynamic library will be successfully load. It also means that when devs
perform testing to the package, the lib will not be found, since at the testing point such
directory does not exist.

After PR #129, this will no longer be a problem. User (aka. You) will be able to import the
object storage server directly, both not needing to worry about library being not found.

scaler/cluster/combo.py
```diff
@@ -56,13 +59,23 @@ def __init__(
         else:
             self._address = ZMQConfig.from_string(address)
 
+        if storage_address is None:
+            self._storage_address = ObjectStorageConfig(self._address.host, get_available_tcp_port())
```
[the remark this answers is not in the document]
RaphaelJ: I first did that, but there is no guarantee that `self._address.port + 1` will be available :(

scaler/client/object_reference.py
```diff
@@ -1,14 +1,15 @@
...
 
+from scaler.utility.identifiers import ObjectID
+
 
 @dataclasses.dataclass
 class ObjectReference:
     name: bytes
-    object_id: bytes
-    size: int
+    object_id: ObjectID
```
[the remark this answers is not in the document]
RaphaelJ: It was never used, so I decided to remove it.

But I can put it back.

scaler/cluster/combo.py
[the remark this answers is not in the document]
rafa-be: I first did that, but there is no guarantee that self._address.port + 1 will be available :(

scaler/client/object_reference.py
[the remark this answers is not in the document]
rafa-be: It was never used, so I decided to remove it.

But I can put it back.

scaler/cluster/combo.py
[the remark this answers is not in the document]
RaphaelJ: Yes. Scheduler only will try to connect to `port + 1`, but combo will open a random port.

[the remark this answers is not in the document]
rafa-be: Yes. Scheduler only will try to connect to `port + 1`, but combo will open a random port.

sharpener6 (approved):

scaler/object_storage/object_storage_server.h
```diff
@@ -157,28 +157,69 @@ class ObjectStorageServer {
...
+        uint64_t value = 1;
+        ssize_t ret = write(this->on_server_ready_fd, &value, sizeof (uint64_t));
+
+        if (ret != sizeof (uint64_t)) {
+            std::cerr << "write to on_server_ready_fd failed: errno=" << errno << std::endl;
+            std::terminate();
+        }
+    }
+
+    static void setTcpNoDelay(tcp::socket& socket, bool is_no_delay) {
```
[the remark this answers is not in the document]
sharpener6: @rafa-be , please move it and then mark this as resolved

scaler/protocol/python/message.py
```diff
@@ -300,86 +323,37 @@ def instruction_type(self) -> ObjectInstructionType:
         return ObjectInstruction.ObjectInstructionType(self._msg.instructionType.raw)
 
     @property
-    def object_user(self) -> bytes:
-        return self._msg.objectUser
+    def object_user(self) -> Optional[ClientID]:
```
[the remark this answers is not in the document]
sharpener6: I will close for now, as the ClientID and SchedulerID is not merged together, which in the future we might need treat SchedulerID as normal ClientID as scheduler can connect to scheduler as client, which will acting like scheduler tree hierarchy

PR #107, opened 2025-05-13
Change monitor address to TCP as required

sharpener6:
in complicate network situation, we need change monitoring to tcp as top or webui monitoring cannot be always be in the same host as scheduler


scaler/scheduler/scheduler.py
```diff
@@ -40,11 +40,11 @@ def __init__(self, config: SchedulerConfig):
                 f"{self.__class__.__name__}: scheduler address must be tcp type: {config.address.to_address()}"
             )
 
-        self._address_monitor = ZMQConfig(
-            type=ZMQType.ipc, host=f"/tmp/{config.address.host}_{config.address.port}_monitor"
-        )
+        if config.monitor_address is None:
+            self._address_monitor = ZMQConfig(type=ZMQType.tcp, host=config.address.host, port=config.address.port + 1)
```
RaphaelJ, line 44: This should be +2 as the object storage server uses +1

rafa-be, line 44: Please use +2 as +1 is the port used by the object storage server.

rafa-be (approved):

rafa-be (approved):

PR #123, opened 2025-05-27
Transport layer redesign

gxuu:
Always read `readme.md`.

### DIFF
--- a/scaler/io/decl.cpp
+++ b/scaler/io/decl.cpp
@@ -1,0 +1,454 @@
+// -*- C++ -*-
+#include <sys/socket.h>
+
+#include <atomic>
+#include <functional>
+#include <map>
+#include <optional>
+
+/* NOTE:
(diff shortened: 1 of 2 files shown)


sharpener6 (approved):

scaler/io/build.py
```diff
@@ -0,0 +1,26 @@
+from cffi import FFI
+
+builder = FFI()
+
+with open("ymq/defs.h") as f:
+    defs = f.read()
+
+builder.cdef(defs)
+
+import os
```
sharpener6, line 10: move all imports to the head of the files

sharpener6: and who is calling this `builder.py`?

sharpener6:
please also fix the linter

scaler/io/build.py
sharpener6: This is already been resolved, look how object storage server is build and distributed, this should be the same and no difference, you don't need this build.py then

sharpener6 (approved):

PR #127, opened 2025-05-30
Create CPython Module Definition, Implement Several Classes

magniloquency:
Creates the initial CPython module definition and hooks it into the CMake build system. This module definition is a work-in-progress and is missing most implementations. Also begings to implement several C++ classes.

At the moment the generated .so file is named incorrectly and cannot be imported without being renamed. This is being investigated.

@gxuu @sharpener6


.gitignore
```diff
@@ -22,3 +22,16 @@ dask-worker-space/*
...
+CMakeCache.txt
+cmake_install.cmake
+Makefile
+CTestTestfile.cmake
+*.o
+*.so
+libgtest*
+compile_commands.json
+protocol/
+version.h
```
magniloquency, line 37: I ran cmake and make, a _lot_ of files were created. Are all of these ignores needed?

scaler/io/ymq/CMakeLists.txt
```diff
@@ -10,3 +9,8 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/scaler/io/ymq)
 
 install(TARGETS ymq
         LIBRARY DESTINATION ${CMAKE_SOURCE_DIR}/scaler/lib)
+
+find_package(Python COMPONENTS Interpreter Development.Module REQUIRED)
+
+Python_add_library(ymqpy MODULE ymqpy.cpp WITH_SOABI)
+install(TARGETS ymqpy DESTINATION ${SKBUILD_PROJECT_NAME})
```
magniloquency, line 16: This generates a file like `ymqpy.cpython-311-x86_64-linix-gnu.so`, we need it to be `ymq.cpython-...`. Is there a way to change the name?

.gitignore
sharpener6: I should not mention meson, forget about what I mentioned

gxuu: You can if you want to. It's nice to have it locally. Because users will be using different things, flags, etc., and the script will not be very helpful to them.
I maintain a very simple script locally.
> {25-05-31 2:53}% cat b.sh
> cmake -S . -B build -DCMAKE_CXX_COMPILER=$(which g++) -GNinja -DCMAKE_BUILD_TYPE=Debug
> cd build && ninja

magniloquency: Fixed. I made a simple makefile as well, let me know what you think

gxuu: I think add `Makefile` is quite confusing when `CMakeLists.txt` presents. It would be nice to just have a local build script. If you really want to push a build script up, please do not name it as `Makefile`, consider `build.sh` or `run_cmake.sh` (please do not use `cmake.sh`, as sometime this is used to generate some cmake helper macro in some projects).

scaler/io/ymq/file_descriptor.hpp
```diff
@@ -18,31 +19,49 @@
...
+    void assert_owned() const { assert(ownership == Ownership::Owned); }
+
 public:
-    ~FileDescriptor() {
+    ~FileDescriptor() noexcept(false) {
         if (auto code = close(fd) < 0)
             throw std::system_error(errno, std::system_category(), "Failed to close file descriptor");
 
         this->fd = -1;
     }
```
magniloquency, line 35: todo: only close if owned

scaler/io/ymq/ymqpy.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+    if (self->shares <= 0) {
+        // TODO: free data here?
+    }
+}
+
+static PyBufferProcs PyBytesYmqBufferProcs = {
+    .bf_getbuffer     = (getbufferproc)PyBytesYmq_getbuffer,
+    .bf_releasebuffer = (releasebufferproc)PyBytesYmq_releasebuffer,
+};
```
magniloquency, line 68: This is the Python buffer protocol, this allows us to have zero-copy between C++ and Python.

https://docs.python.org/3.11/c-api/buffer.html

scaler/io/ymq/ymqpy.cpp
magniloquency, line 1: rename to `pymod_ymq.cpp`

magniloquency:
1. Use `.h` and `.cpp` files only (no `.hpp`)
2. Python modules should be `pymod_*.cpp`

PR #129, opened 2025-05-30
Create Python module Object  Storage Server

gxuu:
Object storage server can now be imported as a Python module.

User should load object storage server module as such: `from scaler.object_storage import object_storage_server` and then call `object_storage_server.run_object_storage_server("addr", "port")`.


tests/object_storage/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+current_dir = os.path.dirname(os.path.abspath(__file__))
+pymod_path = os.path.abspath(os.path.join(current_dir, '../../build/scaler/object_storage'))
+
+sys.path.append(pymod_path)
+
+# Now try to import the module
+try:
+    import object_storage_server
+except ImportError as e:
+    print("Failed to import object_storage_server:", e)
```
sharpener6, line 14:
```python
raise ImportError(f"Failed to import object_storage_server from {pymod_path}") from e
```

gxuu: Resolved.

CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ if(NOT SKBUILD_PROJECT_NAME)
 endif()
 
 if(NOT SKBUILD_PROJECT_VERSION)
-  set(SKBUILD_PROJECT_VERSION "1.9.4")
+  set(SKBUILD_PROJECT_VERSION "1.9.9")
```
sharpener6, line 15: Please create a `version.txt` in python scaler module as sole place that defines version:
1. in `about.py`, directly read the content from same directory's `version.txt` and assign to `__version__`
2. `version.txt` need be included to wheel so it need be defined as package data in `pyproject.toml`
3. change the section `tool.scikit-build.metadata.version` in `pyproject.toml` to read from `version.txt`
4. change `CMakeLists.txt` to get version from `version.txt`

pyproject.toml
```diff
@@ -51,7 +51,7 @@ all = ["python-graphblas", "numpy==2.2.5", "uvloop", "nicegui[plotly]==2.14.1"]
 
 [tool.scikit-build]
 wheel.packages = ["scaler"]
-build.targets=["server", "ymq"]
+build.targets=["server", "ymq", "object_storage_server"]
```
sharpener6, line 54: what's this server about?

gxuu: What's this server is: The underlying object storage server component.
Why do we need it: 
1. To separate the binding logic away from the server interface. This means some component (e.g. C++ `scheduler`), will be able to directly link against the underlying server if needed.
2. Avoid unnecessary linkage. For demonstration: 
```
ldd libserver.so
        linux-vdso.so.1 (0x00007207bf8b0000)
        libcapnp.so.1.1.0 => /usr/lib/libcapnp.so.1.1.0 (0x00007207bf70b000)
        ...
```
Will not contain `libpython` as part of its output. 

Why do we need it to present in the `toml`? Because we still need to build it in order for `"object_storage_server"` to link against it.

gxuu: Perhaps, a better way to name this "server" target is "internal_object_storage_server" or "object_storage_server_base".

CMakeLists.txt
sharpener6: read my comments carefully and do not invent new ways

gxuu: Resolved per your instruction. Please check.

pyproject.toml
sharpener6: Can I know why it cannot be combined to 1?

scaler/about.py
```diff
@@ -1 +1,5 @@
-__version__ = "1.9.9"
+try:
+    with open("version.txt", "r") as f:
+        __version__ = f.read().strip()
+except FileNotFoundError as e:
```
sharpener6, line 4: this try and except does nothing but repeat exception message, just get rid of try except

pyproject.toml
gxuu: And we don't want a thing to link against Python unless it needs to be linked against Python.

rafa-be: Maybe the naming is incorrect then? 

I'm not sure I understand all of this, but one could be something like `object-server` and the Python binding `python-object-server`. In any case, `server` is very misleading, as this is clearly not the only server lib/executable we have.

CMakeLists.txt
```diff
@@ -11,14 +11,24 @@ if(NOT SKBUILD_PROJECT_NAME)
   set(SKBUILD_PROJECT_NAME "scaler")
 endif()
 
-if(NOT SKBUILD_PROJECT_VERSION)
-  set(SKBUILD_PROJECT_VERSION "1.9.4")
```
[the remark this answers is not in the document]
gxuu: There are no reason that I cannot set the `SKBUILD_PROJECT_VERSION` variable from the text file. It is just that I feel like setting `PROJECT_VERSION` is clearer, as it is not always an `SKBUILD`.
Do you want me to change the `PROJECT_VERSION` variable to `SKBUILD_PROJECT_VERSION`?

tests/object_storage/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+current_dir = os.path.dirname(os.path.abspath(__file__))
+pymod_path = os.path.abspath(os.path.join(current_dir, '../../build/scaler/object_storage'))
+
+sys.path.append(pymod_path)
+
+# Now try to import the module
+try:
+    import object_storage_server
+except ImportError as e:
+    raise ImportError(f"Failed to import object_storage_server from {pymod_path}") from e
```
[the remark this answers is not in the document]
gxuu: Because the generated dynamic library is not present when running tests. So `scaler/object_storage`, when test is running, will not contain desired library.

pyproject.toml
gxuu: @rafa-be  various naming changes have been made.

CMakeLists.txt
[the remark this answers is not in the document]
rafa-be: Oh I see. I thought you had to change it to `PROJECT_VERSION`, but if you think `PROJECT_VERSION` is good and does not throw the warning when building, then that's perfect for me.

pyproject.toml
rafa-be: Yeah that's better now, thanks.

scripts/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+if [[ "$1" != "Debug" && "$1" != "Release" ]]; then
```
sharpener6, line 3: remove this option, always put as Debug

scripts/build.sh
```diff
@@ -0,0 +1,13 @@
...
+    echo "Usage: ./scripts/build.sh [Debug|Release]"
+    exit 1
+fi
+
+BUILD_TYPE="$1"
+cmake -S . -B build -DCMAKE_BUILD_TYPE=${BUILD_TYPE}
+cd build && make
+cp scaler/io/ymq/libymq.so ../scaler/io/ymq/.
```
sharpener6, line 11: do not copy explicitly

scaler/object_storage/pymod_object_storage_server.cpp
```diff
@@ -0,0 +1,38 @@
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+
+#include "server_interface.h"
```
sharpener6, line 4:
```c++
#include "object_storage_server.h"
```

scaler/object_storage/pymod_object_storage_server.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+extern "C" {
+static PyObject* object_storage_server_run_object_storage_server(PyObject* self, PyObject* args) {
```
sharpener6, line 7:
```C++
static PyObject* run_object_storage(PyObject* self, PyObject* args)
```

scripts/build.sh
gxuu: resolved

tests/object_storage/CMakeLists.txt
```diff
@@ -33,7 +33,8 @@ add_executable(
   test_scaler_object
   test_scaler_object.cpp
   ${PROJECT_SOURCE_DIR}/scaler/object_storage/io_helper.cpp
-  ${PROJECT_SOURCE_DIR}/scaler/object_storage/server.cpp
+  # ${PROJECT_SOURCE_DIR}/scaler/object_storage/object_storage_server.cpp
```
sharpener6, line 36: why comment out instead of removing?

scaler/object_storage/object_storage_server.h
```diff
@@ -167,5 +168,33 @@ class ObjectStorageServer {
...
+    auto executor = co_await boost::asio::this_coro::executor;
+    tcp::acceptor acceptor(executor, endpoint);
+    for (;;) {
+        auto shared_socket = std::make_shared<tcp::socket>(executor);
+        co_await acceptor.async_accept(*shared_socket, use_awaitable);
+        co_spawn(executor, server.process_request(std::move(shared_socket)), detached);
+    }
+}
+
+inline void run_internal_object_storage_server(ObjectStorageServer& server, std::string name, std::string port) {
```
sharpener6, line 181: this function `run_internal_object_storage_server` should be `ObjectStorageServer.run()`

scaler/object_storage/object_storage_server.h
```diff
@@ -167,5 +168,33 @@ class ObjectStorageServer {
     }
 };
 
+inline awaitable<void> listener(ObjectStorageServer& server, boost::asio::ip::tcp::endpoint endpoint) {
```
sharpener6, line 171: this function should be private function of class ObjectStorageServer

PR #134, opened 2025-06-02
webui: cap displayed cpu usage to 100

cloakedclock:
CPU usage reported by the scheduler sometimes exceeds 100 (eg. 100.4). This PR caps the displayed cpu value to 100 for better aesthetics.


sharpener6:
100.4% is normal

PR #158, opened 2025-06-25
YMQ

magniloquency:


scaler/io/ymq/pymod_ymq/exception.h
```diff
@@ -0,0 +1,109 @@
...
+
+YMQException* YMQException_fromCoreException(YMQState* state, Error* error) {
+    PyObject* code = PyLong_FromLong((long)error->_errorCode);
+
+    if (!code)
+        return nullptr;
+
+    PyObject* message = PyUnicode_FromString(error->what());
+
+    if (!message) {
```
gxuu, line 89: Should there be some form of logging message stating that unexpected things have happened? Notice that `error->what` might be empty but is never null.

scaler/io/ymq/pymod_ymq/io_context.h
```diff
@@ -130,7 +133,11 @@ static PyObject* PyIOContext_createIOSocket(
         return nullptr;
     }
 
-    ioSocket->socket = self->ioContext->createIOSocket(identity, socketType);
+    auto createSocketPromise = std::make_shared<std::promise<void>>();
```
gxuu, line 136: IIRC this was a tmp fix from me. My whole intention was to make the code compiles, so this probably needs some refinement?

scaler/io/ymq/pymod_ymq/io_socket.h
```diff
@@ -29,17 +37,298 @@ static void PyIOSocket_dealloc(PyIOSocket* self) {
...
-
-    // in this function we need to:
-    // 1. create an asyncio future (easy)
-    //    - should be pretty easy, just call standard methods
-    // 2. create a handle to that future that can be completed in a C++ callback (harder)
-    //    - how do we call Python from non-Python threads? where is the handle stored?
-    // 3. await the future (very hard)
-    //    - how do you await in a C extension module? is it even possible?
-    //    - might need to call into Python code to make this work
+    PyMessage* message;
```
gxuu, line 40: Be sure to initialize variable. In this case initialize it to `nullptr`. I would prefer `PyMessage* message{}`.

scaler/io/ymq/pymod_ymq/io_socket.h
```diff
@@ -29,17 +37,298 @@ static void PyIOSocket_dealloc(PyIOSocket* self) {
...
+}
+
+static PyObject* PyIOSocket_send_sync(PyIOSocket* self, PyObject* args, PyObject* kwargs) {
+    PyMessage* message;
+    const char* kwlist[] = {"message", nullptr};
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O", (char**)kwlist, &message)) {
+        Py_RETURN_NONE;
+    }
+
+    std::binary_semaphore sem(0);
```
gxuu, line 68: I don't understand the usage of `binary_semaphore` here. It appears to me to be equivalent to a `std::mutex`. If that's the case, use a `mutex`. Consider using higher level synchronization utility such as `std::latch` if possible.

scaler/io/ymq/pymod_ymq/io_socket.h
```diff
@@ -29,17 +37,298 @@ static void PyIOSocket_dealloc(PyIOSocket* self) {
...
+                //         future, [state, error]() { return YMQException_fromCoreException(state, &*error); });
+                // } else {
+                future_set_result(future, []() { Py_RETURN_NONE; });
+                // }
+            });
+    });
+}
+
+static PyObject* PyIOSocket_send_sync(PyIOSocket* self, PyObject* args, PyObject* kwargs) {
+    PyMessage* message;
```
gxuu, line 62: Same here.

scaler/io/ymq/pymod_ymq/io_socket.h
```diff
@@ -29,17 +37,298 @@ static void PyIOSocket_dealloc(PyIOSocket* self) {
...
+        Py_RETURN_NONE;
+    }
+
+    std::binary_semaphore sem(0);
+    int error = 0;
+
+    self->socket->sendMessage(
+        {.address = message->address ? std::move(message->address->bytes) : Bytes::empty(),
+         .payload = std::move(message->payload->bytes)},
+        [&](int _error) {
```
gxuu, line 74: Do not prefix argument with `_`, they are mostly used when declaring member variables in the current guideline.

If it is just an error, might as well name it `e`.

scaler/io/ymq/pymod_ymq/io_socket.h
```diff
@@ -29,17 +37,298 @@ static void PyIOSocket_dealloc(PyIOSocket* self) {
...
+        Py_RETURN_NONE;
+    }
+
+    PyBytesYMQ* payload = (PyBytesYMQ*)PyObject_CallNoArgs(state->PyBytesYMQType);
+    if (!payload) {
+        Py_DECREF(address);
+        Py_RETURN_NONE;
+    }
+
+    address->bytes = std::move(message.address);
```
gxuu, line 158: Give a better name than just `address`. At first glance I thought it's a bug.

scaler/io/ymq/pymod_ymq/io_socket.h
```diff
@@ -29,17 +37,298 @@ static void PyIOSocket_dealloc(PyIOSocket* self) {
...
+        sem.release();
+    });
+
+    printf("Waiting for bind to complete...\n");
+
+    // block the thread until the callback is called
+    try {
+        sem.acquire();
+    } catch (const std::exception& e) {
+        printf("Exception while waiting for bind: %s\n", e.what());
```
gxuu, line 244: Goes to `stderr`. Also logging message of any kind should be using the same format as `scaler` uses without `[level]` prefixed.

scaler/io/ymq/pymod_ymq/async.h
```diff
@@ -0,0 +1,85 @@
...
+#include <functional>
+
+// First-party
+#include "scaler/io/ymq/pymod_ymq/ymq.h"
+
+// Wraps an async callback that accepts a Python asyncio future
+static PyObject* async_wrapper(PyObject* self, std::function<void(YMQState* state, PyObject* future)> callback) {
+    // replace with PyType_GetModuleByDef(Py_TYPE(self), &ymq_module) in a newer Python version
+    // https://docs.python.org/3/c-api/type.html#c.PyType_GetModuleByDef
+    PyObject* module = PyType_GetModule(Py_TYPE(self));
```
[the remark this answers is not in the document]
magniloquency: renamed all uses to `pyModule`

scaler/io/ymq/bytes.h
```diff
@@ -77,51 +84,23 @@ class Bytes {
...
-    static Bytes empty() { return Bytes {NULL, 0, Owned}; }
-
-    static Bytes copy(const uint8_t* m_data, size_t m_len) {
-        if (m_len == 0)
-            return empty();
-
-        return Bytes {datadup(m_data, m_len), m_len, Owned};
+    // NOTE: Below two functions are not used by the core but appears
+    // to be used by pymod YMQ. - gxu
+    [[nodiscard]] static Bytes empty() { return Bytes {(uint8_t*)nullptr, 0}; }
```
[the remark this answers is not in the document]
magniloquency: I think these changes are inherited from your pr. feel free to remove it there and i'll handle the merge

scaler/io/ymq/pymod_ymq/exception.h
```diff
@@ -0,0 +1,109 @@
...
+// Python
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+#include <structmember.h>
+
+// First-party
+#include "scaler/io/ymq/pymod_ymq/ymq.h"
+#include "ymq.h"
+
+typedef struct {
```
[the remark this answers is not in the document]
magniloquency: done

scaler/io/ymq/pymod_ymq/exception.h
```diff
@@ -0,0 +1,109 @@
...
+    PyException_HEAD;
+} YMQException;
+
+extern "C" {
+
+static int YMQException_init(YMQException* self, PyObject* args, PyObject* kwds) {
+    // check the args
+    PyObject *code = nullptr, *message = nullptr;
```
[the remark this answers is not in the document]
magniloquency: done

scaler/io/ymq/pymod_ymq/exception.h
```diff
@@ -0,0 +1,109 @@
...
+    {"code", (getter)YMQException_code_getter, nullptr, PyDoc_STR("error code"), nullptr},
+    {"message", (getter)YMQException_message_getter, nullptr, PyDoc_STR("error message"), nullptr},
+    {nullptr}  // Sentinel
+};
+
+static PyType_Slot YMQException_slots[] = {
+    {Py_tp_init, (void*)YMQException_init},
+    {Py_tp_dealloc, (void*)YMQException_dealloc},
+    {Py_tp_getset, (void*)YMQException_getset},
+    {0, 0}};
```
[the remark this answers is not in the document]
magniloquency: done

scaler/io/ymq/pymod_ymq/exception.h
magniloquency: > Should there be some form of logging
not in the interface, imo. we should return errors to the user('s python code) and let them decide what to do with it

most python library functions will return `NULL` or `-1` to indicate that an error occurred, and they'll set an exception if that happens. in that case our only responsibility is to exit the function; the python rt will pick up the exc for us. you can read about it here if you would like: https://docs.python.org/3/c-api/exceptions.html

scaler/io/ymq/pymod_ymq/io_socket.h
magniloquency: good catch. fixed

scaler/io/ymq/pymod_ymq/io_socket.h
magniloquency: done

scaler/io/ymq/pymod_ymq/io_socket.h
magniloquency: oh i left that in there by accident. i have removed it. the interface should not be printing anything

scaler/io/ymq/pymod_ymq/io_socket.h
magniloquency: i think semaphores are well-suited for making barriers. a mutex can be used for this purpose, but they're usually more expensive. anyway, it looks like `std::latch` is actually even better for this, it's like `std::barrier` except one-shot, which is exactly what we need. the benefit of a constrained implementation like this is that it opens up ways for it to be implemented more efficiently. anyway, i've swapped this out for `std::latch`

scaler/io/ymq/pymod_ymq/io_context.h
magniloquency: fixed, but it may change again after a pending discussion

scaler/io/ymq/pymod_ymq/async.h
```diff
@@ -0,0 +1,90 @@
...
+    // async
+    callback(state, future);
+
+    return PyObject_CallFunction(state->PyAwaitableType, "O", future);
+}
+
+struct Awaitable {
+    PyObject_HEAD;
+    PyObject* future;
+};
```
rafa-be, line 56: Couldn't this struct name be too common and conflict in the future with some library we might link with YMQ or Scaler?

scaler/io/ymq/pymod_ymq/ymq.h
```diff
@@ -5,45 +5,110 @@
...
+    PyObject* PyMessageType;       // Reference to the Message type
+    PyObject* PyIOSocketType;      // Reference to the IOSocket type
+    PyObject* PyIOContextType;     // Reference to the IOContext type
+    PyObject* PyExceptionType;     // Reference to the Exception type
+    PyObject* PyAwaitableType;     // Reference to the Awaitable type
+};
+
+// this function must be called from a C++ thread
+// it locks the GIL and completes a future
+static void future_do(PyObject* future, std::function<PyObject*()> fn, const char* future_method) {
```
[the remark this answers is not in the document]
rafa-be: Oh I see. Can you update the comment to make it clear that the `std::function` will be called with the GIL acquired?

scaler/io/ymq/pymod_ymq/io_socket.h
```diff
@@ -6,73 +6,381 @@
...
+#include "scaler/io/ymq/pymod_ymq/async.h"
+#include "scaler/io/ymq/pymod_ymq/bytes.h"
+#include "scaler/io/ymq/pymod_ymq/message.h"
 #include "scaler/io/ymq/pymod_ymq/ymq.h"
 
+struct PyIOSocketInner {
+    std::shared_ptr<IOSocket> socket;
+    std::shared_ptr<IOContext> ioContext;
+};
+
```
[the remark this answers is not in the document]
magniloquency: I tested this and it seems to work :) last night I was getting a lot of segmentation faults, but I'm not seeing that now with this. I've removed the `*Inner` structs and am using placement new, now. Thank you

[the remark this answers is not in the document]
rafa-be: Great! That makes the implementation simpler :)

scaler/io/ymq/pymod_ymq/io_socket.h
gxuu: `address` looks pretty much to me as just a `Bytes`, and that's the typical usage in the core. That's why it looks confusing to me when you write `address->bytes`. That's a minor part, you can leave it there.

scaler/io/ymq/pymod_ymq/exception.h
```diff
@@ -0,0 +1,111 @@
...
+    // delegate to the base class init
+    return self->ob_base.ob_type->tp_base->tp_init((PyObject*)self, args, kwds);
+}
+
+static void YMQException_dealloc(YMQException* self) {
+    Py_TYPE(self)->tp_free((PyObject*)self);
+}
+
+static PyObject* YMQException_code_getter(YMQException* self, void* Py_UNUSED(closure)) {
+    return PySequence_GetItem(self->args, 0);  // code is the first item in args
```
gxuu, line 59: Give a name to the second argument instead of using 0. For example, `indexOfExceptionCode`.

scaler/io/ymq/pymod_ymq/exception.h
```diff
@@ -0,0 +1,111 @@
...
+}
+
+static PyObject* YMQException_message_getter(YMQException* self, void* Py_UNUSED(closure)) {
+    return PySequence_GetItem(self->args, 1);  // message is the second item in args
```
gxuu, line 63: Same here.

scaler/io/ymq/pymod_ymq/ymq.h
```diff
@@ -108,33 +173,121 @@ static int ymq_createIOSocketTypeEnum(PyObject* module) {
...
+        PyErr_SetString(PyExc_RuntimeError, "Failed to get iterator for Error enum");
+        return -1;
+    }
+
+    // is this the best way to add a method to each enum item?
+    // in python you can just write: MyEnum.new_method = ...
+    // for some reason this does not seem to work with the c api
+    // docs and examples are unfortunately scarce for this
+    // for now this will work just fine
+    PyObject* item;
```
gxuu, line 241: `item{}`.

scaler/io/ymq/pymod_ymq/io_context.h
```diff
@@ -124,17 +130,27 @@ static PyObject* PyIOContext_createIOSocket(
...
+    PyIOSocket* ioSocket = PyObject_New(PyIOSocket, (PyTypeObject*)state->PyIOSocketType);
     if (!ioSocket) {
         PyErr_SetString(PyExc_RuntimeError, "Failed to create IOSocket instance");
         return nullptr;
     }
 
-    ioSocket->socket = self->ioContext->createIOSocket(identity, socketType);
-    return (PyObject*)ioSocket;
+    // ensure the fields are init
+    new (&ioSocket->socket) std::shared_ptr<IOSocket>();
```
gxuu, line 140: If the space of `ioSocket` has been allocated, why do you still need this?

scaler/io/ymq/pymod_ymq/ymq.h
```diff
@@ -5,45 +5,110 @@
...
+// First-party
+#include "scaler/io/ymq/error.h"
+
+struct YMQState {
+    PyObject* enumModule;     // Reference to the enum module
+    PyObject* asyncioModule;  // Reference to the asyncio module
+
+    PyObject* PyIOSocketEnumType;  // Reference to the IOSocketType enum
+    PyObject* PyErrorCodeType;     // Reference to the Error enum
+    PyObject* PyBytesYMQType;      // Reference to the PyBytesYMQ type
```
gxuu, line 24: Would `PyBytesType` be a better name?

scaler/io/ymq/pymod_ymq/ymq.h
```diff
@@ -108,33 +173,121 @@ static int ymq_createIOSocketTypeEnum(PyObject* module) {
...
+}
+
+static int ymq_createType(
+    PyObject* pyModule,
+    PyObject** storage,
+    PyType_Spec* spec,
+    const char* name,
+    bool add        = true,
+    PyObject* bases = nullptr) {
+    *storage = PyType_FromModuleAndSpec(pyModule, spec, bases);
```
gxuu, line 272: Who calls this function? Is `storage != nullptr` invariant? Add assert before this statement.

magniloquency: it's an internal helper function called just below in `ymq_exec()`. I know that `storage` isn't null, but I can add the assert anyway if you would like?

scaler/io/ymq/pymod_ymq/ymq.h
magniloquency: it sounds too similar to the python builtin, imo: https://docs.python.org/3/c-api/bytes.html#c.PyBytes_Type

magniloquency: `PyBytes_Type`

scaler/io/ymq/pymod_ymq/exception.h
magniloquency: done

scaler/io/ymq/pymod_ymq/async.h
magniloquency: i tried to put them all inside `scaler::ymq`, but it caused _a lot_ of compiler errors for some reason 🤔 more investigation needed

scaler/io/ymq/pymod_ymq/io_context.h
magniloquency: because it's been allocated but not initialized, we don't know how python allocates its memory, but i think it's a good idea to make sure it's initialized from a c++ perspective here to avoid any issues. i'm calling `~shared_ptr()` later on so i need to make sure it's init with something

scaler/io/ymq/pymod_ymq/bytes.h
```diff
@@ -52,56 +54,56 @@ static int PyBytesYmq_init(PyBytesYmq* self, PyObject* args, PyObject* kwds) {
...
 }
 
-static void PyBytesYmq_dealloc(PyBytesYmq* self) {
+static void PyBytesYMQ_dealloc(PyBytesYMQ* self) {
     self->bytes.~Bytes();  // Call the destructor of Bytes
     Py_TYPE(self)->tp_free(self);
 }
 
-static PyObject* PyBytesYmq_repr(PyBytesYmq* self) {
+static PyObject* PyBytesYMQ_repr(PyBytesYMQ* self) {
```
rafa-be, line 62:
```suggestion
static PyObject* PyBytesYMQ_repr(const PyBytesYMQ* self) {
```

scaler/io/ymq/pymod_ymq/bytes.h
```diff
@@ -52,56 +54,56 @@ static int PyBytesYmq_init(PyBytesYmq* self, PyObject* args, PyObject* kwds) {
...
     }
 }
 
-static PyObject* PyBytesYmq_data_getter(PyBytesYmq* self) {
+static PyObject* PyBytesYMQ_data_getter(PyBytesYMQ* self) {
     return PyBytes_FromStringAndSize((const char*)self->bytes.data(), self->bytes.len());
 }
 
-static PyObject* PyBytesYmq_len_getter(PyBytesYmq* self) {
+static PyObject* PyBytesYMQ_len_getter(PyBytesYMQ* self) {
```
rafa-be, line 74:
```suggestion
static PyObject* PyBytesYMQ_len_getter(const PyBytesYMQ* self) {
```

scaler/io/ymq/pymod_ymq/bytes.h
```diff
@@ -52,56 +54,56 @@ static int PyBytesYmq_init(PyBytesYmq* self, PyObject* args, PyObject* kwds) {
...
+static PyObject* PyBytesYMQ_repr(PyBytesYMQ* self) {
     if (self->bytes.is_empty()) {
         return PyUnicode_FromString("<Bytes: empty>");
     } else {
         return PyUnicode_FromFormat("<Bytes: %db>", self->bytes.len());
     }
 }
 
-static PyObject* PyBytesYmq_data_getter(PyBytesYmq* self) {
+static PyObject* PyBytesYMQ_data_getter(PyBytesYMQ* self) {
```
rafa-be, line 70:
```suggestion
static PyObject* PyBytesYMQ_data_getter(const PyBytesYMQ* self) {
```

scaler/io/ymq/pymod_ymq/bytes.h
rafa-be: (if CPython allows it)

scaler/io/ymq/pymod_ymq/ymq.h
```diff
@@ -5,45 +5,110 @@
...
+    PyObject* PyMessageType;       // Reference to the Message type
+    PyObject* PyIOSocketType;      // Reference to the IOSocket type
+    PyObject* PyIOContextType;     // Reference to the IOContext type
+    PyObject* PyExceptionType;     // Reference to the Exception type
+    PyObject* PyAwaitableType;     // Reference to the Awaitable type
+};
+
+// this function must be called from a C++ thread
+// this function will lock the GIL, call `fn()` and use its return value to set the future's result/exception
+static void future_do(PyObject* future, std::function<PyObject*()> fn, const char* future_method) {
```
rafa-be, line 34:
```suggestion
static void future_do(PyObject* future, const std::function<PyObject*()>& fn, const char* future_method) {
```

scaler/io/ymq/pymod_ymq/bytes.h
magniloquency: It probably would compile, but the Python API specifies it as non-const, so I think it's better to leave it that way?

https://docs.python.org/3/c-api/typeobj.html#c.reprfunc

scaler/io/ymq/pymod_ymq/ymq.h
magniloquency: Done

.github/workflows/linter.yml
```diff
@@ -34,11 +34,12 @@ jobs:
           sudo apt update -y
           sudo apt upgrade -y
           sudo DEBIAN_FRONTEND=noninteractive sudo apt install -y tzdata
-          sudo apt install cmake clang curl pkg-config -y
+          sudo apt install cmake clang curl pkg-config g++-14 -y
 
       - name: Make scripts executable
         run: |
           sudo chmod 755 ./scripts/download_install_dependencies.sh
+          sudo chmod 755 ./scripts/build.sh
```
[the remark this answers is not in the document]
gxuu: It is just to make sure we are not running into problems when executing scripts. We can remove it.

scaler/io/ymq/bytes.h
```diff
@@ -10,64 +14,67 @@
...
+        this->_len  = other._len;
+    }
+
+    Bytes& operator=(const Bytes& other) noexcept {
+        Bytes tmp(other);
+        swap(*this, tmp);
+        return *this;
+    }
+
+    friend void swap(Bytes& x, Bytes& y) noexcept {
```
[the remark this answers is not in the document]
sharpener6: Or are you going to address in next PR to better design the classes and remove the usage of friend?

scaler/io/ymq/pymod_ymq/bytes.h
rafa-be: Yeah, leave it that way then

scaler/io/ymq/pymod_ymq/io_context.h
```diff
@@ -45,13 +48,14 @@ static int PyIOContext_init(PyIOContext* self, PyObject* args, PyObject* kwds) {
...
+
     return 0;
 }
 
 static void PyIOContext_dealloc(PyIOContext* self) {
-    // Python knows nothing about C++, so we need to manually call destructors
-    self->ioContext.~shared_ptr();            // Call the destructor of shared_ptr
+    self->ioContext.~shared_ptr();
     Py_TYPE(self)->tp_free((PyObject*)self);  // Free the PyObject
 }
```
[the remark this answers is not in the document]
gxuu: Any object in YMQ should outlive everything that's inside them (see #168) for what I mean by "inside them". The assumption is that the user is responsible for guaranteeing that.

User should be able to guarantee that `IOContext` outlives `IOSocket` pretty trivially. As they would call `IOContext::removeIOSocket` when they want to destroy an `IOSocket`. This should also happen when the program is about to end. Calling `removeIOSocket` is thread safe and synchronized.

Now, why it's not a good idea to use `shared_ptr` to guarantee the lifetime of `IOContext`? This is because we don't want the `IOContext` to destroy in a thread that does not construct the object. In particular, `IOContext` should not be deleted in the event loop thread, as that would cause circular reference problem.

scaler/io/ymq/message_connection_tcp.cpp
```diff
@@ -80,7 +80,8 @@ void MessageConnectionTCP::onCreated() {
     if (_connFd != 0) {
         this->_eventLoopThread->_eventLoop.addFdToLoop(
             _connFd, EPOLLIN | EPOLLOUT | EPOLLET, this->_eventManager.get());
-        _writeOperations.emplace_back(Bytes {_localIOSocketIdentity.data(), _localIOSocketIdentity.size()}, [](int) {});
+        _writeOperations.emplace_back(
+            Bytes::copy((uint8_t*)_localIOSocketIdentity.data(), _localIOSocketIdentity.size()), [](int) {});
```
gxuu, line 84: Please refrain from changing the core; It's better to do it in a separate PR, given the title of this PR is "Introduce YMQ's Python Interface".

It also will causes more trouble when merging. Since the callback types are now changed.

scaler/io/ymq/message_connection_tcp.cpp
```diff
@@ -174,7 +175,7 @@ std::expected<void, int> MessageConnectionTCP::tryReadMessages(bool readOneMessa
 void MessageConnectionTCP::updateReadOperation() {
     while (_pendingRecvMessageCallbacks->size() && _receivedReadOperations.size()) {
         if (isCompleteMessage(_receivedReadOperations.front())) {
-            Bytes address(_remoteIOSocketIdentity->data(), _remoteIOSocketIdentity->size());
+            Bytes address = Bytes::copy((uint8_t*)_remoteIOSocketIdentity->data(), _remoteIOSocketIdentity->size());
```
gxuu, line 178: Same here. Especially do NOT use `Bytes::copy` in the core; Examples' code changes are fine.

The plan was to remove these two functions, as you can see in the master branch
```
    // NOTE: Below two functions are not used by the core but appears
    // to be used by pymod YMQ. - gxu
    [[nodiscard]] static Bytes empty() { return Bytes {(uint8_t*)nullptr, 0}; }
    [[nodiscard]] static Bytes copy(const uint8_t* m_data, size_t m_len) {
        Bytes result;
        result._data = datadup(m_data, m_len);
        result._len  = m_len;
        return result;
    }

```

scaler/io/ymq/pymod_ymq/exception.h
```diff
@@ -0,0 +1,115 @@
...
+#include "pyport.h"
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+#include <structmember.h>
+
+// First-party
+#include "scaler/io/ymq/pymod_ymq/ymq.h"
+#include "ymq.h"
+
+// the order of the members in the exception args tuple
```
gxuu, line 13: If there's no plan of adding namespace in this PR, please prefix these two variables' name by some identifier. For example `PYMOD_YMQ_errorCodeIndex`. Or, declare them as `static constexpr` in each function that's using them.

sharpener6:
@magniloquency, please do what I asked, rebase your commits, otherwise I will decline this PR

sharpener6:
> Please do not remove the `Bytes` copy constructor. Move it to a separate PR.

@magniloquency , please revert this change

scaler/io/ymq/examples/pub_server.cpp
```diff
@@ -26,8 +26,8 @@ int main() {
         std::string payload("Hello from the publisher");
 
         Message publishContent;
-        publishContent.address = Bytes(address.data(), address.size());
-        publishContent.payload = Bytes(payload.data(), payload.size());
+        publishContent.address = Bytes::copy((uint8_t*)address.data(), address.size());
```
sharpener6, line 29: please revert this back

magniloquency: Done

scaler/io/ymq/bytes.h
```diff
@@ -23,13 +23,12 @@ class Bytes {
...
             return;
         delete[] _data;
+        _data = nullptr;
     }
 
-    explicit Bytes(uint8_t* m_data, size_t m_len): _data(m_data), _len(m_len) {}
+    explicit Bytes(uint8_t* data, size_t len): _data(data), _len(len) {}
 
 public:
-    Bytes(char* data, size_t len): _data(datadup((uint8_t*)data, len)), _len(len) {}
```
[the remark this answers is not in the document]
magniloquency: Done

scaler/io/ymq/pymod_ymq/exception.h
magniloquency: Done

scaler/io/ymq/message_connection_tcp.cpp
magniloquency: Restored

gxuu (commented):
Add `const` to passed-in arguments that are not needed to be changed.

scaler/io/ymq/message_connection_tcp.cpp
```diff
@@ -174,7 +174,7 @@ std::expected<void, int> MessageConnectionTCP::tryReadMessages(bool readOneMessa
 void MessageConnectionTCP::updateReadOperation() {
     while (_pendingRecvMessageCallbacks->size() && _receivedReadOperations.size()) {
         if (isCompleteMessage(_receivedReadOperations.front())) {
-            Bytes address(_remoteIOSocketIdentity->data(), _remoteIOSocketIdentity->size());
+            Bytes address = Bytes(_remoteIOSocketIdentity->data(), _remoteIOSocketIdentity->size());
```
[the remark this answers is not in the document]
magniloquency: Done

scaler/io/ymq/pymod_ymq/io_socket.h
```diff
@@ -6,42 +6,339 @@
...
+    if (error) {
+        PyErr_SetString(PyExc_RuntimeError, "Failed to bind to address synchronously");
+        return nullptr;
+    }
+
+    Py_RETURN_NONE;
+}
+
+static PyObject* PyIOSocket_connect(PyIOSocket* self, PyObject* args, PyObject* kwargs) {
+    PyObject* addressObj;
```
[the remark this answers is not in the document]
magniloquency: done

scaler/io/ymq/ymq_test.py
```diff
@@ -1,10 +1,19 @@
...
 
-assert context.num_threads == 2
-assert socket.identity == "my-socket"
-assert socket.socket_type == ymq.IOSocketType.Dealer
+    assert socket.identity == "ident"
+    assert socket.socket_type == ymq.IOSocketType.Binder
+
+    exc = ymq.YMQException(ymq.ErrorCode.InvalidAddressFormat, "oh no!")
+    assert exc.code == ymq.ErrorCode.InvalidAddressFormat
+    assert exc.message == "oh no!"
```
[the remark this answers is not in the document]
magniloquency: Done. We should probably move this file somewhere else and include it in the ci, to be honest

scaler/io/ymq/pymod_ymq/message.h
```diff
@@ -39,29 +39,29 @@ static int PyMessage_init(PyMessage* self, PyObject* args, PyObject* kwds) {
...
             return -1;
         }
     }
 
-    self->address = (PyBytesYmq*)address;
-    self->payload = (PyBytesYmq*)payload;
+    self->address = (PyBytesYMQ*)address;
+    self->payload = (PyBytesYMQ*)payload;
 
     return 0;  // todo
```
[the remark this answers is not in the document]
magniloquency: good point. in this case the todo is actually done, so i have removed it :)

scaler/io/ymq/bytes.h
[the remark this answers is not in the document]
magniloquency: I removed these fns

gxuu:
Given that namespace support and the error handling mechanism (including the callback type modification) are already in the core, merging this PR without the changes (especially the later one) will not work. IMHO, these two changes can be made in another PR. Please clarify if this PR is good to merge if you put the error handling integration and namespace to next PR.

scaler/io/ymq/epoll_context.h
```diff
@@ -72,7 +72,6 @@ class EpollContext {
     InterruptiveConcurrentQueue<Function> _interruptiveFunctions;
     static const size_t _isInterruptiveFd = 0;
     static const size_t _isTimingFd       = 1;
-    static const size_t _reventSize       = 1024;
```
magniloquency, line 75: For an unknown reason (@gxuu and I had a call and could not determine the cause), when this constant is defined, the compiled `.so` fails to import with: `ImportError: /home/.../work/scaler/scaler/io/ymq/libcc_ymq.so: undefined symbol: _ZN6scaler3ymq12EpollContext11_reventSizeE`

Therefore I have removed it for now and written the value in the places it was used. We can try to bring it back later.

scaler/io/ymq/error.h
```diff
@@ -82,8 +82,8 @@ struct Error: std::exception {
 
     constexpr const char* what() const noexcept override { return _logMsg.c_str(); }
 
-    const ErrorCode _errorCode;
-    const std::string _logMsg;
+    ErrorCode _errorCode;
+    std::string _logMsg;
```
magniloquency, line 86: Note: if these are const the type cannot be moved or copied

scaler/io/ymq/epoll_context.h
magniloquency: Nevermind, this has been fixed

magniloquency: We had to mark these `constexpr`

scaler/io/ymq/pymod_ymq/ymq.h
```diff
@@ -78,64 +153,160 @@ static int ymq_createIntEnum(PyObject* module, std::string enumName, std::vector
...
+}
+
+// IDEA: CREATE AN INT ENUM AND ATTACH METHOD AFTERWARDS
+// OR: CREATE A NON-INT ENUM AND USE A TUPLE FOR THE VALUES
+static int ymq_createErrorCodeEnum(PyObject* pyModule, YMQState* state) {
+    std::vector<std::pair<std::string, int>> errorCodeValues = {
+        {"Uninit", (int)Error::ErrorCode::Uninit},
+        {"InvalidPortFormat", (int)Error::ErrorCode::InvalidPortFormat},
+        {"InvalidAddressFormat", (int)Error::ErrorCode::InvalidAddressFormat},
+        {"ConfigurationError", (int)Error::ErrorCode::ConfigurationError},
```
gxuu, line 226: The list has been updated in the core.

scaler/io/ymq/epoll_context.h
```diff
@@ -70,9 +70,9 @@ class EpollContext {
     TimedQueue _timingFunctions;
     DelayedFunctionQueue _delayedFunctions;
     InterruptiveConcurrentQueue<Function> _interruptiveFunctions;
-    static const size_t _isInterruptiveFd = 0;
-    static const size_t _isTimingFd       = 1;
-    static const size_t _reventSize       = 1024;
+    constexpr static const size_t _isInterruptiveFd = 0;
+    constexpr static const size_t _isTimingFd       = 1;
+    constexpr static const size_t _reventSize = 1024;
```
gxuu, line 75: This needs to be correctly aligned. Install clangd, and it will automatically format it for you upon save.

scaler/io/ymq/error.h
gxuu, line 86: Remember to fix in the next PR.

PR #166, opened 2025-07-23
Resolve comment from last PR.

gxuu:
Some comments need further discussion before they can be resolved. Those will not be resolved in this PR. @magniloquency and I should reach consensus before those comments can be resolved.

### DIFF
--- a/scaler/io/ymq/tcp_client.cpp
+++ b/scaler/io/ymq/tcp_client.cpp
@@ -52,87 +52,85 @@ void TcpClient::onCreated() {
                 });
                 break;
         }
-    }
-
-    this->_connFd = sockfd;
-    int ret       = connect(sockfd, (sockaddr*)&_remoteAddr, sizeof(_remoteAddr));
-
-    int passedBackValue = 0;
-    if (ret < 0) {
-        if (errno != EINPROGRESS) {
-            close(sockfd);
-            passedBackValue = errno;
-        } else {
-            _eventLoopThread->_eventLoop.addFdToLoop(sockfd, EPOLLOUT | EPOLLET, this->_eventManager.get());
-            passedBackValue = 0;
-        }
-    } else {
-        std::string id = this->_localIOSocketIdentity;
-        auto sock      = this->_eventLoopThread->_identityToIOSocket.at(id);
-        sock->onConnectionCreated(setNoDelay(sockfd), getLocalAddr(sockfd), getRemoteAddr(sockfd), true);
-
-        passedBackValue = 0;
-    }
-
-    if (_retryTimes == 0 && passedBackValue == 0) {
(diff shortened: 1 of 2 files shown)


scaler/io/ymq/tcp_client.cpp
```diff
@@ -31,30 +30,34 @@ void TcpClient::onCreated() {
...
+    } else {
+        perror("connect");
+        close(sockfd);
+        passedBackValue = errno;
     }
 
     if (_retryTimes == 0) {
         _onConnectReturn(passedBackValue);
         return;
     }
```
magniloquency, line 58: should this check happen after checking `passedBackValue`?

gxuu: No

scaler/io/ymq/tcp_client.cpp
```diff
@@ -31,30 +30,34 @@ void TcpClient::onCreated() {
...
 
-    if (passedBackValue < 0) {
+    if (passedBackValue != 0) {
         printf("SOMETHING REALLY BAD\n");
         exit(-1);
     }
```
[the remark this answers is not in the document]
gxuu: No we will use unrecoverableError to replace the exit

scaler/io/ymq/tcp_client.cpp
```diff
@@ -31,30 +30,34 @@ void TcpClient::onCreated() {
...
-        if (errno != EINPROGRESS) {
-            perror("connect");
-            close(sockfd);
-            passedBackValue = errno;
-        } else {
-            _eventLoopThread->_eventLoop.addFdToLoop(sockfd, EPOLLOUT | EPOLLET, this->_eventManager.get());
-            passedBackValue = 0;
-        }
-    } else {
+    // Very unlikely
```
rafa-be, line 33: You can possibly add these attributes instead:
https://en.cppreference.com/w/cpp/language/attributes/likely

scaler/io/ymq/tcp_client.cpp
```diff
@@ -131,6 +134,7 @@ TcpClient::~TcpClient() noexcept {
     if (_connFd) {
         _eventLoopThread->_eventLoop.removeFdFromLoop(_connFd);
         close(_connFd);
+        _connFd = 0;
```
magniloquency, line 137: Would it be better to use `-1`? `0` is stdin, or even better can we use an `optional<>`?

magniloquency: ok, we can change it later, i'm just worried that `fd = 0` could cause some difficult bugs

rafa-be (dismissed):

sharpener6 (dismissed):

PR #179, opened 2025-08-05
Build cmake presets

gxuu:
Rationale:
- Visual Studio likes `CMakePresets.json` or `CMakeSettings.json`. If none of the file presents, then it stop to work. The former is cross platform and backed by `CMake`, while the other is really just a Microsoft invention that's not cross platform at all.
- The C++ parts of the project has grown to a point that it make sense to introduce presets. This unifies how we build the C++ components.
- Let `CMake` do the dirty cross platform work. The build process used to be in `build.sh`, which unfortunately will not work on Windows. This part should be minimal. The next step should be keeping `build.sh` and adding a new `build.bat`, which is typical for most C/C++ project; Or we rewrite `build.sh` using `Python`.

How to use `CMakePresets.json`: Mostly just run `./scripts/build.sh`. If you are hacky, you can also run
```
cmake --preset unix-build -GNinja [...]
cmake --build --preset unix-build[-linux]
```

Or, you can have your own `CMakeUserPresets.json`. But that seems a bit overkill.

Changes that you might care:
- The build directory is not lower-cased anymore.
- Change in `gitignore` to remove redundant exclusion and add the `CMakeUserPresets.json` that we would like to ignore. Be noted that this file is explicitly for your own purpose in CMake docs, so they should not be checked in.


CMakePresets.json
```diff
@@ -0,0 +1,60 @@
...
+    {
+      "name": "darwin-arm64",
+      "displayName": "Apple ARM 64 bit build",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}"
+      },
+      "binaryDir": "${sourceDir}/build_darwin_arm64"
+    }
+  ],
```
rafa-be, line 22: It seems like you could make a single generic preset for all UNIX builds:

```json
    {
      "name": "unix",
      "displayName": "UNIX build",
      "hidden": false,
      "condition": {
        "type": "equals",
        "lhs": "${hostSystemName}",
        "rhs": "Linux"
      },
      "cacheVariables": {
        "CMAKE_BUILD_TYPE": "Debug",
        "CMAKE_INSTALL_PREFIX": "${sourceDir}"
      },
      "binaryDir": "${sourceDir}/build_${hostSystemName}_${hostArchitecture}"
    },
```

gxuu: Unfortunately hostArchitecure does not exist. 

I tried with a few different approaches, the main difficulties were the Preset is a json file meaning we can't execute any command beside CMake ones; and workflow seems to hate environment variables. 

The last commit with message that makes sense was a generic build. After that I tried with things above but none of them would work.

rafa-be: > Unfortunately hostArchitecure does not exist.

Haha, I shouldn't trust all ChatGPT says!

PR #182, opened 2025-08-06
Resolve build issues on Windows

gxuu:
- Update CMakePresets.json to support Windows build.
- Update CMakeLists so OSS doesn't build on Windows.

Few notes for folks who will work on Windows dev:
- Visual Studio likes `CMakeUserPreseets.json`. That's the best way I know to work with `CMakePresets.json` on Windows. Building on Visual Studio requires the developer to feed in the locally exist include directory if the package is not configured with `find_package`. (In this particular case, `boost`)
- In `CMakeUserPresets.json` it is also a good idea to set up `vcpkg`.

### DIFF
--- a/scaler/CMakeLists.txt
+++ b/scaler/CMakeLists.txt
@@ -1,4 +1,8 @@
-add_subdirectory(object_storage)
+if(LINUX OR APPLE)
+  add_subdirectory(object_storage)
+else()
+  message(WARNING "Not building OSS, as it's not supported on this system currently!")
+endif()
 
 if(LINUX)
   add_subdirectory(io/ymq)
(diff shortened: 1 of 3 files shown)


rafa-be (approved):

scaler/CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
-add_subdirectory(object_storage)
+if(LINUX OR APPLE)
+  add_subdirectory(object_storage)
+else()
+  message(WARNING "Not building OSS, as it's not supported on this system currently!")
```
sharpener6, line 4: message(FATAL_ERROR "...") here

sharpener6: ok

PR #195, opened 2025-08-20
Add a tagged task allocation policy

rafa-be:
This is the first part of adding tagging support for Scaler.

Currently, all tags arguments passed to the worker or the client will be ignored.
That's because the **tagged allocation policy is disabled**, as it requires the additional state machine fixes from @sharpener6.

Once these state machine changes get merged, I'll submit a second PR enabling tagging.

### DIFF
--- a/scaler/protocol/python/message.py
+++ b/scaler/protocol/python/message.py
@@ -127,6 +127,31 @@ class TaskCancel(Message):
         )
 
 
+class TaskLog(Message):
+    class LogType(enum.Enum):
+        Stdout = _message.TaskLog.LogType.stdout
+        Stderr = _message.TaskLog.LogType.stderr
+
+    def __init__(self, msg):
+        super().__init__(msg)
+
+    @property
+    def task_id(self) -> TaskID:
+        return TaskID(self._msg.taskId)
+
+    @property
+    def log_type(self) -> "TaskLog.LogType":
+        return TaskLog.LogType(self._msg.logType.raw)
+
+    @property
+    def content(self) -> str:
+        return self._msg.content
+
+    @staticmethod
+    def new_msg(task_id: TaskID, log_type: "TaskLog.LogType", content: str) -> "TaskLog":
+        return TaskLog(_message.TaskLog(taskId=bytes(task_id), logType=log_type.value, content=content))
+
+
 class TaskResult(Message):
     def __init__(self, msg):
         super().__init__(msg)
@@ -612,6 +637,7 @@ PROTOCOL: bidict.bidict[str, Type[Message]] = bidict.bidict(
         "task": Task,
         "taskCancel": TaskCancel,
         "taskResult": TaskResult,
+        "taskLog": TaskLog,
         "graphTask": GraphTask,
         "graphTaskCancel": GraphTaskCancel,
         "objectInstruction": ObjectInstruction,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,14 @@ dependencies = [
     "tblib",
 ]
 
+[dependency-groups]
+dev = [
+    "black>=25.1.0",
+    "flake8>=7.3.0",
+    "flake8-pyproject>=1.2.3",
+    "mypy>=1.17.1",
+]
+
 [tool.scikit-build.metadata.version]
 provider = "scikit_build_core.metadata.regex"
 regex = '(?P<value>\d+\.\d+\.\d+(\.[1-9]\d*)?)'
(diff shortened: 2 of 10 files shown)


scaler/scheduler/allocate_policy/even_load_allocate_policy.py
```diff
@@ -129,13 +132,7 @@ def remove_task(self, task_id: TaskID) -> Optional[WorkerID]:
         self._worker_queue.decrease_priority(worker)
         return worker
 
-    def get_assigned_worker(self, task_id: TaskID) -> Optional[WorkerID]:
-        if task_id not in self._task_id_to_worker:
-            return None
-
-        return self._task_id_to_worker[task_id]
```
rafa-be, line 136: I removed this method as it basically does the same thing as `get_worker_by_task_id()`.

scaler/protocol/python/message.py
```diff
@@ -526,23 +538,26 @@ def status(self) -> TaskStatus:
         return TaskStatus(self._msg.status.raw)
 
     @property
-    def worker(self) -> WorkerID:
-        return WorkerID(self._msg.worker)
+    def worker(self) -> Optional[WorkerID]:
```
sharpener6, line 541: let's make `WorkerID(b"")` as invalid `InvalidWorkerID`

scaler/version.txt
```diff
@@ -1 +1 @@
-1.11.2
+1.11.3
```
gxuu, line 1: Given that tag is a new feature, should be `1.12.2`?

tests/test_tagged_allocate_policy.py
```diff
@@ -0,0 +1,172 @@
...
+    def test_balancing(self):
+        N_TASKS = MAX_TASKS_PER_WORKER // 2 * 2  # must be even
+
+        allocator = TaggedAllocatePolicy()
+
+        allocator.add_worker(WorkerID(b"worker_1"), {"linux", "gpu"}, MAX_TASKS_PER_WORKER)
+
+        # Assign a few tasks
+
+        for i in range(0, N_TASKS // 2):
```
gxuu, line 133: Why are you dividing it again by 2? Also if we set `MAX_TASKS_PER_WORKER==4`, the loop will be a nop.

tests/test_tagged_allocate_policy.py
```diff
@@ -0,0 +1,172 @@
...
+
+        allocator.add_worker(WorkerID(b"worker_2"), {"windows"}, MAX_TASKS_PER_WORKER)
+        self.assertDictEqual(allocator.balance(), {})
+
+        # Adds a worker that can accept some of the tasks
+
+        allocator.add_worker(WorkerID(b"worker_3"), {"gpu"}, MAX_TASKS_PER_WORKER)
+        balancing_advice = allocator.balance()
+
+        avg_tasks_per_worker = N_TASKS / 3
```
gxuu, line 149: I think `N_TASKS / NUMOFWORKERS` or something like that would be better.

scaler/scheduler/allocate_policy/tagged_allocate_policy.py
```diff
@@ -0,0 +1,279 @@
...
+    def statistics(self) -> Dict:
+        return {
+            worker.worker_id: {"free": worker.n_free(), "sent": worker.n_tasks(), "tags": worker.tags}
+            for worker in self._worker_id_to_worker.values()
+        }
+
+    def __get_available_workers_for_tags(self, tags: Set[str]) -> List[_WorkerHolder]:
+        # Worst-case time complexity is O(n_workers • len(tags))
+
+        if any(tag not in self._tag_to_worker_ids for tag in tags):
```
gxuu, line 269: Would it be better if we make this algorithm do a single pass on `_tag_to_worker_ids`? Seems like there are at least two passes. If the map is small then that's fine.

scaler/scheduler/allocate_policy/tagged_allocate_policy.py
```diff
@@ -0,0 +1,279 @@
...
+        return None
+
+    def assign_task(self, task: Task) -> Optional[WorkerID]:  # type: ignore[override]
+        # FIXME: remove async in TaskAllocatePolicy interface
+
+        # Worst-case time complexity is O(n_workers • len(task.tags))
+
+        available_workers = self.__get_available_workers_for_tags(task.tags)
+
+        if len(available_workers) <= 0:
```
gxuu, line 233: Just equality would be good enough?

gxuu:
Can I get a possible input of `tags`? Say we support more complex rules for tagging, would it be hard to parse? What are the considerations with respect to backward compatibility?

And I agree with @sharpener6 's comment wrt the default value of `ID`s. Is it sane to make special value the default for `ID`s? `0x0` or `0xDEADBEEF` would be some sane default. (I am in favor of the `deadbeef` one as it is more noticeable when you inspect the memory, but that's also more work).

scaler/version.txt
rafa-be: I'll update it to `1.12` once it's enabled. Currently it's not enabled as it requires the state machine changes from sharpener6.

sharpener6 (approved):

PR #196, opened 2025-08-20
Introducing c++ api for using python's logging module

I-Need-Sleep-Asap:


I-Need-Sleep-Asap:
- As of now we have a crash when executing test_object_storage_server case.

Reason:
Here we implement the py <-> c++ wrapper as singleton class. This is because Py_Initialize() & other functions can have 1 instance (due to python's GIL)

The root problem is likely an uncontrolled Py_FinalizeEx() call or a concurrency issue with Python's GIL (Global Interpreter Lock). The Python C API is not thread-safe without explicit GIL management. When a test completes, the server thread exits, but other parts of the application or the testing framework might be interacting with the Python state, or the interpreter's state is being destroyed implicitly.

rafa-be:
@sharpener6 

@gxuu and I called @I-Need-Sleep-Asap. The problem with calling Python's `logging` module from C++ is that it has to happen from the main Python thread.

However, that's not easy as C++/Boost's Asio runs in a separate thread. That messes up with Python's internal state.

YMQ will not solve the issue as its core runs in a separate C++ thread.

Considering this, we think the easiest way to solve it is to have a C++ logging class that mimics Python's (e.g. same formatting), and to somehow forward the `setup_logger()` parameters to C++.

sharpener6:
can you we explict passing formatter and level when constructing

I-Need-Sleep-Asap:
Previously:
image


Currently:
image

scaler/io/ymq/logging.h
```diff
@@ -24,23 +27,65 @@ inline LoggingLevel LOGGING_LEVEL = LoggingLevel::info;
...
+        (message_stream << ... << args);
+    }
+    std::string message = message_stream.str();
+
+    std::ostringstream ts_stream;
+    ts_stream << Timestamp{};
+    std::string timestamp_str = ts_stream.str();
+
+    std::string output = format;
+    replaceAll(output, "%(levelname)s", std::string(convertLevelToString(level)));
```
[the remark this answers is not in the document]
I-Need-Sleep-Asap: @sharpener6 Generally speaking, level name, time, date & other print messages are not very long. Hence, the performance impact would be negligible.

But i have pushed new patch which will parse the characters & replace them on-the-fly in a single pass

scaler/io/ymq/logging.h
```diff
@@ -24,23 +27,83 @@ inline LoggingLevel LOGGING_LEVEL = LoggingLevel::info;
...
+    }
+
+    // Write any remaining literal text after the last placeholder
+    if (last_pos < format.length()) {
+        output_stream.write(format.c_str() + last_pos, format.length() - last_pos);
+    }
+
+    // Use std::print and flush for immediate output
+    std::print("{}\n", output_stream.str());
+    std::fflush(stdout);
```
rafa-be, line 106: I think you'll have to also pass the `log_paths` parameter from Python, as logs might not always be forwarded to stdout.

scaler/io/ymq/logging.h
rafa-be, line 1: I think it would be nice to have a few unit tests for `log()`, especially with special cases (empty string, `%%`, possible overflows ...).

It should be pretty easy if you make the output file configurable, and then capture it in the tests.

rafa-be (changes requested):

scaler/io/ymq/logging.h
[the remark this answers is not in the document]
sharpener6: @I-Need-Sleep-Asap, theoretically all those keywords can happened in the same logging message for many times like (even we should avoid that):

`[%(levelname)s] %(asctime)s: %(message)s %(asctime)s`

scaler/io/ymq/logging.h
I-Need-Sleep-Asap: done! pushed new patch

scaler/io/ymq/logging.h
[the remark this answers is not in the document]
I-Need-Sleep-Asap: @sharpener6  yes my code will deal with the scenario you mentioned.

scaler/io/ymq/logging.h
rafa-be, line 1: Could we somehow avoid to have to pass the `log_format` and `log_path` arguments every time `log()` is called?

Maybe have a `Logger` class that we can use as a field for `ObjectStorageServer`?

```c++
class Logger {
public:
    enum LogLevel { ... };

    constexpr Logger(std::string log_level, std::string log_format, std::string log_path);
    
    template <typename T, typename... Args>
    log(LoggingLevel level, T&& generic_format, Args&&... args);
};

// Usage:

int main()
{
    Logger logger(Logger::LogLevel::INFO, "%(levelname)s: %(message)s", "/dev/stderr");
    
    // ...
    
    logger.log(Logger::LogLevel::ERROR, "Some error occured");
}
```

In addition, this would avoid to re-open the log file every time `log()` is called.

tests/object_storage/test_logging.cpp
rafa-be, line 1: Could you move this file to a new `tests/io/ymq/` directory?

You'll have to create the associated `CMakeLists.txt`. Also, the `add_test_executable()` function from `tests/object_storage/CMakeLists.txt` will have to be moved to `tests/CMakeLists.txt` so that it can be share with all files.

tests/object_storage/test_logging.cpp
```diff
@@ -0,0 +1,181 @@
...
+};
+
+TEST_F(ObjectStorageLoggingTest, TestServerLogsDisconnectionToFile) {
+    {
+        SimpleClient client(ioContext, SERVER_HOST, serverPort);
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    std::string log_content = readLogFile();
+    EXPECT_NE(log_content.find("INFO: ObjectStorageServer: client disconnected"), std::string::npos);
+}
```
rafa-be, line 181: I think this test should be in `test_object_storage_server.cpp`.

rafa-be (approved):

rafa-be (changes requested):
Please resolve the conversations.

scaler/io/ymq/logging.h
sharpener6: I agree

tests/object_storage/test_logging.cpp
I-Need-Sleep-Asap: done. Will push in next patch

scaler/cluster/object_storage_server.py
```diff
@@ -33,4 +33,7 @@ def run(self) -> None:
         setup_logger(self._logging_paths, self._logging_config_file, self._logging_level)
         logging.info(f"ObjectStorageServer: start and listen to {self._storage_address.to_string()}")
 
-        self._server.run(self._storage_address.host, self._storage_address.port)
+        log_format_str, log_level_str, logging_path = get_logger_info(logging.getLogger())
+
+        self._server.run(self._storage_address.host, self._storage_address.port, log_level_str,
+                         log_format_str, logging_path)
```
sharpener6, line 39: please use black to formatting your python code

scaler/io/ymq/logging.h
```diff
@@ -1,47 +1,159 @@
...
+  #ifdef NDEBUG  // Release build
+  static constexpr LoggingLevel LOGGING_LEVEL = LoggingLevel::info;
+  #else  // Debug build
+  static constexpr LoggingLevel LOGGING_LEVEL = LoggingLevel::debug;
+  #endif
+
+  Logger(std::string log_format = "%(levelname)s: %(message)s",
+         std::string log_path = "/dev/stdout",
+         LoggingLevel level = LOGGING_LEVEL)
+      : log_path_(std::move(log_path)), level_(level) {
```
sharpener6, line 33: you need use clang-format to formatting the code

I-Need-Sleep-Asap: FYI, i had to edit .clang-format to make it work

scaler/io/ymq/tcp_client.h
```diff
@@ -58,6 +59,8 @@ class TcpClient {
     sockaddr _remoteAddr;
     int _retryIdentifier;
 
+    Logger logger_;
```
rafa-be, line 62: Typo, should be `_logger`

scaler/io/ymq/tcp_server.h
```diff
@@ -63,6 +64,8 @@ class TcpServer {
 
     std::unique_ptr<EventManager> _eventManager;  // will copy the `onRead()` to itself
 
+    Logger logger_;
```
rafa-be, line 67:
```suggestion
    Logger _logger;
```

scaler/object_storage/object_storage_server.cpp
```diff
@@ -16,8 +16,12 @@ ObjectStorageServer::~ObjectStorageServer()
     closeServerReadyFds();
 }
 
-void ObjectStorageServer::run(std::string name, std::string port)
+void ObjectStorageServer::run(
+    std::string name, std::string port, std::string log_level, std::string log_format, std::string log_path)
 {
+    assert(!log_level.empty() && !log_format.empty());
```
rafa-be, line 22: This should be in the `Logger` class's constructor I think. And maybe raise a `std::invalid_argument` instead of an using assets.

scaler/object_storage/object_storage_server.h
```diff
@@ -65,6 +70,8 @@ class ObjectStorageServer {
     // Some GET and DUPLICATE requests might be delayed if the referenced object isn't available yet.
     std::map<ObjectID, std::vector<PendingRequest>> pendingRequests;
 
+    scaler::ymq::Logger logger_;
```
rafa-be, line 73:
```suggestion
    scaler::ymq::Logger _logger;
```

tests/CMakeLists.txt
```diff
@@ -1,3 +1,24 @@
...
+        PRIVATE GTest::gtest_main
+        PRIVATE cc_object_storage_server
+    )
+
+    add_test(NAME ${test_name} COMMAND ${test_name})
+endfunction()
+
+
 if(LINUX OR APPLE)
+  # This directory fetches Google Test, so it must be included first.
```
rafa-be, line 19: Shouldn't it be better to move the GTest fetching instructions to this file then?

tests/io/ymq/test_logging.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <chrono>
+#include <cstdio>
+#include <fstream>
+#include <limits>
+#include <string>
+#include <thread>
+#include <vector>
+
+#include "scaler/io/ymq/logging.h"
+#include "scaler/object_storage/object_storage_server.h"
```
rafa-be, line 12: This include is not needed anymore.

tests/object_storage/test_object_storage_server.cpp
```diff
@@ -160,7 +164,7 @@ TEST_F(ObjectStorageServerTest, TestGetObject)
         };
 
         requestHeader.payloadLength = UINT64_MAX;
-        requestHeader.requestType   = ObjectRequestType::GET_OBJECT,
+        requestHeader.requestType   = ObjectRequestType::GET_OBJECT;
```
rafa-be, line 167: There lines are useless I think. I'm not sure why it was compiling before.

tests/object_storage/test_object_storage_server.cpp
```diff
@@ -548,3 +552,63 @@ TEST_F(ObjectStorageServerTest, TestMalformedHeader)
         EXPECT_EQ(responseHeader.responseType, ObjectResponseType::SET_O_K);
     }
 }
+
+// This test fixture is specifically for verifying server logging behavior.
+class ObjectStorageLoggingTest: public ::testing::Test {
+protected:
+    static constexpr std::string SERVER_HOST = "127.0.0.1";
+    const std::string log_filename           = "server_integration_log.txt";
```
rafa-be, line 560: You should use [`std::tmpfile`](https://en.cppreference.com/w/cpp/io/c/tmpfile) instead.

rafa-be (changes requested):

scaler/object_storage/object_storage_server.cpp
I-Need-Sleep-Asap: i dont think we need this because we have default values for these params. Ill delete this entirely

tests/object_storage/test_object_storage_server.cpp
I-Need-Sleep-Asap: Im not sure how it i was compiling before but decided to fix it anyways.

tests/object_storage/test_object_storage_server.cpp
I-Need-Sleep-Asap: I thought of using this initially but decided to pass due to 

1. File names are not user friendly & descriptive
2. It seems to be having some permission/security issues with Windows (https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/tmpfile?view=msvc-170)

Although i admit i didn't do much testing on windows yet. So decided to pass on this

i can make the changes to use std::tmpfile & deal with windows issues if we get any in future. What do you say @rafa-be  ?

.clang-format
```diff
@@ -58,7 +58,7 @@ UseTab: Never
 # control function declaration/definition
 AllowShortFunctionsOnASingleLine: InlineOnly
 AllowAllParametersOfDeclarationOnNextLine: true
-BinPackParameters: OnePerLine
+BinPackParameters: false
```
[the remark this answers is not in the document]
rafa-be: Are you sure you're using Clang-format 21, and not another executable that is still in your `$PATH`?

I don't have any issue with 21.1:

```shell
$ clang-format --version
Homebrew clang-format version 21.1.0
$ clang-format scaler/io/ymq/logging.h > /dev/null
$ echo $?
0
```

I had the same error with Clang-format 19 though?

PR #199, opened 2025-08-22
Update linter to test on Windows

gxuu:
# FOR INITIAL REVIEW

ALL functionalities issues have been resolved, and I have been trying to keep things a little bit DRYer in the last few hours. But I don't see significant improvement no matter what I do. As of now, there are basically two sets of everything (build script, install dependencies script, etc.). I would be grateful if anyone offer some insights on how to make this simpler without losing functionalities listed below:
- we need to pin the version of installed dependencies
- If we don't build `capnp` from source, then the package must contain `FindCapnProto.cmake` so we can use `find_package` to find `capnp`.
- be sure to think about release pipeline as well when you make changes. This is mostly related to `download_install_dependencies`.

# ON WHY NOT USE A PACKAGE MANAGER
Explored a few different approaches. All of them were bad and eventually decided to go with the raw approach for dependencies - compile from source.

Here are the approaches I tried:
- chocolately. Not cross platformed, and problematic when you try to specify a version
- vcpkg. Problematic when you try to specify a version. Also they tend to pull in debug build when pinning the version.
- python script instead of a ps1 + a bash. Ended up with two segments of #ifdefs, not worth it.
- conan. Beside the fact that I and rafa share the same nightmare, I gave it a try. No, because not very good with non C++ project as a whole.

# Currently,
- Two scripts, one for Windows, and one for other platforms. Scripts are in bash so it should run on apple as well.
- capnp build is included, but it is not used because OSS is still not turned on. Making OSS run on Windows is a non trivial work, due to the usage of `pipe` and `read/write` etc. It is still included because someday the `client` code will need it anyway.


.github/workflows/linter.yml
rafa-be, line 1: I think you could avoid naming the `(GNU/Linux)` and `(Windows)` postfixes if you name the two build differently, e.g. `build-windows` and `build-linux`.

rafa-be (approved):
LGTM 👍, but do not forget to update the _Build from source_ section in the README with the instruction about how to build the package on Windows.

rafa-be (approved):

PR #224, opened 2025-09-12
Swap asio to ymq for object storage server

gxuu:
# READY FOR REVIEW.

## STATUS QUO
Mimicking the YMQ protocol in the client side works. Tests are passing, see Actions for detail. Simple swap in `SyncObjectStorageConnector` from plain socket to `YMQ` would not work, and it is a _very very_ hard bug to find.

## REGRESSIONS
- The program could resolve ns record but now it can't
- The program could close malicious connection but now it can't
- The program could send objects without copying but now it can't
- The program could actively be shutdown but now it can't (inefficient shutdown only)

### DIFF
--- a/scaler/object_storage/object_storage_server.cpp
+++ b/scaler/object_storage/object_storage_server.cpp
@@ -118,39 +118,32 @@ void ObjectStorageServer::closeServerReadyFds()
     }
 }
 
-ObjectRequestHeader ObjectStorageServer::parseRequestHeader(Bytes message)
-{
-    assert(message.len() == CAPNP_HEADER_SIZE);
-    std::array<uint64_t, ObjectRequestHeader::bufferSize() / CAPNP_WORD_SIZE> buffer {};
-    memcpy(buffer.data(), message.data(), message.size());
-    return ObjectRequestHeader::fromBuffer(buffer);
-}
-
 void ObjectStorageServer::processRequests()
 {
     using namespace std::chrono_literals;
     try {
         std::map<ymq::Configuration::IOSocketIdentity, std::pair<ObjectRequestHeader, Bytes>> identityToFullRequest;
-        std::vector<std::future<std::expected<void, ymq::Error>>> pendingSendMessageFuts;
 
         while (true) {
-            auto invalids = std::ranges::partition(pendingSendMessageFuts, [](const auto& x) { return x.valid(); });
-            pendingSendMessageFuts.resize(std::distance(pendingSendMessageFuts.begin(), invalids.begin()));
+            auto invalids = std::ranges::remove_if(_pendingSendMessageFuts, [](const auto& x) { return !x.valid(); });
+            _pendingSendMessageFuts.erase(invalids.begin(), invalids.end());
 
-            std::ranges::for_each(pendingSendMessageFuts, [](auto& fut) {
+            std::ranges::for_each(_pendingSendMessageFuts, [](auto& fut) {
                 if (fut.wait_for(0s) == std::future_status::ready) {
                     auto res = fut.get();
                     assert(res);
                 }
             });
 
+            // TODO: This adds up to 100ms delay when trying to shutdown the server. This is due to ymq doesn't let you
+            // call removeIOSocket in another thread, which shouldn't be the case.
             auto maybeMessageFut = ymq::futureRecvMessage(_ioSocket);
             while (maybeMessageFut.wait_for(100ms) == std::future_status::timeout) {
                 if (_stopped) {
                     _logger.log(
                         scaler::ymq::Logger::LoggingLevel::info,
                         "ObjectStorageServer: stopped, number of messages leftover in the system = ",
-                        std::ranges::count_if(pendingSendMessageFuts, [](auto& x) {
+                        std::ranges::count_if(_pendingSendMessageFuts, [](auto& x) {
                             return x.valid() && x.wait_for(0s) == std::future_status::timeout;
                         }));
                     return;
@@ -168,7 +161,7 @@ void ObjectStorageServer::processRequests()
 
             auto it = identityToFullRequest.find(identity);
             if (it == identityToFullRequest.end()) {
-                identityToFullRequest[identity].first = parseRequestHeader(std::move(headerOrPayload));
+                identityToFullRequest[identity].first = ObjectRequestHeader::fromBuffer(headerOrPayload);
                 const auto& requestType               = identityToFullRequest[identity].first.requestType;
                 if (requestType == ObjectRequestType::DUPLICATE_OBJECT_I_D ||
                     requestType == ObjectRequestType::SET_OBJECT) {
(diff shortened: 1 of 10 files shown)


scaler/object_storage/object_storage_server.cpp
```diff
@@ -104,72 +118,103 @@ void ObjectStorageServer::closeServerReadyFds()
...
+{
+    using namespace std::chrono_literals;
+    try {
+        std::map<ymq::Configuration::IOSocketIdentity, std::pair<ObjectRequestHeader, Bytes>> identityToFullRequest;
+        std::vector<std::future<std::expected<void, ymq::Error>>> pendingSendMessageFuts;
 
-    setServerReadyFd();
+        while (true) {
+            auto invalids = std::ranges::partition(pendingSendMessageFuts, [](const auto& x) { return x.valid(); });
+            pendingSendMessageFuts.resize(std::distance(pendingSendMessageFuts.begin(), invalids.begin()));
```
rafa-be, line 138: Isn't this `std::ranges::remove_if`?

```suggestion
            std::ranges::remove_if(pendingSendMessageFuts, [](const auto& x) { return !x.valid(); });
```

rafa-be, line 138: What are these invalid futures exactly? Are they safe to ignore?

scaler/object_storage/object_storage_server.cpp
```diff
@@ -104,72 +118,103 @@ void ObjectStorageServer::closeServerReadyFds()
...
+    return ObjectRequestHeader::fromBuffer(buffer);
+}
 
-    _logger.log(scaler::ymq::Logger::LoggingLevel::info, "ObjectStorageServer: started");
+void ObjectStorageServer::processRequests()
+{
+    using namespace std::chrono_literals;
+    try {
+        std::map<ymq::Configuration::IOSocketIdentity, std::pair<ObjectRequestHeader, Bytes>> identityToFullRequest;
+        std::vector<std::future<std::expected<void, ymq::Error>>> pendingSendMessageFuts;
```
rafa-be, line 134: I think that if you make these `private` member fields, you can avoid the backtracking logic of all these futures, by directly calling `this->pendingSendMessageFuts.emplace()` from `writeMessage()`.

That would simplify the code a lot, removing the need of `catFuts()` and `PairOrVectorOf`.

You might argue that then you will not be able to call `processRequests()` from multiple threads. But that's already the case, as there will be concurrency issues with `_ioSocket`.

scaler/object_storage/object_storage_server.h
```diff
@@ -72,57 +69,77 @@ class ObjectStorageServer {
...
+
+        if constexpr (requires { src.begin(); }) {
+            // NOTE: Not the std::move you might expect!
+            dest.resize(dest.size() + src.size());
+            auto first = dest.end() - src.size();
+            std::move(src.begin(), src.end(), first);
+            return;
+        }
+        assert(false);
     }
```
rafa-be, line 132: Isn't this basically re-implement function overloading?

```c++
void catFuts(std::vector<ObjectStorageServer::SendMessageFuture>& dest, std::pair<SendMessageFuture>&& src)
{
    dest.emplace_back(std::move(src.first));
    dest.emplace_back(std::move(src.second));
}
    
void catFuts(std::vector<ObjectStorageServer::SendMessageFuture>& dest, std::vector<SendMessageFuture>&& src) 
{
    // NOTE: Not the std::move you might expect!
    dest.resize(dest.size() + src.size());
    auto first = dest.end() - src.size();
    std::move(src.begin(), src.end(), first);
}
```

tests/object_storage/test_object_manager.cpp
```diff
@@ -13,22 +15,29 @@ TEST(ObjectManagerTestSuite, TestSetObject) {
...
     EXPECT_TRUE(objectManager.hasObject(objectID1));
     EXPECT_EQ(objectManager.size(), 1);
     EXPECT_EQ(objectManager.sizeUnique(), 1);
 
     scaler::object_storage::ObjectID objectID2 {3, 2, 1, 0};
 
-    objectManager.setObject(objectID2, std::move(std::vector(payload)));
+    {
+        scaler::object_storage::ObjectPayload tmp = payload;
+        objectManager.setObject(objectID2, std::move(tmp));
```
rafa-be, line 31: Might be better to call the copy constructor directly

```suggestion
        objectManager.setObject(objectID2, ObjectPayload(payload));
```

tests/object_storage/test_object_storage_server.cpp
```diff
@@ -1,71 +1,85 @@
...
+    }
 
     void writeRequest(const ObjectRequestHeader& header, const std::optional<ObjectPayload>& payload)
     {
         auto buf = header.toBuffer();
 
-        write(boost::asio::buffer(buf.asBytes().begin(), buf.asBytes().size()));
+        Message ymqHeader {};
+        ymqHeader.payload = Bytes((char*)buf.asBytes().begin(), buf.asBytes().size());
+        writeYMQMessage(std::move(ymqHeader));
```
rafa-be, line 52: Making `ObjectRequestHeader::toBuffer()` return `Bytes` could simplify this (maybe rename it `toBytes()` then)

tests/object_storage/test_object_storage_server.cpp
```diff
@@ -511,48 +527,49 @@ TEST_F(ObjectStorageServerTest, TestClientDisconnect)
...
-        };
-
-        client->writeRequest(requestHeader, {payload});
-        client->readResponse(responseHeader, responsePayload);
-
-        EXPECT_EQ(responseHeader.responseType, ObjectResponseType::SET_O_K);
-    }
-}
+// TODO: This is a regression as ymq doesn't support actively closing an underlying connection
+// TEST_F(ObjectStorageServerTest, TestMalformedHeader)
```
rafa-be, line 531: Gtest allows you to disable tests with `DISABLED_`. That way these will not run, but still be compiled:

```suggestion
// TODO: This is a regression as ymq doesn't support actively closing an underlying connection
// TEST_F(ObjectStorageServerTest, DISABLED_TestMalformedHeader)
```

rafa-be (dismissed):

scaler/object_storage/object_storage_server.cpp
```diff
@@ -104,72 +118,103 @@ void ObjectStorageServer::closeServerReadyFds()
...
 
-    setServerReadyFd();
+        while (true) {
+            auto invalids = std::ranges::partition(pendingSendMessageFuts, [](const auto& x) { return x.valid(); });
+            pendingSendMessageFuts.resize(std::distance(pendingSendMessageFuts.begin(), invalids.begin()));
 
-    for (;;) {
-        auto client = std::make_shared<Client>(executor);
+            std::ranges::for_each(pendingSendMessageFuts, [](auto& fut) {
```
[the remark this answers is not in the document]
gxuu: Mostly just a personal habit - I would like to use this when the loop body is trivial. And you don't have to write `x.begin(), x.end()` when using `std::ranges` algorithm, making the code shorter to type.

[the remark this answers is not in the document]
gxuu: Do you want me to change to `for(x: futs)`?

scaler/object_storage/object_storage_server.cpp
gxuu: Indeed. Though `_ioSocket` is designed to be thread safe it is not possible to share it between multiple threads with current set up (aka. The clients send header and payload separately). I will modify this.

scaler/object_storage/object_storage_server.h
gxuu: Indeed, I was reading through some fancy `concept` stuff, and this really should be just overloading functions. I will change this.

rafa-be (dismissed):

scaler/object_storage/object_storage_server.cpp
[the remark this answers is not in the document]
rafa-be: IDK, I always thought `for (:)` is more readable, especially with C++'s ugly lambda syntax. But that's not a big deal.

We should try to standardize to one if the two though.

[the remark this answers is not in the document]
gxuu: I'd say if it is only one line of the loop body, and the loop body doesn't interfere with environment outside the loop body (so `std::for_each` in this case would take a lambda that is one or two line and doesn't capture anything), then use `for_each`.

rafa-be (dismissed):

sharpener6 (approved):

PR #228, opened 2025-09-16
introduce toml files for configuring scalar component's flags

I-Need-Sleep-Asap:


scaler/config.py
```diff
@@ -0,0 +1,239 @@
...
+                    key = f"{parent_name}.{child_field.name}"
+                    value = getattr(nested_obj, child_field.name)
+                    lines.append(f'{key} = "{value}"')
+        return "\n".join(lines)
+
+    @classmethod
+    def get_config(cls: Type[Self], filename: Optional[str] = None, args: Optional[Namespace] = None) -> Self:
+        """
+        Loads configuration from a TOML file and overrides it with command-line arguments.
+        This method is "smart": it allows unprefixed arguments (e.g., --web_port) when they
```
I-Need-Sleep-Asap, line 49: @rafa-be @sharpener6 : shall i use this crude method of resolving "same flag names" across several Configs? Or can i use https://docs.pydantic.dev/latest/ ?

scaler/config.py
```diff
@@ -0,0 +1,239 @@
...
+    for key, value in source.items():
+        if isinstance(value, dict) and key in destination and isinstance(destination[key], dict):
+            destination[key] = _deep_merge(value, destination[key])
+        else:
+            destination[key] = value
+    return destination
+
+
+@dataclasses.dataclass
+class Config(ABC):
```
sharpener6, line 30: this is not abstract base class, it has implementation

scaler/config.py
```diff
@@ -0,0 +1,239 @@
...
+                    logging.warning(f"Command-line argument '{key}' did not match any config field and was ignored.")
+
+        # Merge TOML data with command-line arguments, prioritizing args
+        final_data = copy.deepcopy(toml_data)
+        _deep_merge(args_data, final_data)
+
+        return _from_dict(cls, final_data)  # type: ignore [type-var]
+
+
+class _DataclassProtocol(Protocol):
```
sharpener6, line 115: I didn't see the reason why we are using protocol

e117649:
There are some regressions when parsing the positional argument `address` with the cli which should either be inferred or  update help string here https://github.com/finos/opengris-scaler/blob/7374432298be7a290e8069366a7240003202b429/scaler/entry_points/scheduler.py#L129

```bash
> python run_scheduler.py "tcp://127.0.0.1:2345"
WARNING:root:Command-line argument 'address' is ambiguous (exists in ['scheduler', 'object_storage']). Please use a prefix, e.g., 'scheduler_address'. Argument ignored.
> /workspaces/scaler/scaler/config.py(255)_from_dict()
-> return cls(**kwargs)
(Pdb)
```

```bash
> python run_cluster.py "tcp://127.0.0.1:2345" --num-of-workers=4
WARNING:root:Command-line argument 'heartbeat_interval' did not match any config field and was ignored.
WARNING:root:Command-line argument 'address' is ambiguous (exists in ['scheduler', 'object_storage']). Please use a prefix, e.g., 'scheduler_address'. Argument ignored.
  File "/workspaces/scaler/run_cluster.py", line 4, in <module>
    main()
  File "/workspaces/scaler/scaler/entry_points/cluster.py", line 154, in main
    scaler_config = ScalerConfig.get_config(args.config, args)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/scaler/scaler/config.py", line 98, in get_config
    return _from_dict(cls, final_data)  # type: ignore [type-var]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/scaler/scaler/config.py", line 251, in _from_dict
    nested_objects[field.name] = _from_dict(cast(Type[DataclassType], field.type), data[field.name])
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/scaler/scaler/config.py", line 255, in _from_dict
    return cls(**kwargs)
           ^^^^^^^^^^^^^
TypeError: WorkerConfig.__init__() got an unexpected keyword argument 'task_queue_size'
```

I-Need-Sleep-Asap:
> `tomli` is missing from dependencies in pyproject.toml, also are there any reasons we need it over tomllib or can we just use tomllib?

@e117649  : tomlib is introduced in python 3.11 but our pipeline suses 3.10 python

I-Need-Sleep-Asap:
> There are some regressions when parsing the positional argument `address` with the cli which should either be inferred or update help string here
> 
> https://github.com/finos/opengris-scaler/blob/7374432298be7a290e8069366a7240003202b429/scaler/entry_points/scheduler.py#L129
> 
> ```shell
> > python run_scheduler.py "tcp://127.0.0.1:2345"
> WARNING:root:Command-line argument 'address' is ambiguous (exists in ['scheduler', 'object_storage']). Please use a prefix, e.g., 'scheduler_address'. Argument ignored.
> > /workspaces/scaler/scaler/config.py(255)_from_dict()
> -> return cls(**kwargs)
> (Pdb)
> ```
> 
> ```shell
> > python run_cluster.py "tcp://127.0.0.1:2345" --num-of-workers=4
> WARNING:root:Command-line argument 'heartbeat_interval' did not match any config field and was ignored.
> WARNING:root:Command-line argument 'address' is ambiguous (exists in ['scheduler', 'object_storage']). Please use a prefix, e.g., 'scheduler_address'. Argument ignored.
> Traceback (most recent call last):
>   File "/workspaces/scaler/run_cluster.py", line 4, in <module>
>     main()
>   File "/workspaces/scaler/scaler/entry_points/cluster.py", line 154, in main
>     scaler_config = ScalerConfig.get_config(args.config, args)
>                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/workspaces/scaler/scaler/config.py", line 98, in get_config
>     return _from_dict(cls, final_data)  # type: ignore [type-var]
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/workspaces/scaler/scaler/config.py", line 251, in _from_dict
>     nested_objects[field.name] = _from_dict(cast(Type[DataclassType], field.type), data[field.name])
>                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/workspaces/scaler/scaler/config.py", line 255, in _from_dict
>     return cls(**kwargs)
>            ^^^^^^^^^^^^^
> TypeError: WorkerConfig.__init__() got an unexpected keyword argument 'task_queue_size'
> ```

I thought i fixed it. Not sure how this was missed. must be while resolving merge conflicts. Thanks! ill look into it & add them in unit test

rafa-be (dismissed):
Good job, looks way better than what we had before 👍

As an additional improvement, as the `config.py` file becomes quite large, we could split it, e.g. like this:

```
scaler/config
├── __init__.py
├── mixins.py  # Config, ConfigValue
├── zmq.py  # ZMQType, ZMQConfig
├── scheduler.py  # SchedulerConfig
...
```

scaler/entry_points/scheduler.py
```diff
@@ -2,118 +2,75 @@
...
 
 def get_args():
     parser = argparse.ArgumentParser("scaler scheduler", formatter_class=argparse.ArgumentDefaultsHelpFormatter)
-    parser.add_argument("--io-threads", type=int, default=DEFAULT_IO_THREADS, help="number of io threads for zmq")
+
+    parser.add_argument("--config", type=str, default=None, help="Path to the TOML configuration file.")
+    parser.add_argument("--zmq-io-threads", type=int, help="number of io threads for zmq")
     parser.add_argument(
-        "--max-number-of-tasks-waiting",
+        "--max_number_of_tasks_waiting",
```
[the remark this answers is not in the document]
I-Need-Sleep-Asap: its a typo. Fixed it

rafa-be (approved):

README.md
```diff
@@ -108,7 +108,7 @@ The object storage server, scheduler and workers can also be started from the co
 First, start the object storage server:
 
 ```bash
-$ scaler_object_storage_server tcp://127.0.0.1:2346
+$ scaler_object_storage_server -n 127.0.0.1 -p 2346
```
rafa-be, line 111: Is this correct? Looks like the entry-point is still expecting a string:
https://github.com/I-Need-Sleep-Asap/opengris-scaler/blob/ef894342c0bed0bf9aed13fdd6269d0d4aba7ad4/scaler/entry_points/object_storage_server.py

I-Need-Sleep-Asap:
hello,
@rafa-be  approved my PR but has additional feedback. His concerns were: some of the fields such as scheduler_address is not marked as Optional in config.py

Most of the modules NEED scheduler_address except few (such as webui). This phenomenon is true for other few flags too.
Not all flags are mandatory for every module.
Our current approach of having single config & toml file will complicate & makes it tough to map these complex dependencies between flags & modules.
We might need to rework entire code & introduce several config classes & several toml files which adds additional complexity.

Current PR is **_functionally correct_**. Hence i was wondering if you guys needs to make it semantically correct too. This will add more complexity. Please lemme know your thoughts @sharpener6 @1597463007 @e117649

rafa-be:
> hello, @rafa-be approved my PR but has additional feedback. His concerns were: some of the fields such as scheduler_address is not marked as Optional in config.py
> 
> Most of the modules NEED scheduler_address except few (such as webui). This phenomenon is true for other few flags too. Not all flags are mandatory for every module. Our current approach of having single config & toml file will complicate & makes it tough to map these complex dependencies between flags & modules. We might need to rework entire code & introduce several config classes & several toml files which adds additional complexity.
> 
> Current PR is **_functionally correct_**. Hence i was wondering if you guys needs to make it semantically correct too. This will add more complexity. Please lemme know your thoughts @sharpener6 @1597463007 @e117649

The problem is that some config values might be mandatory for some executables, but not for others.

As an example, `--monitor-address` is a mandatory parameter for `scaler_ui` and `scaler_top`, but not for `scaler_scheduler` (as it will just open a random socket if not provided).

So providing a single config class for all these executables, like suggested in this PR, does not work.

From the example:

```python
@dataclasses.dataclass
class SchedulerConfig(Config):
    ...
    monitor_address: Optional[ZMQConfig] = None
    ...
```

Is valid for `scaler_scheduler`, but does should be this for `scaler_ui`:

```python
@dataclasses.dataclass
class SchedulerConfig(Config):
    ...
    monitor_address: Optional[ZMQConfig] = None
    ...
```

I-Need-Sleep-Asap:
> > > hello, @rafa-be approved my PR but has additional feedback. His concerns were: some of the fields such as scheduler_address is not marked as Optional in config.py
> > > Most of the modules NEED scheduler_address except few (such as webui). This phenomenon is true for other few flags too. Not all flags are mandatory for every module. Our current approach of having single config & toml file will complicate & makes it tough to map these complex dependencies between flags & modules. We might need to rework entire code & introduce several config classes & several toml files which adds additional complexity.
> > > Current PR is **_functionally correct_**. Hence i was wondering if you guys needs to make it semantically correct too. This will add more complexity. Please lemme know your thoughts @sharpener6 @1597463007 @e117649
> > 
> > 
> > The problem is that some config values might be mandatory for some executables, but not for others.
> > As an example, `--monitor-address` is a mandatory parameter for `scaler_ui` and `scaler_top`, but not for `scaler_scheduler` (as it will just open a random socket if not provided).
> > So providing a single config class for all these executables, like suggested in this PR, does not work.
> > From the example:
> > ```python
> > @dataclasses.dataclass
> > class SchedulerConfig(Config):
> >     ...
> >     monitor_address: Optional[ZMQConfig] = None
> >     ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is valid for `scaler_scheduler`, but should be this for `scaler_ui`:
> > ```python
> > @dataclasses.dataclass
> > class SchedulerConfig(Config):
> >     ...
> >     monitor_address: ZMQConfig
> >     ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After some though, I think the approach here is slightly broken. We cannot have a common config class hierarchy for all these executables. These should be different for all the executables (`scaler_scheduler`, `scaler_cluster`, `scaler_ui`, `scaler_top` ...).
> 
> Agree, they should be separate config class, each section should have their own class, and some of them are required, some of them are not, just like this `monitoring_address`, please stop try to unify it, this project should be fairly easy.
> 
> They should all have separate class like SchedulerConfig, WorkerConfig, ClientConfig.
> 
> in the end, when user combined commandline argument and toml config (like set union, but command line config will take precedence), then send to SchedulerConfig, WorkerConfig, ClientConfig for validate (required or not, type correct or not, sanity checks)

@sharpener6  @rafa-be  I have changed the entire logic. Now each module has their toml & config. No reuse across several configs. I was able to test all scaler_* py modules  successfully  except symphony because it throws below error:
image

scaler/config/config.py
rafa-be, line 1: This file is now pretty big.

I think it could be easily splitted, similarly to how we split things in other parts of the code base:

```
scaler/config
├── __init__.py
├── mixins.py  # Config, ConfigValue
├── zmq.py  # ZMQType, ZMQConfig
├── scheduler.py  # SchedulerConfig
...
```

README.md
```diff
@@ -126,7 +120,7 @@ $ scaler_scheduler tcp://127.0.0.1:2345
 Finally, start a set of workers (a.k.a. a Scaler *cluster*) that connects to the previously started scheduler:
 
 ```bash
-$ scaler_cluster -n 4 tcp://127.0.0.1:2345
+$ scaler_cluster -n 4 --scheduler-address "tcp://127.0.0.1:2345"
```
rafa-be, line 123: Can you double check these examples? 

I have the feeling these do not match the entry points.

scaler/config/cluster.toml
rafa-be, line 1: Why do we need all these toml files in the code base? Having these seem quite redundant to me, as all config classes have their own default values, that are basically the same.

I think there should only be a basic example in the documentation about how to write a config file, explaining to the users how to map the CLI parameter names to the toml variables.

I-Need-Sleep-Asap: Since the modules( ex: scheduler) never change/removed (most probably!), having one example per module is great since user dont have to create one (may be never!) which will reduce mistakes by users.  @rafa-be please lemme know your thoughts

scaler/config/config.py
rafa-be: @sharpener6 @gxuu input on this?

scaler/config/cluster.toml
I-Need-Sleep-Asap: Ok ill delete all toml files & add a section in README talking about toml file support

scaler/config/config.py
```diff
@@ -0,0 +1,494 @@
...
+
+def parse_capabilities(capability_string: str) -> Dict[str, int]:
+    """Parses a capability string like 'linux,cpu=4' into a dictionary."""
+    capabilities: Dict[str, int] = {}
+    if not capability_string:
+        return capabilities
+    for item in capability_string.split(","):
+        name, _, value = item.partition("=")
+        if value:
+            capabilities[name.strip()] = int(value)
```
gxuu: This doesn't really handle all cases. For example `linux,cpu=MostPowerful` gives `ValueError` iirc. The call stack seems to not handle this.

Who makes the input? If I get it correctly, it parses whatever user may have in a config file, right?

scaler/config/mixins.py
```diff
@@ -0,0 +1,98 @@
...
+    @abc.abstractmethod
+    def from_string(cls, value: str) -> Self:
+        pass
+
+    @abc.abstractmethod
+    def __str__(self) -> str:
+        pass
+
+
+class Config:
```
rafa-be, line 27:
```suggestion
class ConfigValue(metaclass=abc.ABCMeta):
```

scaler/config/config.py
```diff
@@ -0,0 +1,494 @@
...
+
+    def __str__(self) -> str:
+        return self.to_string()
+
+    def to_string(self) -> str:
+        return f"tcp://{self.host}:{self.port}"
+
+    @classmethod
+    def from_string(cls, value: str) -> "ObjectStorageConfig":
+        address_format = r"^tcp://([a-zA-Z0-9\.\-]+):([0-9]{1,5})$"
```
gxuu, line 243: The Object Storage Server really don't expect hostname now after the migration from `asio` to `ymq`. So this will have to change. It only expects dot-form ip address (namely `127.0.0.1`).

And for simple parse like this I would not choose to use regexp.

scaler/config/mixins.py
I-Need-Sleep-Asap: i already have `configvalue(abc)` in above lines of this file

scaler/config/config.py
I-Need-Sleep-Asap: I just moved existing code to new place. i didnt author this but ill fix it now. Thanks!

scaler/config/mixins.py
sharpener6: If this class is only for the classmethod, then you should make this function a pure function

sharpener6: I didn't understand why this Config class exists

scaler/config/config.py
rafa-be: Yeah indeed. Before it was fine because the function was being called from argparse, and the raised `ValueError` would be converted to a nicer CLI error message.

Now it will raise a `ValueError`, which isn't extremely bad, but maybe less friendly than the prevous CLI message.

However you might be able to restore the old behavior by providing the `from_string()` methods to argparse:

```python
parser.add_argument(
    "--per-worker-capabilities",
    "-pwc",
    type=WorkerCapabilities.from_string,
    default=WorkerCapabilities({}),
    help='comma-separated capabilities provided by the workers (e.g. "-pwc linux,cpu=4")',
)

Comment (I-Need-Sleep-Asap) [reply scaler/config/config.py]: ok ill add this now

File: scaler/config/mixins.py (hunk 8 above)
Comment (I-Need-Sleep-Asap) [reply scaler/config/mixins.py]: As per sharperner's suggestion, i removed the class entirely

**Review by rafa-be — DISMISSED** [review]

File: scaler/config/values.py [hunk 10]
```diff
@@ -0,0 +1,169 @@
...
+    def from_string(cls, value: str) -> Self:
+        pass
+
+    @abc.abstractmethod
+    def __str__(self) -> str:
+        pass
+
+
+@dataclasses.dataclass
+class ZMQConfig(ConfigValue):
```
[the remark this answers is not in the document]
I-Need-Sleep-Asap: I am putting all configs at one place. Seems like you want separate files for each config. Will work on it @sharpener6

scaler/config/worker_adapter_native.py
```diff
@@ -0,0 +1,42 @@
+import dataclasses
+from typing import Optional, Tuple
+
+from .values import ZMQConfig, ObjectStorageConfig, WorkerCapabilities
+
+from scaler.io import config as defaults
+
+
+@dataclasses.dataclass
+class NativeWorkerAdapterConfig:
```
sharpener6, line 10: this should move to `config/section/native_worker_adapter.py`

scaler/config/common.py
```diff
@@ -0,0 +1,29 @@
...
+            try:
+                capabilities[name.strip()] = int(value)
+            except ValueError:
+                raise ValueError(f"Invalid capability value for '{name}'. " f"Expected an integer, but got '{value}'.")
+        else:
+            capabilities[name.strip()] = -1
+    return capabilities
+
+
+class ZMQType(enum.Enum):
```
sharpener6, line 22: move to `config/types/zmq.py`

scaler/config/common.py
```diff
@@ -0,0 +1,29 @@
+import enum
+from typing import Dict
+
+
+def parse_capabilities(capability_string: str) -> Dict[str, int]:
```
sharpener6, line 5: move to `WorkerCapabitlities` class, make it class method as it ties to it

scaler/config/webui.py
```diff
@@ -0,0 +1,16 @@
+import dataclasses
+
+from .values import ZMQConfig
+
+
+@dataclasses.dataclass
+class WebUIConfig:
```
sharpener6, line 7: move to `config/section/webui.py`

scaler/config/top.py
```diff
@@ -0,0 +1,13 @@
+import dataclasses
+
+from .values import ZMQConfig
+
+
+@dataclasses.dataclass
+class TopConfig:
```
sharpener6, line 7: this should be moved to `config/section/top.py`

scaler/entry_points/cluster.py
```diff
@@ -1,109 +1,66 @@
...
-    DEFAULT_HARD_PROCESSOR_SUSPEND,
-    DEFAULT_HEARTBEAT_INTERVAL_SECONDS,
-    DEFAULT_IO_THREADS,
-    DEFAULT_NUMBER_OF_WORKER,
-    DEFAULT_PER_WORKER_QUEUE_SIZE,
-    DEFAULT_TASK_TIMEOUT_SECONDS,
-    DEFAULT_TRIM_MEMORY_THRESHOLD_BYTES,
-    DEFAULT_WORKER_DEATH_TIMEOUT,
-)
+from scaler.config.section.cluster import ClusterConfig
```
sharpener6, line 5: how this will work? you don't have __init__.py file in section

I-Need-Sleep-Asap: @sharpener6  Its not a requirement anymore to have __init__.py because of Implicit Namespace Packages (introduced as PEP420 for python 3.3 or later)
To avoid confusion, i have added __init__.py files now

docs/source/tutorials/configuration.rst
```diff
@@ -93,3 +93,71 @@ This can be set using the CLI:
...
+    * For example, the flag ``--num-of-workers`` becomes the TOML key ``num_of_workers``.
+    * You can discover all available keys by running any command with the ``-h`` or ``--help`` flag.
+
+Practical Scenarios & Examples
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Here is an example TOML file for Scaler's cluster.
+
+
+**cluster.toml**
```
sharpener6, line 130: `cluster.toml`? can you make it `example_config.toml`, and I asked them to be in the same toml file with different sections

rafa-be (approved):

rafa-be (approved):

rafa-be (approved):

PR #230, opened 2025-09-17
Rework github workflows

sharpener6:
- Rework all github workflows

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,8 +56,8 @@ scaler_native_worker_adapter = "scaler.entry_points.native_adapter:main"
 [project.optional-dependencies]
 uvloop = ["uvloop"]
 graphblas = ["python-graphblas", "numpy==2.2.6"]
-gui = ["nicegui[plotly]==2.24.1"]
-all = ["python-graphblas", "numpy==2.2.6", "uvloop", "nicegui[plotly]==2.24.1"]
+gui = ["nicegui[plotly]==2.24.2"]
+all = ["python-graphblas", "numpy==2.2.6", "uvloop", "nicegui[plotly]==2.24.2"]
 
 [tool.scikit-build]
 cmake.source-dir = "."


.github/actions/publish-pypi/action.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Build Wheel (Linux)
+      if: inputs.os == 'Linux'
+      shell: bash
+      run: python -m cibuildwheel --output-dir dist
+      env:
+        CIBW_BUILD: "*manylinux_x86_64"
+        CIBW_SKIP: "pp*"
+        CIBW_MANYLINUX_X86_64_IMAGE: "manylinux_2_28"
+
+#    - name: Publish to PyPI
```
gxuu, line 34: This probably should be uncommented?

rafa-be (approved):
Looks better.

Do you think we could run the documentation and wheel workflows on all pull requests (instead of on release only), skipping the publishing part?

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,43 @@
...
+permissions:
+  contents: read
+
+jobs:
+  build_and_test:
+    name: Build And Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-latest ] # , windows-latest ]  , macos-latest ]
```
gxuu, line 18: I think this will create a pending status on Github (build) that is never resolved even if tests completed successfully. Maybe worth considering.

CMakeLists.txt
```diff
@@ -48,30 +56,64 @@ else()
...
     add_compile_definitions(WIN32_LEAN_AND_MEAN)
     add_compile_options("/EHsc")
+   
+elseif(APPLE)
+    if(CMAKE_OSX_ARCHITECTURES MATCHES "arm64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "arm64|aarch64")
+        # Mac Arm
+        message(STATUS "macOS on ARM64")
+
+        add_compile_options(-Wall -Wextra)
+        add_link_options(-Wl,-rpath,@loader_path)
```
gxuu, line 73: Not all components need this option. For example `cc_ymq` does not need this. Only components that depend on other components need this. (Just a note)

sharpener6:
> Looks better.
> 
> Do you think we could run the documentation and wheel workflows on all pull requests (instead of on release only), skipping the publishing part?

I already did this

.github/actions/publish-pypi/action.yml
sharpener6: This publish-pypi is still not working

CMakeLists.txt
sharpener6: Will address in next PR then

1597463007 (approved):

Issue #213, opened 2025-09-02
Scaler cluster worker init script


sharpener6:
In scaler cluster, everytime when a processor starts, it should have the capability to import a python module (init step) before taking the tasks, this will enable the scaler worker's capability to able to

1. start an environment on demand 
3. preload some data/library when worker needs

initially, the cluster command argument should be `--preload "foo.bar:preload_function(arg1, arg2)"`

if preload is None, then it doesn't have any init step, if it's string, we need parse and execute it under the same python interpreter

In previous example, assume there is a `foo/bar.py` file, and inside defined a `preload_function(arg1, arg2)` function

You don't need implement following, following is for FYI:
Later when scheduler send start worker instruction to WorkerManagerAdaptor, it will take the preload string and execute it

PR #234, opened 2025-09-18
Cluster worker init script support via --preload

e117649:
Closes https://github.com/finos/opengris-scaler/issues/213


sharpener6:
Please also add section for the sphinx documentation

scaler/worker/agent/processor/processor.py
```diff
@@ -98,6 +100,15 @@ def __initialize(self):
 
         self.__register_signals()
 
+        # Execute optional preload hook if provided
+        if self._preload:
+            try:
+                from scaler.worker.preload import execute_preload
+
+                execute_preload(self._preload)
+            except Exception:
```
sharpener6, line 109: if failed to execute the preload, I think the worker should log and stopped and quit

tests/test_preload_module.py
```diff
@@ -0,0 +1,21 @@
+import logging
+
+# Global variable to test preload functionality
+PRELOAD_VALUE = None
+
+
+def setup_global_value(value: str = "default") -> None:
```
sharpener6, line 7: if this file is a test utility then move to `tests.utility.preload_module`

rafa-be (approved):

rafa-be (approved):

docs/source/tutorials/configuration.rst
```diff
@@ -80,6 +80,45 @@ For the list of available settings, use the CLI command:
...
 
+**Preload Hook**
+
+Workers can execute an optional initialization function before processing tasks using the ``--preload`` option. This enables workers to:
+
+* Set up environments on demand
+* Preload data, libraries, or models  
+* Initialize connections or state
+
+The preload specification follows the format ``module.path:function(args, kwargs)`` where:
```
gxuu, line 91: Why is it just one function? Wouldn't it be easier to just execute a file? I mean, from security pov they are the same, right?

scaler/worker/preload.py
```diff
@@ -0,0 +1,86 @@
...
+def _parse_preload_spec(spec: str) -> Tuple[str, str, Optional[List[Any]], Optional[Dict[str, Any]]]:
+    """
+    Parse 'pkg.mod:func(arg1, kw=val)' using AST.
+    Returns (module_path, func_name, args_or_None, kwargs_or_None).
+    If expression is a simple name (no args), returns args=None, kwargs=None.
+    """
+    if ":" not in spec:
+        raise PreloadSpecError("preload must be in 'module.sub:func(...)' format")
+
+    module_part, obj_expr = spec.split(":", 1)
```
gxuu, line 23: Are their sanity check against `module_part`? Say I input `:hello(1,pi,e)`, it's not a valid function, right?

scaler/worker/preload.py
```diff
@@ -0,0 +1,86 @@
...
+    Example: 'foo.bar:preload_function("a", 2)'
+    """
+    module_path, func_name, args, kwargs = _parse_preload_spec(spec)
+    logging.info("preloading: %s:%s with args=%s kwargs=%s", module_path, func_name, args, kwargs)
+
+    try:
+        module = importlib.import_module(module_path)
+    except ImportError:
+        if module_path.endswith(".py") and os.path.exists(module_path):
+            hint = f"Failed to find module. Did you mean '{module_path.rsplit('.', 1)[0]}:{func_name}'?"
```
gxuu, line 64: I think it would be more decent to special check against `module_path` being empty or not. Give a hint or something.

gxuu:
Can anyone please explain to me why we emphasize on the execution of _one_ function before worker loads?

Why is it better than just `exec` a file? There are many advantage with this approach -
a) we get rid of all the `spec` parsing logic, 
b) we don't stuff everything we want to do to an _init function_, 
c) We can choose to execute code in other language, if we have corresponding interpreter in our environment. We can execute code that is written in C++, should we decide that's a safe thing to do.

gxuu:
I talked with @sharpener6 and he says the function approach is preferred and the parsing logic fix can be pushed to next PR. Feel free to resolve my comments ( I have no permission doing that ).

sharpener6:
> Can anyone please explain to me why we emphasize on the execution of _one_ function before worker loads?
> 
> Why is it better than just `exec` a file? There are many advantage with this approach - a) we get rid of all the `spec` parsing logic, b) we don't stuff everything we want to do to an _init function_, c) We can choose to execute code in other language, if we have corresponding interpreter in our environment. We can execute code that is written in C++, should we decide that's a safe thing to do.

you just need 1 entry point that does execute once when processor get initialized, normally it is used to load or import some library or load very big data in memory, dask/ray they all have same functionalities

scaler/worker/agent/processor/processor.py
```diff
@@ -98,6 +100,16 @@ def __initialize(self):
 
         self.__register_signals()
 
+        # Execute optional preload hook if provided
+        if self._preload:
+            try:
+                from scaler.worker.preload import execute_preload
```
sharpener6, line 106: why import here?

rafa-be (approved):

sharpener6 (approved):

PR #250, opened 2025-09-26
Move YMQ Examples to `examples/`

magniloquency:


CMakeLists.txt
```diff
@@ -108,6 +108,7 @@ include_directories(${CAPNP_INCLUDE_DIRS})
 include_directories(${PROJECT_SOURCE_DIR})
 
 add_subdirectory(scaler)
+add_subdirectory(examples)
```
gxuu, line 111: Move it to the bottom of the file, place it together with `add_subdirectory(tests)`. Do this so that we don't build examples in the release build. Also this makes sure examples will not be included into the releases.

magniloquency: done

sharpener6:
looks good, I just have second thought, please rename the directory `cpp_examples` to `cpp`, so it will be `examples/cpp/ymq/...`

PR #266, opened 2025-10-02
Client and processor uses pyYMQ to connect to the OSS

rafa-be:
Rewrites the `SyncObjectStorageConnector` implementation to use pyYMQ instead of raw sockets.

There was an issue with how pyYMQ interact with signals and Python threads. Before, pyYMQ was raising an exception if any signal got raised during a blocking call (e.g. `recvMessage`). Now it follows Python's syscall behavior that **restores interrupted blocking operations unless an exception got raised in the signal handler**:

>  • **time.sleep**(secs)
>    Suspend execution of the calling thread for the given number of seconds. The argument may be a floating-point number to indicate a more precise sleep time.
>    **If the sleep is interrupted by a signal and no exception is raised by the signal handler, the sleep is restarted with a recomputed timeout.**

See [PEP475](https://peps.python.org/pep-0475/).


scaler/io/ymq/CMakeLists.txt
```diff
@@ -61,14 +61,6 @@ if(LINUX)
...
 
     add_library(py_ymq SHARED
-        pymod_ymq/async.h
-        pymod_ymq/bytes.h
-        pymod_ymq/exception.h
-        pymod_ymq/message.h
-        pymod_ymq/io_context.h
-        pymod_ymq/io_socket.h
-        pymod_ymq/utils.h
-        pymod_ymq/ymq.h
```
rafa-be, line 71: Any reason we were adding the header files to `add_library()`?

gxuu: I added most of the header files in ymq with `add_library`. This is to mainly satisfy `clangd`. Essentially they should be noop on code that's generated.

gxuu:
closes #240

scaler/io/ymq/pymod_ymq/ymq.h
```diff
@@ -17,11 +17,13 @@
 
 // First-party
 #include "scaler/io/ymq/error.h"
-#include "scaler/io/ymq/pymod_ymq/utils.h"
 
 struct YMQState {
-    int wakeupfd_wr;
-    int wakeupfd_rd;
+    PyThreadState* threadState;
```
magniloquency, line 22: I think this might need to be stored on the stack

The state obj is module-wide but thread state is well per-thread. I think it's possible that the GIL is released in a sync op, then maybe another thread enters a sync op and saves the thread state, in which case this would be overwritten and I think that would be bad.

I'm not 100% sure if that could actually happen but I think it's safer to store it on the stack, that's what the docs seem to suggest as well: https://docs.python.org/3/c-api/init.html#releasing-the-gil-from-extension-code

scaler/io/ymq/pymod_ymq/waiter.h
```diff
@@ -0,0 +1,116 @@
...
+    void signal()
+    {
+        if (eventfd_write(*_waiter, 1) < 0) {
+            throw std::runtime_error("Failed to signal waiter: " + std::string(std::strerror(errno)));
+        }
+    }
+
+    // Block until signaled through `signal()` and return `true`.
+    //
+    // If interrupted by an OS signal and an exception is raised by a Python signal handler, returns `false`.
```
magniloquency, line 63: This is genius. Thank you so much for debugging this

Note: You can resolve this comment

scaler/io/ymq/pymod_ymq/waiter.h
```diff
@@ -0,0 +1,116 @@
...
+            if (pfds[0].revents & POLLIN)
+                return true;  // signal() called
+
+            if (pfds[1].revents & POLLIN) {
+                // Interrupted by an OS signal. Run the Python signal handlers and wait again.
+                PyEval_RestoreThread(_ymqState->threadState);
+                if (PyErr_CheckSignals() == -1) {
+                    PyErr_SetString(
+                        *_ymqState->PyInterruptedExceptionType,
+                        "A synchronous YMQ operation was interrupted by a signal handler exception");
```
magniloquency, line 95: In this case I'm actually not sure if we need to raise an error here. My understanding is that the Python signal handler has raised an error, and therefore I think all we need to do if return. From the docs:

> if a Python signal handler raises an exception, the error indicator is set and the function returns -1 immediately

I think a good example of this is Ctrl+C -> KeyboardInterrupt

PR #269, opened 2025-10-02
Move Object Storage Server Tests to `tests/cpp`

magniloquency:


tests/cpp/CMakeLists.txt
```diff
@@ -1 +1,3 @@
+# this fetches Google Test, so it must be included first.
```
magniloquency, line 1:
```suggestion
# This fetches Google Test, so it must be included first.
```

sharpener6 (approved):

Issue #288, opened 2025-10-06
Support tags on Web UI


cloakedclock:
Web UI should color code the task stream based on the task's tags, if available. Task log should also display associated tags.

Color currently codes the task state: in-progress, success, failed, etc. This should be conveyed in a different way.

Example:

<img width="1463" height="701" alt="Image" src="https://github.com/user-attachments/assets/da603e27-8c85-441b-b354-0935839ec40d" />

A "+" hatching represents ongoing tasks, and "x" represents failed tasks. Solid filled bars represent successful tasks.

An issue with this is that for short tasks and short bars, the patterns are not clear. A colored outline may be necessary.

PR #289, opened 2025-10-06
webui: add tag support

cloakedclock:
This PR addresses https://github.com/finos/opengris-scaler/issues/288

Adds
- Tag-based coloring for stream page when tags are available
- Tags column in task log
- "x" hatching and red outline for failed tasks, replacing previous red fill
- "+" hatching and yellow outline for in-progress tasks, replacing previous yellow fill


scaler/utility/metadata/task_info_metadata.py
```diff
@@ -0,0 +1,36 @@
+import dataclasses
+import struct
+from typing import List
+
+
+@dataclasses.dataclass
```
rafa-be, line 6: I'm not sure the capabilities should get reported to the monitor through `metadata`, but instead through a `capabilitities` field in `StateTask`, similar to the one in `Task`:

```python
class StateTask(Message):
    @property
    def capabilities(self) -> Dict[str, int]:
        ...
```

rafa-be (changes requested):

Issue #335, opened 2025-10-10
Separate C++ and Python source files in the source tree


rafa-be:
Currently, the C++ and Python code are organized in the same directory tree. It would be desirable to re-organize the hierarchy so that the C++ and Python codes reside in their own directories.

Suggested new directory structure:

```
.
├── CMakeLists.txt
├── pyproject.toml
├── src
│   ├── cpp
│   │   ├── CMakeLists.txt
│   │   ├── object_storage
│   │   │   ├── CMakeLists.txt
│   │   │   └── object_manager.h
│   │   └── ymq
│   │       └── CMakeLists.txt
│   ├── run_cluster.py
│   ├── run_scheduler.py
│   └── scaler
│       ├── __init__.py
│       └── client
│           ├── __init__.py
│           └── client.py
└── tests
    ├── client
    │   └── test_client.py
    └── cpp
        ├── CMakeLists.txt
        ├── object_storage
        │   └── test_object_manager.cpp
        └── ymq
```

Issue #212, opened 2025-09-01
Reorganize unit-tests to match codebase structure


rafa-be:
Currently, all unit test files are in the root of the `tests/` directory. We should organize tests so that these mirror the `scaler/` codebase structure.

Current structure:

```
tests/
├── test_client.py
├── test_future.py
├── test_serializer.py
├── test_balance.py
├── test_cluster.py
├── test_ui.py
...
```

Improved structure:

```
tests/
├── client/
│   ├── test_client.py
│   ├── test_future.py
│   └── serializer/
│       └── test_serializer.py
├── cluster/
│   └── test_cluster.py
├── scheduler/
│   ├── test_balance.py
│   └── ...
├── ui/
│   └── test_ui.py
└── ...
```

PR #355, opened 2025-10-15
Reorganize source tree

e117649:
Closes https://github.com/finos/opengris-scaler/issues/335 and closes https://github.com/finos/opengris-scaler/issues/212


examples/cpp/ymq/automated_echo_client.cpp
```diff
@@ -7,10 +7,10 @@
...
 #include <string>
 
-#include "scaler/io/ymq/io_context.h"
-#include "scaler/io/ymq/io_socket.h"
-#include "scaler/io/ymq/simple_interface.h"
-#include "scaler/io/ymq/typedefs.h"
+#include "cpp/ymq/io_context.h"
+#include "cpp/ymq/io_socket.h"
+#include "cpp/ymq/simple_interface.h"
+#include "cpp/ymq/typedefs.h"
```
rafa-be: Could you make these `#include` omit the `cpp/` directory?

```suggestion
#include "ymq/io_context.h"
#include "mq/io_socket.h"
#include "ymq/simple_interface.h"
#include "ymq/typedefs.h"
```

That way these should match those of the source code.

src/cpp/object_storage/defs.h
```diff
@@ -2,7 +2,7 @@
 
 #include <memory>  // std::shared_ptr
 
-#include "scaler/io/ymq/bytes.h"
+#include "../ymq/bytes.h"
```
rafa-be: Not sure we want relative includes in C++

gxuu: +1.

examples/cpp/ymq/automated_echo_client.cpp
e117649: moved to src/cpp/scaler and only using absolute includes now

e117649:
> @e117649 Did you tried to build the wheel and install the wheel and import scaler working?

it is not working, I'll have it fixed in 10 minutes

src/cpp/scaler/io/ymq/CMakeLists.txt
```diff
@@ -83,6 +86,14 @@ if(LINUX)
         LINKER_LANGUAGE CXX
     )
 
+    # Copy .so file to source directory for development/testing
+    add_custom_command(TARGET py_ymq POST_BUILD
```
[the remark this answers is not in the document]
e117649: I thought there might be a reason for CMAKE_LIBRARY_OUTPUT_DIRECTORY to build to CMAKE_BINARY_DIR but since there isn't I set it to build directly to the source directory

PR #358, opened 2025-10-15
Fix C++ Warnings

magniloquency:
Closes #317

PR #375, opened 2025-10-17
Fix Errors on Windows introduced by #358

### LINKED ISSUE #376: ymq's test should be in the same directory
Currently we have `CMakeLists.txt` that looks like this:

```cmake
if(LINUX OR APPLE)
  # Add the directory for the C++ tests.
  add_subdirectory(cpp)

  # Add the new directory for io tests.
  add_subdirectory(io/ymq)
endif()

```

They should be merged. From the blame history it looks like we decided to move all C++ test to a subdirectory named `cpp`.

gxuu:
# READY FOR REVIEW

See commits title.

closes #376

### DIFF
--- a/scaler/ui/memory_window.py
+++ b/scaler/ui/memory_window.py
@@ -5,15 +5,12 @@ from nicegui import ui
 
 from scaler.protocol.python.message import StateTask
 from scaler.ui.setting_page import Settings
-from scaler.ui.utility import format_timediff, get_bounds, make_tick_text, make_ticks
+from scaler.ui.utility import format_timediff, get_bounds, make_tick_text, make_taskstream_ticks
 from scaler.utility.formatter import format_bytes
 from scaler.utility.metadata.profile_result import ProfileResult
 
-CHART_NAME = "Memory Usage"
+CHART_NAME = "Memory Usage (in bytes)"
 X_AXIS_GRID_LINES = False
-
-Y_AXIS_TICK_VALS = [0, 1024, 1024 * 1024, 1024 * 1024 * 1024, 1024 * 1024 * 1024 * 1024]
-Y_AXIS_TICK_TEXT = ["0", "1KB", "1MB", "1GB", "1TB"]
 Y_AXIS_GRID_LINES = True
 
 
@@ -28,7 +25,7 @@ class MemoryChart:
         self._start_time = datetime.datetime.now() - datetime.timedelta(minutes=30)
 
     def setup_memory_chart(self, settings: Settings):
-        with ui.card().classes("w-full").style("height: 30vh"):
+        with ui.card().classes("w-full").style("height: 600px"):
             self._plot_data = {
                 "type": "scatter",
                 "fill": "tozeroy",
@@ -55,8 +52,9 @@ class MemoryChart:
                         "zeroline": False,
                     },
                     "yaxis": {
-                        "tickvals": Y_AXIS_TICK_VALS,
-                        "ticktext": Y_AXIS_TICK_TEXT,
+                        "tickmode": "auto",
+                        "nticks": 4,
+                        "tickformat": "~s",
                         "autorange": True,
                         "automargin": True,
                         "rangemode": "nonnegative",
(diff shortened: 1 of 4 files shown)


tests/cpp/ymq/CMakeLists.txt
```diff
@@ -1 +1,10 @@
-add_test_executable(test_ymq test_ymq.cpp)
```
sharpener6: I think for all platforms, ymq tests should passed.

When add support for windows, it should make changes that will enable windows, then add matrix

So, please revert this

gxuu: The thing is when developing locally this creates error because this is Linux specific code (as of now).

We still want the tests that can be build (for example this `test_ymq_logging`)to build so that we can test along developing.

Windows build for YMQ has always been enabled and it didn't report problem only because it's not part of the CI.

So I think it's more reasonable to turn it on after `test_ymq` is cross platformed as well, preferably at the same time as CI for Windows is on. The name `test_ymq` may be confusing in cross platformed context, because now it really is just `test_ymq_linux`.

**Please read above**, if you still insist, then I will revert.

tests/CMakeLists.txt
```diff
@@ -32,6 +32,17 @@ function(add_test_executable test_name source_file)
     )
 
     add_test(NAME ${test_name} COMMAND ${test_name})
+
+    # Post build: We need to copy if dll is not present, otherwise the test program won't find gtest.dll
+    # See https://stackoverflow.com/questions/69978314/cmake-with-gtest-on-windows-build-starts-test-but-shared-libs-cannot-be-found
+    if (WIN32)
```
gxuu: Be noted that `gtest` has been set to always be linked dynamically. Also I think it's unwise to link statically when it comes to test cases. Not sure whether we still want to link everything statically. But this is an easier change to first make sure the test is runnable. Perhaps we revisit it later.

tests/cpp/ymq/CMakeLists.txt
sharpener6: You should read my above comment once again.

current workflow only tests with linux, if you want enable windows on CI, all ymq tests need fixed and tested on the windows. when adding matrix for windows, both linux and windows have to pass at the same time.

Windows build for ymq is NOT enabled right now, check the workflows again

tests/CMakeLists.txt
sharpener6: remove the static link for testing, anyway we are not packing it, as cibuildwheel will package libs, no need static linked needed and capnp doesn't work very on static linked

.clang-format
```diff
@@ -58,7 +58,7 @@ UseTab: Never
 # control function declaration/definition
 AllowShortFunctionsOnASingleLine: InlineOnly
 AllowAllParametersOfDeclarationOnNextLine: true
-BinPackParameters: OnePerLine
+# BinPackParameters: OnePerLine
```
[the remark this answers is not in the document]
sharpener6: I don't have any problem with this setting on vs code, and even this is a problem, you should open another PR, this is not related to the purpose of the PR

tests/CMakeLists.txt
```diff
@@ -20,24 +20,30 @@ function(add_test_executable test_name source_file)
 
     target_include_directories(${test_name} PRIVATE ${CMAKE_BINARY_DIR})
 
-    target_link_libraries(${test_name} PRIVATE 
-        $<$<CXX_COMPILER_ID:GNU,Clang>:-Wl,-Bstatic>
-        object_storage_server_objs
-        ymq_objs
+    target_link_libraries(${test_name} PRIVATE
         $<$<CXX_COMPILER_ID:GNU,Clang>:-Wl,-Bdynamic>
```
sharpener6: Just remove those static flags

sharpener6: no need, just delete those lines starts with `$<...`

gxuu: resolved.

sharpener6 (approved):

PR #382, opened 2025-10-19
webui: show worker capabilities

cloakedclock:
- capability color legend-style boxes next to worker names in task stream
- resize task stream based on number of rows
- make memory window larger for better visibility
- make memory window ticks dynamic

Note:
Worker's capabilities are not exposed by the scheduler. Currently associations are made when a task with certain capability requirements are routed to a worker, as that implies the worker supports those capabilities. This means workers only show their capabilities when they're used.
I think it's debatable whether this is desirable or if all capabilities should always be shown, so I avoided making scheduler changes for now.


cloakedclock:
Simple example:
taskstream_small

Many rows:
taskstream_many_rows

Many capabilities:
taskstream_many_capabilities


Memory chart:
memory_example

Capability legend overlapping with task bars:
capability_overlap

I initially tried putting this on the left of workers and pushing the graph right to fit, but I wasn't able to add hovertext without making it part of the graph. It's also difficult to put them below the axis labels and keep them aligned. I settled on this: a trace for each capability, with a white background for better visibility against task bars.

PR #397, opened 2025-10-24
Fix Windows YMQ bug

gxuu:
- Fixes expected WSAECONNABORTED lead to crash
- Fixes sendCursor might be too large
- Fixes too much async operation being posted to the kernel

### DIFF
--- a/scaler/io/ymq/message_connection_tcp.cpp
+++ b/scaler/io/ymq/message_connection_tcp.cpp
@@ -364,7 +364,13 @@ void MessageConnectionTCP::onRead()
         return;
     }
 
-    // The idea is, we only submit another pending operation if we have read some bytes previously.
+    // NOTE:
+    // Because there was no way to differentiate Read/Write operation on Windows, current behaviour is that whenever an
+    // operation arrives, I call both onRead and onWrite. This creates erroneous behaviour as we are posting too much
+    // operations into the queue maintained by the kernel.
+    // Sometimes, we don't really need to queued in another operation, as we typically know that when onRead is being
+    // called with no bytes being read, we know this is a false positive call (introduce by a write-available
+    // notification for example) and the previous ReadFile notification is still in the operating system's kernel.
     if (!_readSomeBytes) {
         return;
     }
@@ -397,7 +403,6 @@ void MessageConnectionTCP::onWrite()
         return;
     }
 
-
     auto res = trySendQueuedMessages();
     if (res) {
         updateWriteOperations(res.value());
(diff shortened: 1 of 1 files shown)


scaler/io/ymq/message_connection_tcp.cpp
```diff
@@ -187,7 +192,7 @@ std::expected<void, MessageConnectionTCP::IOError> MessageConnectionTCP::tryRead
             if (myErrno == WSAEWOULDBLOCK) {
                 return std::unexpected {IOError::Drained};
             }
-            if (myErrno == WSAECONNRESET || myErrno == WSAENOTSOCK) {
+            if (myErrno == WSAECONNRESET || myErrno == WSAENOTSOCK || myErrno == WSAECONNABORTED) {
```
gxuu: On Windows, the behaviour of connection aborting is not very clear - You can get `WSAECONNABORTED` (Note that `ECONNABORTED` is not presented on GNU) when the remote connection hit <ctrl-c>.

scaler/io/ymq/message_connection_tcp.cpp
```diff
@@ -357,6 +363,11 @@ void MessageConnectionTCP::onRead()
     if (!_connFd) {
         return;
     }
+
+    // The idea is, we only submit another pending operation if we have read some bytes previously.
+    if (!_readSomeBytes) {
```
gxuu: Because there was no way to differentiate Read/Write operation on Windows, current behaviour is that whenever an operation arrives, I call both `onRead` and `onWrite`. This creates erroneous behaviour as we are posting too much operations into the queue maintained by the kernel.

Sometimes, we don't really need to queued in another operation, as we typically know that when `onRead` is being called with no bytes being read, we know this is a false positive call (introduce by a write-available notification for example) and the previous `ReadFile` notification is still in the operating system's kernel.

scaler/io/ymq/message_connection_tcp.cpp
```diff
@@ -407,13 +419,14 @@ void MessageConnectionTCP::onWrite()
...
-        ++_sendCursor;  // Next onWrite() will not be called until the asyncop complete
 
-        const bool writeFileRes = WriteFile((HANDLE)(SOCKET)_connFd, addr, 1, nullptr, _eventManager.get());
+        const size_t len        = 1;
+        const bool writeFileRes = WriteFile((HANDLE)(SOCKET)_connFd, addr, len, nullptr, _eventManager.get());
         if (writeFileRes) {
             onWrite();
             return;
         }
+        updateWriteOperations(len);
```
gxuu: If you don't `updateWriteOperations`, the `_sendCursor` will not be reset and that breaks the assumption that `0 <= _sendCursor <= HEADER_SIZE + message.payload.length`

sharpener6: put those as comments in the code

PR #418, opened 2025-11-04
Let TcpServer use RawServerTCPFD 

gxuu:
# READY FOR REVIEW

This PR makes `TcpServer` to use the internal `RawServerTCPFD` such that all cross platform drudge work is inside `internal/`.


src/cpp/scaler/ymq/internal/raw_server_tcp_fd.h
```diff
@@ -0,0 +1,40 @@
...
+    RawServerTCPFD(sockaddr addr);
+    ~RawServerTCPFD();
+    void prepareAcceptSocket(void* notifyHandle);
+    std::vector<std::pair<uint64_t, sockaddr>> getNewConns();
+
+    bool setReuseAddress();
+    void bindAndListen();
+    auto nativeHandle() const noexcept
+    {
+#ifdef _WIN32
```
sharpener6: move this platform depended implementation into cpp file

src/cpp/scaler/ymq/internal/raw_server_tcp_fd.h
```diff
@@ -0,0 +1,40 @@
...
+#endif  //_WIN32
+#ifdef __linux__
+        return (int)_serverFD;
+#endif
+    }
+
+private:
+    uint64_t _serverFD;
+    sockaddr _addr;
+#ifdef _WIN32
```
sharpener6: Those you can keep it here, but in the next PR, you need make this RawServerTCPFD to be virtual class or template, so hide those system specific code to their class only

src/cpp/scaler/ymq/internal/raw_server_tcp_fd.h
gxuu: Currently it's like
`auto nativeHandle()  const noexcept { return (RawSocketType)_serverFD; }`

With respect to the system specific implementation instead of `ifdef` approach - let's put it on hold until the transition of the upper layer complete first.

PR #433, opened 2025-11-16
Fix .so Files in Wheel

### LINKED ISSUE #431: ModuleNotFoundError: No module named 'scaler.io.ymq._ymq' in wheel
Regression introduced in https://github.com/finos/opengris-scaler/pull/406 only occurs in wheel

```python
>>> import scaler
  File "<stdin>", line 1, in <module>
  File "/workspaces/scaler/venv/lib/python3.12/site-packages/scaler/__init__.py", line 4, in <module>
    from .cluster.cluster import Cluster
  File "/workspaces/scaler/venv/lib/python3.12/site-packages/scaler/cluster/cluster.py", line 10, in <module>
    from scaler.worker.worker import Worker
  File "/workspaces/scaler/venv/lib/python3.12/site-packages/scaler/worker/worker.py", line 19, in <module>
    from scaler.io.ymq import ymq
  File "/workspaces/scaler/venv/lib/python3.12/site-packages/scaler/io/ymq/ymq.py", line 15, in <module>
    from scaler.io.ymq._ymq import BaseIOContext, BaseIOSocket, Bytes, ErrorCode, IOSocketType, Message, YMQException
ModuleNotFoundError: No module named 'scaler.io.ymq._ymq'
```
also triggered by running `scaler_cluster` or `scaler_scheduler`

magniloquency:
closes #431


scripts/build.ps1
```diff
@@ -8,6 +8,7 @@ $OS = "windows"
 $ARCH = "x64"
 $BUILD_DIR = "build_${OS}_${ARCH}"
 $BUILD_PRESET = "${OS}-${ARCH}"
+$INSTALL_PREFIX = "src"
```
gxuu: I think instead of setting this, you should change `CMakePresets.json`. It's either that we should change all presets' install location to `${sourceDir}/src`, or write new presets and allow user to pass in the `BUILD_PRESET` of their choice.

I think this should not be hard coded. As it's been proven this source directory change can actually happen. :)

gxuu:
It turns out that `CMakePresets.json` was not changed when @e117649 updated the source directory in #355. As a result, files are installed to wrong locations. Let us fix it in this PR.

PR #441, opened 2025-11-20
Make code in YMQ Internal directory meet coding guidelines

gxuu:
# READY FOR REVIEW.

- Rename `TcpClient` to `TCPClient`, `TcpServer` to `TCPServer`, and `IocpContext` to `IOCPContext` to meet coding guidelines.
- Remove per-function ifdefs for YMQ internal directory

### DIFF
--- a/src/cpp/scaler/ymq/CMakeLists.txt
+++ b/src/cpp/scaler/ymq/CMakeLists.txt
@@ -15,7 +15,6 @@ add_library(ymq_objs OBJECT
     event_loop_thread.cpp
 
     event_manager.h
-    # file_descriptor.h
 
     message_connection_tcp.h
     message_connection_tcp.cpp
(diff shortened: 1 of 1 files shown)


src/cpp/scaler/ymq/internal/raw_server_tcp_fd.cpp
```diff
@@ -1,116 +1,12 @@
...
-#ifdef _WIN32
-        CancelIoEx((HANDLE)_serverFD, nullptr);
-#endif
-        CloseAndZeroSocket(_serverFD);
-    }
-}
-
 bool RawServerTCPFD::setReuseAddress()
 {
     int optval = 1;
```
sharpener6: you can make those implementation as pure function as `bool set_reuse_address(fd, ...)`, and put those into `utility.h` under same directory, then for both `raw_server_tcp_fd_linux.cpp` and `raw_server_tcp_fd_windows.cpp`, do the same implemenation:
```cpp
bool RawServerTCPFD::setReuseAddress()
{
    return set_reuse_address(_serverFD, ...);
}
```

src/cpp/scaler/ymq/internal/raw_server_tcp_fd.cpp
```diff
@@ -20,6 +20,7 @@ bool RawServerTCPFD::setReuseAddress()
 void RawServerTCPFD::bindAndListen()
 {
     if (bind(_serverFD, &_addr, sizeof(_addr)) == -1) {
+        const auto serverFD = _serverFD;
```
gxuu: Before this change, because `closeAndZeroSocket` will zero the socket, the value printed out from `unrecoverableError` right below it will always print 0 being the `serverFD`. This change ensures we don't have this misleading error message.

sharpener6 (approved):

Issue #407, opened 2025-10-31
Pymod interface for YMQ is not in a namespace


gxuu:
Ideally, this code should be in the `scaler::ymq` namespace.

PR #443, opened 2025-11-21
Refactor YMQ utility classes

rafa-be:
These are small refactoring I had to do to the utility C++ classes to support macOS:
- simplified and more cross-platform `Timestamp` implementation;
- moved `EventManager`'s `#ifdef`s to their respective `EventLoop` instances;
- moved the `Pipe` classes to the main code base, as I'm going to reuse them for the new UNIX event loop implementation;
- add a fallback implementation of `std::move_only_function` as this is not well supported on all platforms yet;
- created a `scaler::utility` namespace for all these utility classes.

The new UNIX event pool will come in the next PR.

### DIFF
--- a/src/cpp/scaler/utility/io_helpers.cpp
+++ b/src/cpp/scaler/utility/io_helpers.cpp
@@ -9,7 +9,7 @@
 namespace scaler {
 namespace utility {
 
-IOResult readExact(std::span<uint8_t> buffer, std::function<IOResult(const std::span<uint8_t>&)> reader) noexcept
+IOResult readExact(std::span<uint8_t> buffer, std::function<IOResult(std::span<uint8_t>)> reader) noexcept
 {
     size_t cursor = 0;
 
--- a/src/cpp/scaler/utility/pipe/pipe_reader.cpp
+++ b/src/cpp/scaler/utility/pipe/pipe_reader.cpp
@@ -3,6 +3,7 @@
 #include <cstdint>
 #include <span>
 
+#include "scaler/utility/io_helpers.h"
 #include "scaler/utility/io_result.h"
 
 namespace scaler {
@@ -11,18 +12,7 @@ namespace pipe {
 
 IOResult PipeReader::readExact(std::span<uint8_t> buffer) const noexcept
 {
-    size_t cursor = 0;
-
-    while (cursor < buffer.size()) {
-        IOResult result = this->read(buffer.subspan(cursor));
-        cursor += result.bytesTransferred;
-
-        if (result.error) {
-            return IOResult::failure(result.error.value(), cursor);
-        }
-    }
-
-    return IOResult::success(cursor);
+    return utility::readExact(buffer, [this](std::span<uint8_t> buffer) { return this->readBytes(buffer); });
 }
 
 }  // namespace pipe
--- a/src/cpp/scaler/utility/pipe/pipe_writer.cpp
+++ b/src/cpp/scaler/utility/pipe/pipe_writer.cpp
@@ -3,26 +3,17 @@
 #include <cstdint>
 #include <span>
 
+#include "scaler/utility/io_helpers.h"
 #include "scaler/utility/io_result.h"
 
 namespace scaler {
 namespace utility {
 namespace pipe {
 
-IOResult PipeWriter::writeAll(std::span<const uint8_t> buffer) const noexcept
+IOResult PipeWriter::writeAll(const std::vector<std::span<const uint8_t>>& buffers) const noexcept
 {
-    size_t cursor = 0;
-
-    while (cursor < buffer.size()) {
-        IOResult result = this->write(buffer.subspan(cursor));
-        cursor += result.bytesTransferred;
-
-        if (result.error) {
-            return IOResult::failure(result.error.value(), cursor);
-        }
-    }
-
-    return IOResult::success(cursor);
+    return utility::writeAll(
+        buffers, [this](const std::vector<std::span<const uint8_t>>& buffers) { return this->writeBytes(buffers); });
 }
 
 }  // namespace pipe
--- a/examples/cpp/ymq/ipc_echo_client.cpp
+++ b/examples/cpp/ymq/ipc_echo_client.cpp
@@ -6,12 +6,13 @@
 #include <string>
 #include <thread>
 
-#include "scaler/error/error.h"
+#include "scaler/utility/error.h"
 #include "scaler/ymq/io_context.h"
 #include "scaler/ymq/io_socket.h"
 #include "scaler/ymq/simple_interface.h"
 #include "scaler/ymq/typedefs.h"
 
+using namespace scaler::utility;
 using namespace scaler::ymq;
 
 int main()
--- a/examples/cpp/ymq/ipc_echo_server.cpp
+++ b/examples/cpp/ymq/ipc_echo_server.cpp
@@ -4,11 +4,12 @@
 #include <iostream>
 #include <memory>
 
-#include "scaler/error/error.h"
+#include "scaler/utility/error.h"
 #include "scaler/ymq/io_context.h"
 #include "scaler/ymq/io_socket.h"
 #include "scaler/ymq/simple_interface.h"
 
+using namespace scaler::utility;
 using namespace scaler::ymq;
 
 int main()
--- a/src/cpp/scaler/utility/io_helpers.h
+++ b/src/cpp/scaler/utility/io_helpers.h
@@ -12,7 +12,7 @@ namespace utility {
 // Attempt to read data into the specified buffer using the provided reader function.
 //
 // It continues reading until the buffer is completely filled or an error occurs.
-IOResult readExact(std::span<uint8_t> buffer, std::function<IOResult(const std::span<uint8_t>&)> reader) noexcept;
+IOResult readExact(std::span<uint8_t> buffer, std::function<IOResult(std::span<uint8_t>)> reader) noexcept;
 
 // Writes all data from a sequence of byte spans using a provided writer function.
 //
--- a/src/cpp/scaler/utility/pipe/pipe_reader.h
+++ b/src/cpp/scaler/utility/pipe/pipe_reader.h
@@ -33,10 +33,7 @@ public:
     PipeReader(const PipeReader&)            = delete;
     PipeReader& operator=(const PipeReader&) = delete;
 
-    // read up to buffer.size(), returning the number of bytes read.
-    IOResult read(std::span<uint8_t> buffer) const noexcept;
-
-    // read exactly buffer.size().
+    // Read up to buffer.size(), stopping on the first error. Return the total number of bytes written.
     IOResult readExact(std::span<uint8_t> buffer) const noexcept;
 
     // returns the native handle for this pipe reader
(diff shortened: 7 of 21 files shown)


rafa-be:
@gxuu 

- Concerning the `#ifdef` vs CMake conditionals, I am fine with both. What I find important is to keep independent `_windows` and `_unix` files, and to limit us to a single `#ifdef` per file. The IDE argument makes sense. 

- the `::pipe` namespace is because I'm trying to make the namespace structure match the directory structure.

src/cpp/scaler/utility/move_only_function.h
```diff
@@ -0,0 +1,61 @@
...
+// Use feature-test macro to detect support for std::move_only_function.
+// This works across GCC, Clang, and MSVC on all platforms.
+// Otherwise, we provide a basic implementation.
+#if defined(__cpp_lib_move_only_function) && __cpp_lib_move_only_function >= 202110L
+template <typename T>
+using MoveOnlyFunction = std::move_only_function<T>;
+#else
+
+template <typename R, typename... Args>
+class MoveOnlyFunction;
```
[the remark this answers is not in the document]
gxuu: Plus, the type hint is more ergonomic on my machine
```c++
  // MoveOnlyFunction<typename R>
  // MoveOnlyFunction<typename R, typename Args>
```

src/cpp/scaler/utility/move_only_function.h
```diff
@@ -0,0 +1,61 @@
...
+template <typename R, typename... Args>
+class MoveOnlyFunction<R(Args...)> {
+public:
+    MoveOnlyFunction() = default;
+
+    MoveOnlyFunction(MoveOnlyFunction&&) noexcept = default;
+
+    MoveOnlyFunction& operator=(MoveOnlyFunction&&) noexcept = default;
+
+    template <typename F>
```
gxuu: Let us constraint this type. On `gcc version 15.2.1 20251112 (GCC)`, `g++ -o func func.cpp -std=c++20`, when try to compile this statement:
`MoveOnlyFunction<void(int)> f = 3;`
The compilers give nasty errors:

```
func.cpp: In instantiation of ‘R MoveOnlyFunction<R(Args ...)>::CallableContainer<F>::operator()(Args ...) const [with F = int; R = void; Args = {int}]’:
func.cpp:39:7:   required from here
   39 |     R operator()(Args... args) const override { return std::invoke(f, std::forward<Args>(args)...); }
      |       ^~~~~~~~
func.cpp:39:67: error: no matching function for call to ‘invoke(int&, int)’
   39 |     R operator()(Args... args) const override { return std::invoke(f, std::forward<Args>(args)...); }
      |                                                        ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
func.cpp:39:67: note: there is 1 candidate
In file included from func.cpp:2:
/usr/include/c++/15.2.1/functional:119:5: note: candidate 1: ‘template<class _Callable, class ... _Args> constexpr std::invoke_result_t<_Fn, _Args ...> std::invoke(_Callable&&, _Args&& ...)’
  119 |     invoke(_Callable&& __fn, _Args&&... __args)
      |     ^~~~~~
/usr/include/c++/15.2.1/functional:119:5: note: template argument deduction/substitution failed:
In file included from /usr/include/c++/15.2.1/concepts:48,
                 from func.cpp:1:
/usr/include/c++/15.2.1/type_traits: In substitution of ‘template<class _Fn, class ... _Args> using std::invoke_result_t = typename std::invoke_result::type [with _Fn = int&; _Args = {int}]’:
/usr/include/c++/15.2.1/functional:119:5:   required by substitution of ‘template<class _Callable, class ... _Args> constexpr std::invoke_result_t<_Fn, _Args ...> std::invoke(_Callable&&, _Args&& ...) [with _Callable = int&; _Args = {int}]’
func.cpp:39:67:   required from ‘R MoveOnlyFunction<R(Args ...)>::CallableContainer<F>::operator()(Args ...) const [with F = int; R = void; Args = {int}]’
   39 |     R operator()(Args... args) const override { return std::invoke(f, std::forward<Args>(args)...); }
      |                                                        ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
func.cpp:39:7:   required from here
   39 |     R operator()(Args... args) const override { return std::invoke(f, std::forward<Args>(args)...); }
      |       ^~~~~~~~
/usr/include/c++/15.2.1/type_traits:3296:11: error: no type named ‘type’ in ‘struct std::invoke_result<int&, int>’
 3296 |     using invoke_result_t = typename invoke_result<_Fn, _Args...>::type;
      |           ^~~~~~~~~~~~~~~

```

If I understand correctly, this type should be constrained as:
```c++
  template <typename F>
  requires std::invocable<F, Args...> && std::convertible_to<std::invoke_result_t<F, Args...>, R>
```
Which seems to work on my machine. :)

Corresponding error message becomes:
```
func.cpp: In function ‘int main()’:
func.cpp:47:35: error: conversion from ‘int’ to non-scalar type ‘MoveOnlyFunction<void(int)>’ requested
   47 |   MoveOnlyFunction<void(int)> f = 3;
      |                                   ^
```
So it's more clear.

src/cpp/scaler/utility/pipe/pipe_utils_unix.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include "scaler/utility/error.h"
+#include "scaler/utility/pipe/pipe.h"
+
+namespace scaler {
+namespace utility {
+namespace pipe {
+
+std::pair<long long, long long> create_pipe()
+{
+    int fds[2];
```
gxuu: zero memory

rafa-be:
@gxuu @magniloquency @sharpener6 

I just pushed a new implementation of `Pipe` that **works with both blocking and non blocking syscalls**, overcoming concerns @gxuu had while trying to implement a common API for sockets.

The idea is that the cross-platform syscalls (`read()`, `write()`, `accept()` ...) always return this new [`IOResult`](https://github.com/rafa-be/scaler/blob/ac7903c1fc2305445e4d7fb8fd7aac50a52db29d/src/cpp/scaler/utility/io_result.h) object:

```c++
struct IOResult {
    enum class Error { WouldBlock, EndOfFile };

    std::optional<Error> error;
    size_t bytesTransferred;

    static IOResult success(size_t bytesTransferred = 0) { return IOResult(std::nullopt, bytesTransferred); }

    static IOResult failure(Error error, size_t bytesTransferred = 0) { return IOResult(error, bytesTransferred); }
};
```

For example the [`PipeReader`](https://github.com/rafa-be/scaler/blob/ac7903c1fc2305445e4d7fb8fd7aac50a52db29d/src/cpp/scaler/utility/pipe/pipe_reader.h) API becomes this:

```c++
class PipeReader {
public:
    // read up to buffer.size(), returning the number of bytes read.
    IOResult read(std::span<uint8_t> buffer) const noexcept;

    // read exactly buffer.size().
    IOResult readExact(std::span<uint8_t> buffer) const noexcept;

    const int64_t fd() const noexcept;

    void setNonBlocking() const noexcept;
};
```

Thus the pipe can now be used like this:

```c++
PipeReader reader;
    
// Blocking pipe
{
    std::array<uint8_t, message.size()> buffer;
    IOResult result = reader.readExact(buffer);
        
    if (result.error) {
        std::cerr << "Failed to read from pipe\n";
        std::terminate();
    }
}
    
// Non-blocking pipe
{
    reader.setNonBlocking();
    
    eventLoop.addFd(reader.fd());
    
    std::array<uint8_t, message.size()> buffer;
    IOResult result = reader.readExact(buffer);
    
    if (result.error == IOResult::Error::WouldBlock) {
        // The read didn't fully completed, but we still might have read some bytes.
        auto completedBuffer = std::span<>(buffer).first(result.bytesTransmitted);
        processIncompleteRead(completedBuffer);
        return;
    } else if (result.error) {
        std::cerr << "Failed to read from pipe\n";
        std::terminate();
    }
    
    processMessage(buffer);   
}
```

src/cpp/scaler/utility/pipe/pipe_reader.cpp
gxuu: IIRC @sharpener6 required to remove system independent implementation files. Do we still hold up to this rule, or we should move this to `pipe_reader_{system}.cpp`?

src/cpp/scaler/utility/pipe/pipe_writer.cpp
gxuu: Same here.

src/cpp/scaler/utility/io_result.h
```diff
@@ -0,0 +1,21 @@
...
+
+namespace scaler {
+namespace utility {
+
+// The result of a possibly non-blocking I/O action.
+struct IOResult {
+    enum class Error { WouldBlock, EndOfFile };
+
+    std::optional<Error> error;
+    size_t bytesTransferred;
```
gxuu: Should these variables be put after these `static` functions? What is required in the coding standard?

src/cpp/scaler/utility/io_helpers.cpp
```diff
@@ -0,0 +1,75 @@
...
+IOResult writeAll(
+    const std::vector<std::span<const uint8_t>>& buffers,
+    std::function<IOResult(const std::vector<std::span<const uint8_t>>&)> writer) noexcept
+{
+    if (buffers.empty()) {
+        return IOResult::success(0);
+    }
+
+    // Copies the buffers to a linked list, so that we can pop() the first items in O(1).
+    std::list<std::span<const uint8_t>> bufferList(buffers.cbegin(), buffers.cend());
```
gxuu: Another way to resolve this is to provide another argument to the `writer` function such that the first argument is separated from the rest. Imagine something like this:
```
writer(firstPointer, firstPointerSize, restPointers, startIndex);
```


The old implementation with prefix sum perform `O(log n)` look-up given `n` pointers. The complexity came from the reconstruction of `vector` in every iteration. Suppose the first write to `sendmsg` returns `(n - 1) * 2048+1` bytes with a list of pointers (there's `n` of such pointers) such that each pointer points to a buffer of length `1024`, then this implementation will do `n-1` removal of pointers while the prefix sum implementation will do roughly `log(n-1)` steps in binary search. With above `writer` signature, you can imagine the actual filled in argument to be somewhat like this:
```
writer((n-1)Pointer, 1, vec, vec.size())
```
through this we can avoid reconstructing the pointer holder in each iter.

But that complicated the code. If you are happy with current solution, resolve this comment directly.

examples/cpp/ymq/timestamp.cpp
```diff
@@ -1,15 +1,15 @@
-#include "scaler/ymq/timestamp.h"
+#include "scaler/utility/timestamp.h"
 
 #include <iostream>
 
 using namespace std::chrono_literals;
-using namespace scaler::ymq;
+using namespace scaler::utility;
```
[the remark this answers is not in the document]
gxuu: Some information here:
I use `using namespace xxx` in example files because:
- the example code would have less noise.
- the code is limited in scope and doesn't pollute the environment.


There are other places that use `using namespace xxx` in places that I believe are better to be removed. For example the Python Interface of YMQ is `using namespace xxx` in header files. Related issue: #407.

gxuu (approved):

src/cpp/scaler/utility/io_result.h
rafa-be: @sharpener6 ?

sharpener6:
@rafa-be , I think this one should get closed?

PR #446, opened 2025-11-24
Unix Domain Socket support for YMQ on Linux

### LINKED ISSUE #379: Consider adopting Ray's plasma in-memory object store architecture/implementation
Ray uses `/dev/shm` on Linux and `/tmp` on MacOS to share object data with processes on the same machine.

We could re-engineer OSS to support over the wire and also inter-process communication. Instead of having workers fetch the same data if they are on the same machine, we can optionally place a secondary OSS on the machine that acts as a cache backing the primary OSS. All data can be read through shared memory buffers greatly increasing IO performance and reducing network usage.

https://docs.ray.io/en/latest/ray-core/scheduling/memory-management.html

gxuu:
# READY FOR REVIEW.

- Currently operating on Linux only. The primary reason to have Unix Domain Socket is to boost speed between Object Storage Server and other components when running on the same machine. Since Object Storage Server is not supported on Windows, there seems to be no point to have it on Windows.
- Only sanity check has been passed. Meaning echo server and client works. Reconnect has different semantics than TCP connections, as the connection, once established, cannot die. The reconnect for Unix Domain Socket means when client tries to connect to the server prior to server established, the message passing should behave correctly.
- Currently Windows' corresponding implementation will just `assert(false)` and give user the reason why they failed.
- Resolved issue that client connect to an IPC address that no one is listening on (was getting `ECONNREFUSED`).
- User may `connectTo` one IP address + one IPC address, same for `bindTo`. This allows fast transmission when some workers are running locally and some workers running remotely.

closes #379

### DIFF
--- a/src/cpp/scaler/ymq/internal/raw_stream_client_handle_windows.cpp
+++ b/src/cpp/scaler/ymq/internal/raw_stream_client_handle_windows.cpp
@@ -38,8 +38,9 @@ RawStreamClientHandle::RawStreamClientHandle(SocketAddress remoteAddr)
 
     _impl->_connectExFunc = {};
     if (_impl->_remoteAddr.nativeHandleType() == SocketAddress::Type::IPC) {
-        std::cerr << "Hitting IPC Socket address type, not supported on this system!\n";
-        assert(false);
+        unrecoverableError({
+            Error::ErrorCode::IPCOnWinNotSupported,
+        });
     }
 
     auto tmp = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
--- a/src/cpp/scaler/ymq/internal/raw_stream_server_handle_windows.cpp
+++ b/src/cpp/scaler/ymq/internal/raw_stream_server_handle_windows.cpp
@@ -39,8 +39,9 @@ RawStreamServerHandle::RawStreamServerHandle(SocketAddress addr): _impl(std::mak
     _impl->_acceptExFunc = {};
     memset(_impl->_buffer, 0, sizeof(_impl->_buffer));
     if (_impl->_addr.nativeHandleType() == SocketAddress::Type::IPC) {
-        std::cerr << "Hitting IPC Socket address type, not supported on this system!\n";
-        assert(false);
+        unrecoverableError({
+            Error::ErrorCode::IPCOnWinNotSupported,
+        });
     }
 
     _impl->_serverFD = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);


src/cpp/scaler/ymq/network_utils.h
```diff
@@ -84,10 +117,34 @@ inline std::expected<sockaddr, int> stringToSockaddr(const std::string& address)
...
+        assert(tmp);
+        res._type              = SocketAddress::Type::TCP;
+        *(sockaddr*)&res._addr = std::move(tmp.value());
+        res._addrLen           = sizeof(sockaddr);
+    } else {
+        // Unreachable as we currently don't support other types of system socket
+        std::unreachable();
+    }
+    return res;
+}
```
rafa-be: Coudn't these be methods/constructors of `SocketAddress`?

src/cpp/scaler/ymq/network_utils.h
```diff
@@ -137,7 +199,13 @@ inline sockaddr getRemoteAddr(int fd)
...
 
-    return remoteAddr;
+    SocketAddress res {};
+
+    res._type    = remoteAddrLen == sizeof(sockaddr) ? SocketAddress::Type::TCP : SocketAddress::Type::IPC;
+    res._addrLen = remoteAddrLen;
+    memcpy(&res._addr, &remoteAddr, res._addrLen);
+
+    return res;
 }
```
rafa-be: Shouldn't these functions be methods of the `RawXXXHandle`s?

gxuu: I prefer it here.

- I tend to use in-class method when a routine cannot operate without some in-class data. In this case, we are not doing that. So this case behaves much like a pure function. I tend to use just a function for these cases.
- Because of how `RawXXXHandle` is operating now, if we put this as in class methods, then we would have to repeat ourselves twice. This is not ideal.

src/cpp/scaler/ymq/network_utils.h
gxuu: Well actually, I want those to be free-functions because
- `SocketAddress` is a POD class, that basically dislike constructors.
- I want those to be free function in the hope that they can be reused because the function is non-intrusive enough that it can be easily extracted out.

I don't have a strong opinion on this. If you have a strong opinion on this, please let me know. If you don't have a strong opinion on this, then I will just leave it as is.

src/cpp/scaler/ymq/internal/defs.h
```diff
@@ -63,5 +64,15 @@ using RawSocketType = int;
...
+
+// Windows doesn't define this symbol, but has support with AF_UNIX. So we define it ourselves.
+#ifdef _WIN32
+#define UNIX_PATH_MAX 108
+
+struct sockaddr_un {
+    ADDRESS_FAMILY sun_family;    /* AF_UNIX */
+    char sun_path[UNIX_PATH_MAX]; /* pathname */
+};
+#endif  // _WIN32
```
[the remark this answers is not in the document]
gxuu: If not satisfied, we should have this re-opened and apply your new suggestion.

src/cpp/scaler/error/error.h
```diff
@@ -56,7 +56,8 @@ struct Error: public std::exception {
             case ErrorCode::Uninit: return "";
             case ErrorCode::InvalidPortFormat: return "Invalid port format, example input \"tcp://127.0.0.1:2345\"";
             case ErrorCode::InvalidAddressFormat:
-                return "Invalid address format, example input \"tcp://127.0.0.1:2345\"";
+                return "Invalid address format, example input \"tcp://127.0.0.1:2345\" or "
+                       "\"IPC:///tmp/domain_socket_name.sock\"";
```
magniloquency:
```suggestion
                       "\"ipc:///tmp/domain_socket_name.sock\"";
```

sharpener6 (approved):

PR #450, opened 2025-11-25
YMQ Remove usage of defs.h

gxuu:
# READY FOR REVIEW.

This PR removes the usage of `defs.h` by adopting the PImpl Idiom. Most system dependent members are defined in the Implementation structure, in their corresponding implementation files.

### DIFF
--- a/src/cpp/scaler/ymq/internal/raw_stream_connection_handle_windows.cpp
+++ b/src/cpp/scaler/ymq/internal/raw_stream_connection_handle_windows.cpp
@@ -14,12 +14,6 @@
 #include <ws2tcpip.h> // inet_pton
 // clang-format on
 
-#undef SendMessageCallback
-#define __PRETTY_FUNCTION__ __FUNCSIG__
-#define EPOLLIN             (0)
-#define EPOLLOUT            (0)
-#define EPOLLET             (0)
-
 namespace scaler {
 namespace ymq {
 
--- a/src/cpp/scaler/ymq/internal/raw_stream_server_handle_windows.cpp
+++ b/src/cpp/scaler/ymq/internal/raw_stream_server_handle_windows.cpp
@@ -5,7 +5,6 @@
 #include "scaler/error/error.h"
 #include "scaler/ymq/internal/network_utils.h"
 #include "scaler/ymq/internal/raw_stream_server_handle.h"
-#include "scaler/ymq/internal/network_utils.h"
 
 // clang-format off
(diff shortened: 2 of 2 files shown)


rafa-be (approved):

src/cpp/scaler/ymq/internal/network_utils_windows.cpp
```diff
@@ -0,0 +1,162 @@
...
+namespace ymq {
+
+void closeAndZeroSocket(void* fd)
+{
+    uint64_t* tmp = (uint64_t*)fd;
+    closesocket(*tmp);
+    *tmp = 0;
+}
+
+std::expected<SocketAddress, int> stringToSockaddr(const std::string& address)
```
magniloquency: This function looks identical to the one implemented for Linux. Maybe we could have just one implementation?

src/cpp/scaler/ymq/internal/raw_stream_connection_handle_windows.cpp
```diff
@@ -3,10 +3,23 @@
...
+#include <winsock2.h>
+#include <mswsock.h>
+#include <ws2tcpip.h> // inet_pton
+// clang-format on
+
+#undef SendMessageCallback
+#define __PRETTY_FUNCTION__ __FUNCSIG__
+#define EPOLLIN             (0)
+#define EPOLLOUT            (0)
+#define EPOLLET             (0)
```
rafa-be: Looks like these defines are not used in this file.

src/cpp/scaler/ymq/internal/socket_address.h
```diff
@@ -0,0 +1,42 @@
...
+    int nativeHandleLen() noexcept;
+
+    friend void swap(SocketAddress& x, SocketAddress& y) noexcept
+    {
+        using std::swap;
+        swap(x._impl, y._impl);
+    }
+
+    struct Impl;
+    Impl* _impl;
```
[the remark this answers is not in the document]
gxuu: No. There's a constraint for using `unique_ptr` when applying pointer to implementation idiom. The `unique_ptr` requires the internal object's size must be defined in certain cases that I didn't dig into. This makes sense because `unique_ptr` does a little bit more than raw pointer.

I will try to replace it and see if it compiles. If it doesn't, I will go back and resolve this comment.

rafa-be (approved):

sharpener6 (approved):

PR #451, opened 2025-11-27
Refactor Testing Utils

magniloquency:
Based on #449

Refactors testing utils


tests/cpp/ymq/py_mitm/mitm_types.py
```diff
@@ -52,19 +52,19 @@ def rewrite(self, pkt: IP, ack: Optional[int] = None, data=None):
         )
 
 
-class AbstractMITMInterface(ABC):
+class MITMInterface(ABC):
```
magniloquency: Note: 'Interface' here means network interface

PR #461, opened 2025-12-10
Fix Test Client Close Connection

magniloquency:
Based on #451

The semantics of the test have been kept, except now that the server performs the disconnect instead of the server. This is because the server (a binder) does not throw an error when trying to recv with no peers, only the client (a connector) does.

### DIFF
--- a/tests/cpp/ymq/test_ymq.cpp
+++ b/tests/cpp/ymq/test_ymq.cpp
@@ -444,34 +444,34 @@ TestResult pubsub_publisher(std::string address, std::string topic, void* sem, i
     return TestResult::Success;
 }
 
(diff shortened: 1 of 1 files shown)


tests/cpp/ymq/net/i_socket.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <string>
+
+class ISocket {
```
sharpener6: All interfaces, no need named as `ISocket`, name as `Socket`, means more general name, for the implementation of this `Socket`, it can be `TCPSocket`, `UDPSocket` like this

PR #477, opened 2025-12-19
Introduce a libuv C++ wrapper

rafa-be:
This add a `scaler::uv` with a basic C++ RAII wrapper over libuv loop, signal, timer and async callback objects.

See examples/cpp/uv/simple_loop.cpp for an usage example.

Socket classes will come in a later PR.

### DIFF
--- a/src/cpp/scaler/object_storage/object_storage_server.cpp
+++ b/src/cpp/scaler/object_storage/object_storage_server.cpp
@@ -1,6 +1,7 @@
 #include "scaler/object_storage/object_storage_server.h"
 
 #include <algorithm>
+#include <csignal>
 #include <cstdint>
 #include <exception>
 #include <future>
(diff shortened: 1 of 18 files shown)


scripts/build.sh
```diff
@@ -2,18 +2,23 @@
 #
 # This script builds and installs in-place the C++ components.
 #
+# If `--clean` is specified, remove any existing cached build files before building.
+#
 # Usage:
-#      ./scripts/build.sh
+#      ./scripts/build.sh [--clean]
```
gxuu: do you think we should keep the build script interface consistent across different platform?

side note:
I have command like this stay in the terminal history so that I seldom run script when building:
`; cmake --build build; cmake --install build;`

maybe it would be of help to you as well.

scripts/library_tool.sh
```diff
@@ -39,7 +38,7 @@ PREFIX=$(readlink -f "${PREFIX}")
 mkdir -p "${PREFIX}/include/"
 
 show_help() {
-    echo "Usage: ./library_tool.sh [boost|capnp] [download|compile|install] [--prefix=DIR]"
+    echo "Usage: ./library_tool.sh [boost|capnp|uv] [download|compile|install] [--prefix=DIR]"
```
gxuu: `|uv` -> `|libuv` so that it matches what we actually take

tests/cpp/uv/test_uv.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+TEST_F(UVTest, Error)
+{
+    ASSERT_EQ(Error(UV_EAGAIN), Error(UV_EAGAIN));
+    ASSERT_NE(Error(UV_EINTR), Error(UV_EAGAIN));
+
+    ASSERT_EQ(Error(UV_EBUSY).name(), "EBUSY");
+    ASSERT_EQ(Error(UV_EPIPE).message(), "broken pipe");
+
+#ifdef _WIN32
```
gxuu: It's probably pedantic to ask, but does our coding guideline specifies the usage of ifdefs for test files?

@sharpener6

sharpener6 (approved):

PR #481, opened 2025-12-29
Reduce Algorithmic Complexity of `AsyncPriorityQueue`

### LINKED ISSUE #462: Lower algorithmic complexity of AsyncPriorityQueue
Our implementation of [`AsyncPriorityQueue`](https://github.com/finos/opengris-scaler/blob/main/src/scaler/utility/queues/async_priority_queue.py) has a O(n) complexity on `remove()` and `decrease_priority()`, caused by a O(n) call to `List.index()`.

That queue and it's related [`AsyncSortedPriorityQueue`](https://github.com/finos/opengris-scaler/blob/main/src/scaler/utility/queues/async_sorted_priority_queue.py) are used:
- by the scheduler for sorting workers by load. Thus removing a worker or assigning a task to it has a O(n_workers) time complexity;
- by the worker for sorting tasks by priority (required for nested tasks). Thus receiving and finishing a task on the worker has a O(n_worker_tasks) complexity.

By avoiding these calls to `List.index()`, we could lower the complexity of these operation to O(1) or O(log n_workers).

gxuu:
# READY FOR REVIEW.

This PR addresses #462. `AsyncPriorityQueue` now uses `sortedcontainers.SortedDict` to store entries. With this PR:
- `decrease_priority` will have algorithmic complexity of `O(lg n)` instead of `O(n)`
- `remove` will have algorithmic complexity of `O(lg n)` instead of `O(n)`
- `max_priority_item` will have algorithmic complexity of `O(1)` instead of `O(n)`
- More test is added so that the behaviour of the `AsyncPriorityQueue` is better specified.

NOTES:
- It's wasn't clear from the usage and the original test whether repetitive of Priority is allowed.
    - With the new implementation, it is always allowed, and 
    - the order of insertion is preserved when inserting multiple elements with the same priority.
    - Because the order of insertion is preserved, `AsyncPriorityQueue` is now the same as `AsyncSortedPriorityQueue`.
- One thing that confused me was the semantic of `decrease_priority` and `max_priority_item`. Apparently "priority" can mean two different things. 
    - The priority of the item, and 
    - the priority value of the item
    - the word "priority" have the first meaning in `max_priority_item` and the second meaning in `decrease_priority`
    - I added a docstring in their implementation rather than changing their name, as my understanding to the word might not be the most common one.
- The member variable `self._queue` is actually a `sortedcontainers.SortedDict`. This is awkward, but in order for the inheritance from `asyncio.Queue` to work properly, we have to use this variable name. On the other hand, the implementation of `asyncio.Queue` doesn't assume `self._queue` to be an actual Queue. Therefore, it is safe to use a dict here.

Closes #462


src/scaler/utility/queues/async_priority_queue.py
```diff
@@ -1,70 +1,67 @@
...
 
     def __len__(self):
         return len(self._queue)
 
     def _init(self, maxsize):
-        self._queue: List[List] = []
+        # self._locator contains {data : List[[priority, data], count]}
+        # self._queue contains {(priority, count) : data}
         self._locator: Dict[bytes, List] = {}
+        self._queue = SortedDict()
```
rafa-be: You can add more complete typing:

```suggestion
        self._locator: Dict[bytes, List[Tuple[...]]] = {}
        self._queue: Dict[Tuple[...],  bytes] = SortedDict()
```

src/scaler/utility/queues/async_priority_queue.py
```diff
@@ -1,70 +1,67 @@
...
-        # this operation should be O(n), mark it as invalid entry in the heap is not good idea as those invalid
-        # entry will never get removed, so we used heapq internal function _siftdown to maintain min heap invariant
-        item = self._locator[data]
-        i = self._queue.index(item)  # O(n)
-        item[0] = self.__to_lower_priority(item[0])
-        heapq._siftdown(self._queue, 0, i)  # type: ignore[attr-defined]
+        """Decrease the priority *value* of an item in the queue, effectively move data closer to the front.
+
+        Notes:
+            - *priority* in the signature means the priority *value* of the item.
```
rafa-be: So this one decreases the user specified priority (the priority value), not the internal item count?

If yes, then I don't think you have to inform the user about these implementation details, and this could be better be a an inline comment (`#`), and not part of the user exposed docstring (`"""`).

rafa-be:
> Because the order of insertion is preserved, AsyncPriorityQueue is now the same as AsyncSortedPriorityQueue.

Would it make sense to remove `AsyncSortedPriorityQueue` then?

gxuu:
@rafa-be The new commit addressed your comments https://github.com/finos/opengris-scaler/pull/481/commits/948875d8f80ea3aa909d1ffd1aa4bf467e75f586

src/scaler/utility/queues/async_priority_queue.py
```diff
@@ -1,70 +1,83 @@
...
+
+        def __lt__(self, other):
+            return (self.priority, self.count) < (other.priority, other.count)
+
+        def __hash__(self):
+            return hash((self.priority, self.count))
+
+    @dataclass
+    class LocatorValue:
+        map_key: "AsyncPriorityQueue.MapKey"
```
rafa-be:
```suggestion
        map_key: AsyncPriorityQueue.MapKey
```

Doesn't that work too?

rafa-be (approved):
LGTM :)

src/scaler/utility/queues/async_priority_queue.py
gxuu: No.
```
line 29, in LocatorValue
    map_key: AsyncPriorityQueue.MapKey
             ^^^^^^^^^^^^^^^^^^
NameError: name 'AsyncPriorityQueue' is not defined
```

rafa-be (approved):

PR #486, opened 2026-01-06
Add sockets support to `scaler::uv`

rafa-be:
Add a `TCPSocket` and a `TCPServer` class.

For example, a very basic client :

```c++
Loop loop = Loop::init().value();

const SocketAddress address = SocketAddress::IPv4("127.0.0.1", 80).value();

auto onClientConnected = [&](std::expected<void, Error>&& result) {
    // Send the message to the server
    const std::vector<uint8_t> message {'h', 'e', 'l', 'l', 'o'};
    
    client.write(message, [](std::expected<void, Error>&& result) {
        std::cerr << "Message sent\n";
    });
};

client.connect(address, onClientConnected);

loop.run(UV_RUN_DEFAULT);
```

### DIFF
--- a/examples/cpp/wrapper/uv/simple_loop.cpp
+++ b/examples/cpp/wrapper/uv/simple_loop.cpp
@@ -12,24 +12,24 @@
 
 int main()
 {
-    scaler::uv::Loop loop = exitOnFailure(scaler::uv::Loop::init());
+    scaler::wrapper::uv::Loop loop = exitOnFailure(scaler::wrapper::uv::Loop::init());
 
     std::cout << "Event loop initialized successfully\n";
 
     // Setting up an Async callback
-    scaler::uv::Async async =
-        exitOnFailure(scaler::uv::Async::init(loop, []() { std::cout << "\tAsync callback executed!\n"; }));
+    scaler::wrapper::uv::Async async =
+        exitOnFailure(scaler::wrapper::uv::Async::init(loop, []() { std::cout << "\tAsync callback executed!\n"; }));
     exitOnFailure(async.send());
 
     // Setting up a 1 sec. repeating Timer
-    scaler::uv::Timer timer = exitOnFailure(scaler::uv::Timer::init(loop));
+    scaler::wrapper::uv::Timer timer = exitOnFailure(scaler::wrapper::uv::Timer::init(loop));
     exitOnFailure(timer.start(
         std::chrono::milliseconds(1000),  // Initial delay
         std::chrono::milliseconds(1000),  // Repeat every 1 second
         []() { std::cout << "\tTimer fired\n"; }));
 
     // Add a Signal handler that stops the loop on Ctrl+C
-    scaler::uv::Signal signal = exitOnFailure(scaler::uv::Signal::init(loop));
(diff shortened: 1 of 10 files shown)


src/cpp/scaler/uv/stream.h
```diff
@@ -0,0 +1,191 @@
...
+        handle().setData({});  // force destruction of the callback object
+    }
+
+    // See uv_write
+    //
+    // The buffers' content (inner std::span<uint8_t>) must remain valid until the callback is called. The user is
+    // responsible for freeing these buffers.
+    std::expected<WriteRequest, Error> write(
+        std::span<std::span<const uint8_t>> buffers, WriteCallback&& callback) noexcept
+    {
```
rafa-be: @gxuu I'am thinking about maybe using YMQ's Bytes class here instead, moving the buffer object instead of using `std::span`.

gxuu (commented):
- Currently all member variables in ymq are prefixed with `_`, this include POD types. Please change the variables of POD types to be prefixed with `_` as well.
- I see some overlap classes (e.g. `Error`, `SocketAddress`). Do you think it's a good idea of using these directly in ymq and remove the old ones?
- The namespace is `scaler::uv` which I think is better to be put inside `scaler::ymq::` when it acts as an underlying utility to `ymq` because other components should interact with `uv` through `ymq`.
- Many variables are not `const` when they are `const` variables. E.g. error code, written bytes, etc. Making these variables to be `const` would be nice.
- Some variables are not initialized. E.g. `char buffer[INET6_ADDRSTRLEN];`. Initialize them by suffixing `{}`.
- Why banning some constructors? Constructors are banned when you have to use `init` to report errors, that's fine; But some cases are not like this. For example, I should be able to create a `SocketAddress` that doesn't represent any addresses.
- Where is the utility to support IPC? I currently only see `tcp`. Beware that #379 demands IPC support. 

Other stuff LGTM

src/cpp/scaler/wrapper/uv/tcp.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+namespace scaler {
+namespace wrapper {
+namespace uv {
+
+namespace details {
+
+std::expected<SocketAddress, Error> getSockName(const uv_tcp_t& handle) noexcept
+{
+    sockaddr_storage storage;
```
gxuu: `storage{}`;

src/cpp/scaler/wrapper/uv/socket_address.cpp
```diff
@@ -0,0 +1,87 @@
...
+    if (err) {
+        return std::unexpected {Error {err}};
+    }
+
+    return SocketAddress(addr);
+}
+
+SocketAddress SocketAddress::fromSockAddr(const sockaddr* address) noexcept
+{
+    if (address->sa_family == AF_INET) {
```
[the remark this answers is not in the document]
rafa-be: Correct.

However libuv does not support `AF_UNIX` addresses, as these are platform specific. For this kind of IPC sockets, they use [another mechanism](https://docs.libuv.org/en/v1.x/pipe.html) that is cross platform.

So I am making anything not `AF_INET` or `AF_INET6` unreachable.

PR #494, opened 2026-01-08
Lint C++ in CI with clang-format

### LINKED ISSUE #470: Linter should react to incorrect formatting for C++ code
Currently, there are GitHub Actions that checks whether Python code is correctly formatted using `mypy`. Similar thing should be done to C++ code in this project. However, we lack one for now.

### A. Stop pipeline when `clang-format` reports reformat needs to be done.
Pros:
- User is aware that they need to re-format the code.
- User can choose to rebase and clear out useless reformat commit, good for git-history
- User can generally chose to not format some files (e.g. thirdparties code)

Cons:
- Not a fully automated process
- Confusing on output as `clang-format` doesn't tell you how you can fix it
- forcing user to use clang-format, which is probably not feasible to them

### B. Automatically patch a reformat commit with clang-format.
Pros:
- User doesn't need to install clang-format necessarily or use an interactive plugin
- User doesn't need to worry about formatting at all
- Fully automated process

Cons:
- git-history might have redundant commit that only contains reformat changes
- some files might be unexpectedly formatted (e.g. `concurrentqueue.h`) if we want non hacky implementation in the pipeline
- Pipeline will take longer as the process will be 1) try compiling 2) try linting 3) recompile

# NOTE:
- There should be no blockers on either approach. As it turns out, user may run `clang-format -i --Werror --dry-run -- filename` to achieve similar effect as running `mypy`.
- It's yet to be determined which approach to go with. I personally would go with option 2 as there's no reason to force developers to install extra dependencies as we would like to make community contributor's development experience smoother.

gxuu:
Close #470.

### DIFF
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -20,6 +20,10 @@ runs:
       run: |
         sudo apt update -y
         sudo apt install -y build-essential tzdata cmake clang curl pkg-config g++-14
+        wget https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        sudo ./llvm.sh 21
+        sudo apt install clang-format-21
         sudo chmod 755 ./scripts/library_tool.sh
         sudo chmod 755 ./scripts/build.sh
 
--- a/.github/actions/run-linter/action.yml
+++ b/.github/actions/run-linter/action.yml
@@ -32,3 +32,16 @@ runs:
       shell: bash
       run: |
         mypy --install-types --non-interactive .
+
+    - name: Lint with clang-format
+      if: inputs.os == 'Linux'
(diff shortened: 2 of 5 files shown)


.github/actions/setup-env/action.yml
```diff
@@ -20,6 +20,10 @@ runs:
       run: |
         sudo apt update -y
         sudo apt install -y build-essential tzdata cmake clang curl pkg-config g++-14
+        wget https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        sudo ./llvm.sh 21
+        sudo apt install clang-format-21
```
sharpener6: @gxuu how long this takes??

Issue #445, opened 2025-11-21
YMQ `removeIOSocket` implementation is bad


gxuu:
Currently the gist of that implementation is this piece of code:

```cpp
        auto eventLoopThread = rawSocket->_eventLoopThread;
        rawSocket->_eventLoopThread->_eventLoop.executeNow([=] { rawSocket->requestStop(); });
        std::promise<void> promise;
        auto future = promise.get_future();

        // NOTE: This `count` and `maxCount` is needed as a safety net so that
        // we don't wait forever on querying numOfConnections.
        // If the remote end is using YMQ as internal communication tool, then
        // we don't need this safety net. This is because YMQ closes a connection
        // if the remote end shutdown write. This results to an event in local.
        // If the remote end does not close connection upon local end shutdown
        // write, the local end will never get any event for remote socket close,
        // and therefore the connection will stay alive in the system.
        int count                 = 0;
        auto waitToRemoveIOSocket = [&](const auto& self) -> void {
            constexpr static int maxCount = 8;
            rawSocket->_eventLoopThread->_eventLoop.executeNow([&] {
                rawSocket->_eventLoopThread->_eventLoop.executeLater([&] {
                    if (rawSocket->numOfConnections() && count < maxCount) {
                        ++count;
                        self(self);
                        return;
                    }
                    rawSocket->_eventLoopThread->removeIOSocket(rawSocket);
                    promise.set_value();
                });
            });
        };
        waitToRemoveIOSocket(waitToRemoveIOSocket);

        future.wait();
```

In the ideal situation, the call sequence is this (ignoring thread context):
```cpp
rawSocket->requestStop();
if(rawSocket->numOfConnections() == 0) {
  _eventLoopThread->removeIOSocket(rawSocket);
}
```
whereas `requestStop` above forces every underlying connection to send `shutdown(connection, SHUT_WR)`.

Suppose Alice issued `removeIOSocket`, the flowchart looks like this:
```
Alice sends Shutdown Write to Bob
Bob receives Shutdown Write
Bob sends Shutdown Write
Bob closes the socket
Alice receives Shutdown Write
Alice closes the socket
Alice IOSocket destructs
removeIOSocket (Alice) returns
```

Dilemma:
- when Alice uses ymq and Bob uses raw tcp, Bob will not close connection when Alice send Shutdown Write to Bob. Thus connections stays alive, and Alice will not exit from `removeIOSocket` call.
[...]

PR #497, opened 2026-01-09
fix: Implement WorkerDisconnectNotification Protocol

### LINKED ISSUE #465: Workers do not acknowledge cancel request
The worker, when shutdown requested through signal handle, is sending a `DisconnectRequest`. It should wait for `DisconnectResponse` afterwards. This was never done. Current version ( `main` branch ) exits immediately after sending `DisconnectRequest`.

To prove this, we have this in the `scheduler` side,
```python
    async def on_disconnect(self, worker_id: WorkerID, request: DisconnectRequest):
        await self.__disconnect_worker(request.worker)
        await self._binder.send(worker_id, DisconnectResponse.new_msg(request.worker)) # <-- HERE
```

And the worker currently doesn't check `DisconnectResponse` message type. Because of race condition, the `worker` almost always exits before checking the last message from the `scheduler`, thus no exception was thrown.

There's a more detailed explanation in #430. This issue is also fixed by that PR. Please check accordingly.

magniloquency:
Closes #465

### Summary
- Replaces the `DisconnectRequest`/`DisconnectResponse` two-way handshake with a one-way `WorkerDisconnectNotification` (WDN). Workers send WDN during graceful shutdown; the scheduler immediately cleans up and re-queues in-flight tasks without waiting for heartbeat timeout.
- The old handshake was broken: the scheduler hung trying to send `DisconnectResponse` to an already-dead worker socket.

### Cap'n Proto union renumbering (breaking wire change)
Removing `disconnectRequest @11` and `disconnectResponse @12` from the `Message` union freed two slots, so **all union discriminants from slot 11 onward were renumbered** (shifted down by 2). `WorkerDisconnectNotification` occupies the new slot `@26`. Any peer built against the old schema will misread message types — both sides must be updated together.

| Old slot | New slot | Field |
|---|---|---|
| @11 | removed | `disconnectRequest` |
| @12 | removed | `disconnectResponse` |
| @13 | @11 | `stateClient` |
| @14 | @12 | `stateObject` |
| @15 | @13 | `stateBalanceAdvice` |
| @16 | @14 | `stateScheduler` |
| @17 | @15 | `stateWorker` |
| @18 | @16 | `stateTask` |
| @19 | @17 | `stateGraphTask` |
| @20 | @18 | `clientDisconnect` |
| @21 | @19 | `clientShutdownResponse` |
| @22 | @20 | `processorInitialized` |
| @23 | @21 | `informationRequest` |
| @24 | @22 | `informationResponse` |
| @25 | @23 | `workerManagerHeartbeat` |
| @26 | @24 | `workerManagerHeartbeatEcho` |
| @27 | @25 | `workerManagerCommand` |
| — | @26 | `workerDisconnectNotification` (new) |

### Changes
- **Protocol:** `WorkerDisconnectNotification` struct added (reusing the old `DisconnectResponse` struct body); `DisconnectRequest` struct removed. Python bindings and `worker.md` updated accordingly.
- **Worker:** Sends WDN at graceful-shutdown time across both YMQ and ZMQ backends, then tears down connectors. `__graceful_shutdown` and the `DisconnectResponse` handler removed; signal handling unified via `__destroy`.
- **Scheduler:** Routes WDN to `WorkerController.on_disconnect_notification`, which calls `__disconnect_worker` and returns with no reply. `on_disconnect` (the old `DisconnectRequest` handler) removed.

### Bug fix: UV_ENOTCONN crash in YMQ shutdown
When a worker sends WDN and immediately closes its TCP connection, the scheduler's subsequent `uv_shutdown()` call on the same socket races with the peer's close and can return `UV_ENOTCONN`. The original `shutdownClient()` passed this error unconditionally to `UV_EXIT_ON_ERROR`, which called `std::exit(1)`. The fix checks for `UV_ENOTCONN` and treats it as a successful disconnect — the connection is already gone, so there is nothing to shut down.


src/scaler/worker/worker.py
```diff
@@ -275,14 +269,7 @@ def __run_forever(self):
...
 
     def __register_signal(self):
-        backend = get_scaler_network_backend_from_env()
-        if backend == NetworkBackend.tcp_zmq:
-            self._loop.add_signal_handler(signal.SIGINT, self.__destroy)
-        elif backend == NetworkBackend.ymq:
-            self._loop.add_signal_handler(signal.SIGINT, lambda: asyncio.ensure_future(self.__graceful_shutdown()))
-
-    async def __graceful_shutdown(self):
-        await self._connector_external.send(DisconnectRequest.new_msg(self.identity))
```
gxuu: Are we not sending `WorkerDisconnectNotification` when receiving signals?

magniloquency: I changed it back to sending when the worker exits as it's just a notification now and no response is expected. For YMQ I found that sometimes it gets a `ConnectorSocketClosedByRemoteEnd` exception, so I've just added a check for that, but for some reasons it seems to hang sometimes (difficult to reproduce, seems kind of random). One time I also got a segmentation fault during the `test_graph_fail` test. But I'm not sure if these are related to my changes?

src/scaler/protocol/capnp/message.capnp
```diff
@@ -101,6 +101,10 @@ struct DisconnectRequest {
     worker @0 :Data;
 }
 
+struct WorkerDisconnectNotification {
+    worker @0 :Data;
```
sharpener6: move this struct below `DisconnectResponse`

src/scaler/protocol/worker.md
```diff
@@ -226,3 +226,13 @@ consider the worker dead.
 * worker: Worker ID
 
 When a `DisconnectRequest` message is sent, the worker should disconnect from the scheduler.
+
+### WorkerDisconnectNotification `WDN` (optional)
+
+| message_type | worker  |
+|:------------:|:-------:|
```
sharpener6: @magniloquency this md file need be updated, as they are using capnp protocol, this md is misleading, please create a new issue to remove those

magniloquency: I have reverted changes to this file, and created an issue: #503

src/scaler/protocol/capnp/message.capnp
magniloquency: done

src/scaler/worker/worker.py
magniloquency: I have attached gdb to the frozen process, here's the 10 most recent stack frames:

```
#0  0x00007fe6c1f1872d in syscall () from /usr/lib/libc.so.6
#1  0x00007fe6c0edf000 in std::__atomic_futex_unsigned_base::_M_futex_wait_until (this=<optimized out>, __addr=0x55f483ca0b60, __val=2147483648, __has_timeout=<optimized out>, __s=...,
    __ns=...) at /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/futex.cc:122
#2  0x00007fe6bfd42b49 in std::__atomic_futex_unsigned<2147483648u>::_M_load_and_test_until (this=0x55f483ca0b60, __assumed=0, __operand=1, __equal=true,
    __mo=std::memory_order::acquire, __has_timeout=false, __s=std::chrono::duration = { 0s }, __ns=std::chrono::duration = { 0ns }) at /usr/include/c++/15.2.1/bits/atomic_futex.h:111
#3  0x00007fe6bfd41636 in std::__atomic_futex_unsigned<2147483648u>::_M_load_and_test (this=0x55f483ca0b60, __assumed=0, __operand=1, __equal=true, __mo=std::memory_order::acquire)
    at /usr/include/c++/15.2.1/bits/atomic_futex.h:160
#4  0x00007fe6bfd3f2a0 in std::__atomic_futex_unsigned<2147483648u>::_M_load_when_equal (this=0x55f483ca0b60, __val=1, __mo=std::memory_order::acquire)
    at /usr/include/c++/15.2.1/bits/atomic_futex.h:218
#5  std::__future_base::_State_baseV2::wait (this=0x55f483ca0b50) at /usr/include/c++/15.2.1/future:362
#6  0x00007fe6bfd41512 in std::__basic_future<void>::wait (this=0x7ffcabc784d0) at /usr/include/c++/15.2.1/future:725
#7  0x00007fe6bfd3de02 in scaler::ymq::IOContext::removeIOSocket (this=0x55f47e3d9630, socket=std::shared_ptr<scaler::ymq::IOSocket> (empty) = {...})
    at /home/xxx/work/opengris-scaler/src/cpp/scaler/ymq/io_context.cpp:94
#8  0x00007fe6bfce63a1 in PyIOSocket_dealloc (self=0x7fe6abe72f40) at /home/xxx/work/opengris-scaler/src/cpp/scaler/ymq/pymod_ymq/io_socket.h:40
#9  0x00007fe6c22fb85d in _Py_DECREF (op=<optimized out>) at ./Include/object.h:500
#10 _Py_XDECREF (op=<optimized out>) at ./Include/object.h:567
```

It seems like `IOContext::removeIOSocket()` is hanging for some reason

magniloquency: maybe https://github.com/finos/opengris-scaler/issues/445 is related? total guess

gxuu: > maybe https://github.com/finos/opengris-scaler/issues/445 is related? total guess

Likely not. In fact that issue addressed a possible cause of hanging.

Inspecting the log, I see 
> #7  0x00007fe6bfd3de02 in scaler::ymq::IOContext::removeIOSocket (this=0x55f47e3d9630, socket=std::shared_ptr<scaler::ymq::IOSocket> (empty) = {...})

which is not valid because assumption is that the argument must represent a `shared_ptr` that's not empty.

gxuu (dismissed):

e117649 (changes requested):
1. We should probably bump the magic byte (`{'Y','M','Q',1}` -> `{'Y','M','Q',2}`)

2. Trust the binder source, not the payload

```python
async def on_disconnect_notification(self, worker_id: WorkerID, notification: WorkerDisconnectNotification):
    await self.__disconnect_worker(notification.worker)
```

`worker_id` (the identity the binder attaches to the connection) is ignored in favour of a field the
sender fills in. Any connected peer can disconnect any worker by name, which re-dispatches that
worker's in-flight tasks while it is still running them. `on_heartbeat` right above already trusts
the source, so this is inconsistent as well as looser.

The old handler had the same shape, so this isn't a regression — but the message is being redefined
here, which makes it cheap to fix: drop the `worker` field entirely and use `worker_id`. The
notification then carries no payload at all, which is what it actually means. Note that
`test_on_disconnect_notification_unknown_worker_is_safe` currently pins the payload-wins behaviour,
so it would need to flip with it.

3. A straggler heartbeat can resurrect the worker

Between the WDN send resolving and `self._task.cancel()` taking effect, the heartbeat routine can
still fire once. `on_heartbeat` -> `add_worker` re-registers the worker that just said goodbye, and
the scheduler is back to waiting out `worker_timeout_seconds` — the exact behaviour this PR removes,
now intermittent. The window is small, but a flag that no-ops the heartbeat routine before sending
the WDN would close it.

4. Shutdown can still hang, and a second signal doesn't help

`__notify_scheduler_and_stop` awaits a send with no timeout. If the connection is wedged rather than
closed, `__destroy()` never runs and SIGINT/SIGTERM appear to do nothing; a second signal just
schedules another coroutine that awaits another send. This looks like the intermittent
exit-time hang reported earlier in this PR's comments. An `asyncio.wait_for` around
`__graceful_shutdown()`, and/or having a repeat signal call `__destroy()` directly, would make the
worker always stoppable.

Smaller notes

* `test_graceful_shutdown_re_dispatches_task_immediately` uses 5.62s of its 10s budget on an idle
  8-core VM, and 5.0s of that is `noop_sleep(5)` itself — so the margin being squeezed is the
  re-dispatch plus three process spawns (manager, worker, processor). `noop_sleep(1)` with a 20-30s
  timeout tests the same distinction (WDN vs the 120s heartbeat timeout) with much more room on a
  loaded runner.
* The unit test covers the ZMQ path (`__teardown`); the YMQ path (`__notify_scheduler_and_stop`),
  which is the default backend, is covered only by the integration test.
* `docs/source/tutorials/additional_features.rst:25-27` still lists `DisconnectRequest` /
  `DisconnectResponse` in the sample `scaler top` output.
* Duplicated identically in `worker/worker.py` and `worker_manager_adapter/worker_process.py`:
  `__schedule_graceful_shutdown` / `__notify_scheduler_and_stop` / `__graceful_shutdown` are now
  the same three methods in both. Pre-existing duplication, but this PR grows it.

magniloquency:
@e117649:

1. done, bumped the magic to 2
2. makes sense, and removed worker id from the payload -- it's empty now
3. fixed by slightly changing when the notification is sent (it's back to pre-rebase behaviour, my mistake)
4. added a timeout

worker and worker process duplication: i think we should address this more broadly later on. in my opinion our "hpc" style of worker managers are under-specified

e117649 (dismissed):

src/scaler/worker/worker.py
```diff
@@ -44,6 +43,16 @@
...
 
+# YMQ errors that mean the scheduler connection is already gone.
+_EXPECTED_TEARDOWN_ERROR_CODES = frozenset(
+    {ymq.ErrorCode.ConnectorSocketClosedByRemoteEnd, ymq.ErrorCode.SocketStopRequested}
+)
+
+# How long teardown waits for the exit notification to be sent. The notification only saves the
+# scheduler from waiting out the heartbeat timeout, so it is never worth blocking our own exit on:
+# a connection that is wedged rather than closed would otherwise hang teardown indefinitely.
+_NOTIFY_SCHEDULER_TIMEOUT_SECONDS = 5
```
rafa-be: This should go to `scaler.config.defaults`.

src/scaler/worker_manager_adapter/worker_process.py
```diff
@@ -32,6 +31,16 @@
...
 
+# YMQ errors that mean the scheduler connection is already gone.
+_EXPECTED_TEARDOWN_ERROR_CODES = frozenset(
+    {ymq.ErrorCode.ConnectorSocketClosedByRemoteEnd, ymq.ErrorCode.SocketStopRequested}
+)
+
+# How long teardown waits for the exit notification to be sent. The notification only saves the
+# scheduler from waiting out the heartbeat timeout, so it is never worth blocking our own exit on:
+# a connection that is wedged rather than closed would otherwise hang teardown indefinitely.
+_NOTIFY_SCHEDULER_TIMEOUT_SECONDS = 5
```
rafa-be: Same as https://github.com/finos/opengris-scaler/pull/497/changes#r3895816163

rafa-be (approved):

Issue #316, opened 2025-10-09
Fix uses of `asyncio.get_event_loop()`


magniloquency:
While I was debugging YMQ I saw a few warnings from Python that we were calling this function while there was no current event loop. This will become a runtime error in Python 3.14 and this function will be removed in Python 3.16, so we need to find an alternative. `asyncio.get_running_loop()` may be suitable?

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop

PR #501, opened 2026-01-13
Fix issue on python>=3.14

sharpener6:


magniloquency:
It looks like this will close https://github.com/finos/opengris-scaler/issues/316

magniloquency (approved):

PR #553, opened 2026-02-10
Add timeout to future canceling

magniloquency:
When the client future manager cancels futures, it does so without a timeout. if a future is poorly behaved, this can cause the program to hang. This PR applies a 5s timeout to canceling all futures, and emits a warning if a future failed to cancel in time. this protects us against hangs in some cases


src/scaler/client/agent/future_manager.py
```diff
@@ -35,7 +35,11 @@ def cancel_all_futures(self):
 
         logging.info(f"canceling {len(futures_to_cancel)} task(s)")
         for future in futures_to_cancel:
-            future.cancel()
+            try:
+                future.cancel(timeout=5.0)
```
sharpener6: forgot where we should put global constant?

magniloquency: I've moved this to a constant defined at the top of this file. I considered putting it in defaults.py, but because "default" strongly implies that this value is configurable, and it's not, I decided against it

PR #554, opened 2026-02-10
Asyncio fixes

magniloquency:
This PR addresses several asyncio-related concerns:

- Always explicitly close asyncio event loops on exit
- Use call soon threadsafe to cancel tasks on the asyncio thread
- Check if the loop is closed before setting future status in ymq


src/scaler/client/agent/client_agent.py
```diff
@@ -119,7 +119,11 @@ def __initialize(self):
 
     def run(self):
         self._loop = asyncio.new_event_loop()
-        self._loop.run_until_complete(self._run())
+
```
sharpener6: can you refactor this to

```python
def run_task_forever(loop, task):
    ...

```

PR #568, opened 2026-02-18
Port existing MITM and integration tests to YMQ-libuv 

rafa-be:
This ports existing MITM and integration tests that were in the old YMQ implementation (`tests/cpp/ymq/test_ymq.cpp`) to YMQ-libuv.

Fixes couple of minor YMQ-libuv issues that were discovered (linked to IPC, addressing and TCP_NODELAY).

A lot of the remaining changes a were relating to the incorrect behavior of the MITM gateway on Windows (see issue #576).


tests/cpp/uv_ymq/test_sockets.cpp
rafa-be: This file contains the non-MITM tests from the old `test_ymq.cpp` file, without major change.

I separated the MITM from the non-MITM tests because macOS will only be able to execute the non-MITM tests.

tests/cpp/ymq/py_mitm/main.py
rafa-be: This file required a lot of changes because the exchanged TCP segments differ slightly between the YMQ and YMQ-libuv tests. In particular, the SYN and FIN/RST sequences were slightly different.

While this makes the tracking of the TCP state more robust, it's still far from properly implementing the TCP state machine.

PR #577, opened 2026-02-26
Fix MITM using unique source ports for each server-side connection

### LINKED ISSUE #576: YMQ MITM gateway on Windows does not handle the TCP state-machine properly
On Windows, the MITM gateway is [implemented using a dummy binding socket](https://github.com/finos/opengris-scaler/blob/main/tests/cpp/ymq/py_mitm/windivert.py). The client connects to this socket while the MITM script intercepts and rewrites packets, relaying them to the server.

TCP segments from the client have their tuple rewritten from `[client_ip, client_port, mitm_ip, mitm_port]` to `[mitm_ip, mitm_port, server_ip, server_port]`. Since `mitm_port` is static while `client_port` is dynamically allocated, the server sees all clients as originating from the same TCP stream.

**This breaks the MITM's support for multiple clients or re-connections**. When a client reconnects using a new `client_port`, the server still identifies it as the old client (same MITM→server tuple).

This causes TCP state machine corruption and breaks the re-connection test on YMQ-libuv:

Image

The original YMQ test has slightly different timing and error handling, and surprisingly finishes its execution normally. But looking at the exchanged TCP segments shows that the issue also occurs.

magniloquency:
## Summary

Fixes #576.

When the MITM forwarded client traffic to the server it always used \`mitm_port\` as the source port, so the server saw all clients (including reconnections) as the same TCP 4-tuple. This corrupted the server's TCP state machine on reconnection.

- Introduce a \`next_server_port\` counter initialised to \`mitm_port\` and incremented on each new client SYN, so every server-side connection gets a unique source port.
- Update the packet filter to accept packets on \`server_conn.local_port\` in addition to \`mitm_port\`, so server responses to the unique port are not dropped.
- Disable Python's stdout buffering in the embedded interpreter so that \`print()\` output from the MITM scripts is not silently dropped when forked child processes exit via \`std::exit()\` without calling \`Py_Finalize()\`. Uses \`Py_UnbufferedStdioFlag\` on Python < 3.12 and \`PyConfig.buffered_stdio\` on Python ≥ 3.12.
- Re-enable \`UVYMQMitmTest/Reconnect\` on Windows, which was disabled pending this fix.

Note: this fix handles sequential reconnections (the primary failure mode described in #576), but the MITM framework still only tracks a single \`client_conn\`/\`server_conn\` pair and does not support multiple simultaneous clients.

## Cleanup

- Consolidate repeated platform-specific IP selection (`#ifdef __linux__` / `_WIN32`) into a `getMitmIPs()` helper in `testing.h/cpp`, replacing four identical blocks in the MITM tests.
- Replace non-deterministic \`randomPort()\` calls for \`mitm_port\` with static unique ports (23575–23578), matching the pattern of the existing static \`remote_port\` values (23571–23574). Removes the now-unused \`randomPort()\` utility.

### DIFF
--- a/tests/cpp/uv_ymq/test_mitm.cpp
+++ b/tests/cpp/uv_ymq/test_mitm.cpp
@@ -136,15 +136,9 @@ TestResult reconnectClientMain(std::string address)
 // for more, see the python mitm files
 TEST_F(UVYMQMitmTest, Passthrough)
 {
-#ifdef __linux__
-    auto mitm_ip   = "192.0.2.4";
-    auto remote_ip = "192.0.2.3";
-#else
-    auto mitm_ip   = "127.0.0.1";
-    auto remote_ip = "127.0.0.1";
-#endif
-    auto mitm_port   = randomPort();
-    auto remote_port = 23571;
+    auto [mitm_ip, remote_ip] = getMitmIPs();
+    auto mitm_port            = 23579;
+    auto remote_port          = 23571;
 
     // the Python program must be the first and only the first function passed to test()
     // we must also pass `true` as the third argument to ensure that Python is fully started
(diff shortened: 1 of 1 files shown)


rafa-be (dismissed):
LGTM.

PR #578, opened 2026-02-26
Add Python wrappers for `scaler::uv_ymq`

rafa-be:


scripts/library_tool.sh
```diff
@@ -109,7 +109,7 @@ elif [ "$1" == "libuv" ]; then
         tar -xzf "${THIRD_PARTY_DOWNLOADED}/${UV_FOLDER_NAME}.tar.gz" -C "${THIRD_PARTY_COMPILED}"
 
         cd "${THIRD_PARTY_COMPILED}/${UV_FOLDER_NAME}"
-        cmake -B build -DCMAKE_INSTALL_PREFIX="${PREFIX}" -DBUILD_TESTING=OFF
+        cmake -B build -DCMAKE_INSTALL_PREFIX="${PREFIX}" -DBUILD_TESTING=OFF -DCMAKE_POSITION_INDEPENDENT_CODE=ON
```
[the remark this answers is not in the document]
rafa-be: Sadly, this one calls libuv's CMake scripts, not our. I'm not sure we can do it differently.

src/cpp/scaler/uv_ymq/pymod/address.h
```diff
@@ -0,0 +1,142 @@
...
+
+static int PyAddress_init(PyAddress* self, PyObject* args, PyObject* kwds)
+{
+    const char* addressStr = nullptr;
+    Py_ssize_t addressLen  = 0;
+    const char* kwlist[]   = {"address", nullptr};
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "s#", (char**)kwlist, &addressStr, &addressLen))
+        return -1;
+
+    auto result = scaler::uv_ymq::Address::fromString(std::string(addressStr, addressLen));
```
magniloquency: Hmm, it looks like an unnecessary copy is being made here?

src/cpp/scaler/uv_ymq/pymod/address.h
```diff
@@ -0,0 +1,142 @@
...
+    if (!result.has_value()) {
+        PyErr_SetString(PyExc_ValueError, result.error().what());
```
magniloquency: Shouldn't this be a ymq exception?

src/cpp/scaler/uv_ymq/pymod/address.h
```diff
@@ -0,0 +1,142 @@
...
+}
+
+static PyObject* PyAddress_fromAddress(UVYMQState* state, const scaler::uv_ymq::Address& address)
+{
+    if (!state)
+        return nullptr;
+
+    PyAddress* pyAddress = PyObject_New(PyAddress, reinterpret_cast<PyTypeObject*>(*state->PyAddressType));
+    if (!pyAddress)
+        return nullptr;
```
magniloquency: I think it's more usual to 'call' the type to create an instance (thus invoking its `__init__()` method). In Python you do this like `PyAddress(...)`, and here you can use the call APIs to do the same.

https://docs.python.org/3/c-api/call.html

magniloquency (commented):
A few nits but nothing major. Overall looks good

sharpener6 (approved):

PR #583, opened 2026-03-05
Merge FixedNativeWorkerAdapter into NativeWorkerAdapter

magniloquency:
## Dependencies

**Depends on #582** — this PR must be merged after #582.

---

## Summary

`FixedNativeWorkerAdapter` was originally forked from an old version of `NativeWorkerAdapter` and had diverged significantly. This PR merges the two by adding a `NativeWorkerManagerMode` enum (`DYNAMIC` / `FIXED`) to a unified `NativeWorkerAdapter` class, then deletes the old fixed-native files entirely.

---

### Mode enum

A `NativeWorkerManagerMode(enum.Enum)` with values `DYNAMIC` (default) and `FIXED` is added to `native_worker_adapter.py`. The config gains a `mode` field and a `-n`/`--num-of-workers` backward-compatibility alias for `--max-workers` (matching the old `FixedNativeWorkerAdapterConfig` CLI interface). The `configure_parser` method is typed with `parser: argparse.ArgumentParser`.

FIXED mode validates that `max_workers >= 0`. The threshold is `>= 0` rather than `>= 1` so that `SchedulerClusterCombo(n_workers=0)` continues to work — several tests create a combo with zero workers and then attach adapters manually.

### Config and TOML section rename (breaking change)

`NativeWorkerAdapterConfig` is renamed to `NativeWorkerManagerConfig` and the TOML section key changes from `[native_worker_adapter]` to `[native_worker_manager]`. Existing TOML configuration files must update this section header.

### Mode-specific `run()` paths; lazy ZMQ setup

The two modes take separate code paths in `run()`:

- **DYNAMIC mode** enters the existing async event loop via `_setup_zmq()` + `run_task_forever()`, unchanged from the original `NativeWorkerAdapter`.
- **FIXED mode** takes a completely separate, synchronous path (`_run_fixed()`): it spawns exactly `max_workers` worker subprocesses and blocks until they all exit. There is no event loop, no ZMQ connector, and no heartbeats — workers connect directly to the scheduler themselves. `register_event_loop` is called at the top of `_run_fixed()`, matching the DYNAMIC path behaviour. SIGTERM/SIGINT handlers terminate all living worker subprocesses on shutdown, preventing orphaned processes.

ZMQ objects (`context`, `connector`) are no longer created in `__init__`; they are created in `_setup_zmq()` called at the top of the DYNAMIC path. This keeps the adapter object picklable so callers can do `multiprocessing.Process(target=adapter.run).start()`. `_ident` is also initialised to `None` in `__init__` alongside other deferred ZMQ attributes, so it is always present on the object.

### `combo.py` subprocess pattern

```python
self._worker_adapter = NativeWorkerAdapter(config)
self._worker_adapter_process = multiprocessing.Process(target=self._worker_adapter.run)
self._worker_adapter_process.start()
```

The adapter instance is stored as `self._worker_adapter` (not discarded) so that tests can read config attributes such as `._address` from the main-process reference. The `is_alive()` guard in `shutdown()` handles the case where FIXED mode has already self-exited before teardown. The scheduler starts before the worker manager process so workers connect to a ready scheduler on first attempt.

### New `--worker-type` parameter

An optional `worker_type` field is added to `NativeWorkerManagerConfig` so callers can override the worker ID prefix (default: `NAT` for DYNAMIC mode, `FIX` for FIXED mode). `ECSWorkerAdapter` uses this to pass `--worker-type ECS`, making ECS workers identifiable in logs and monitoring. A clarifying comment is added to the `task_arn` field in `ECS WorkerGroupInfo` explaining why no worker ID tracking is needed.


design/583_merge-native-worker-managers/amendment-1.md
```diff
@@ -0,0 +1,223 @@
+# Amendment 1: Simplified FIXED Mode, Config Rename, and `--worker-type` Parameter
```
sharpener6: remove those

PR #618, opened 2026-03-19
C++ components build on macOS and Clang++ (fixes #614, #547)

### LINKED ISSUE #547: Simplify CMake building scripts for Python modules
We have a lot of duplicate code for building Python modules for the [OSS](https://github.com/finos/opengris-scaler/blob/81560c4957ead206b15d58520d2215a462042706/src/cpp/scaler/object_storage/CMakeLists.txt#L23-L50), [YMQ](https://github.com/finos/opengris-scaler/blob/81560c4957ead206b15d58520d2215a462042706/src/cpp/scaler/ymq/CMakeLists.txt#L79-L140) and [containers](https://github.com/finos/opengris-scaler/blob/81560c4957ead206b15d58520d2215a462042706/src/cpp/scaler/utility/CMakeLists.txt#L29-L119).

These account for a lot of the boilerplate code our `CMakeLists.txt` have, and could be simplified with a custom CMake function, e.g.:

```CMake
add_python_module(
    NAME <Python Module Name>
    TARGET <C++ object/library name>
    SOURCES <Source files>
    INSTALL_DESTINATION src/scaler/<directory>
)
```

### LINKED ISSUE #614: Scaler C++ components build on macOS.
Make the C++ components of Scaler (OSS and YMQ) build successfully on macOS.

rafa-be:
Makes the C++ components build on macOS and Clang.

Does not work with the built-in compiler as it's outdated (Clang 16.0). Homebrew's Clang works for me:

```bash
$ brew install llvm
$ export CC=/opt/homebrew/opt/llvm/bin/clang
$ export CXX=/opt/homebrew/opt/llvm/bin/clang++ 
$ export CPPFLAGS="-I/opt/homebrew/opt/llvm/include"
$ export CMAKE_PREFIX_PATH="/opt/homebrew/opt/llvm"
$ export LDFLAGS="-L/opt/homebrew/opt/llvm/lib/c++"
$ clang++ --version
Homebrew clang version 22.1.1
Target: arm64-apple-darwin24.1.0
Thread model: posix
InstalledDir: /opt/homebrew/Cellar/llvm/22.1.1/bin
Configuration file: /opt/homebrew/etc/clang/arm64-apple-darwin24.cfg
```

Also closes #547 by introducing a new CMake helper function `scaler_add_python_module()` to compile, link and install a Python library.

There is not yet any macOS support on the CI, and no Pypi build.

Also: **removes all `#ifdef _WIN32/__LINUX__`** in the main C++ code 🎉. 
There are still a few of these in the test code (mostly required by the MITM tests).

### DIFF
--- a/src/scaler/protocol/capnp/status.capnp
+++ b/src/scaler/protocol/capnp/status.capnp
@@ -55,10 +55,19 @@ struct WorkerManagerStatus {
 
 struct ScalingManagerStatus {
     managedWorkers @0 :List(Pair);
+    workerManagerDetails @1 :List(WorkerManagerDetail);
 
     struct Pair {
         workerManagerID @0 :Data;
         workerIDs @1 :List(Data);
+    }
+
+    struct WorkerManagerDetail {
+        workerManagerID @0 :Data;
+        identity @1 :Text;
+        lastSeenS @2 :UInt8;
+        maxTaskConcurrency @3 :UInt32;
+        capabilities @4 :Text;
     }
 }
 
--- a/src/scaler/protocol/python/status.py
+++ b/src/scaler/protocol/python/status.py
@@ -1,4 +1,4 @@
-from typing import Dict, List
+from typing import Any, Dict, List, Optional
 
 from scaler.protocol.capnp._python import _status  # noqa
 from scaler.protocol.python.common import TaskState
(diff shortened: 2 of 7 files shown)


tests/utility/test_one_to_many_dict.py
```diff
@@ -128,8 +128,8 @@ def test_iter_alive(self):
         count = 0
         for _ in self._dict:
             count += 1
-            for i in range(1, 12):
-                self._dict.add(i, i)
+            # Small number of keys should not trigger a rehash
+            self._dict.add("2", 1)
         self.assertTrue(count == 1)
```
rafa-be: @gxuu this test is very dependent on the STL implementation.

LLVM's lib C++ seems to extend the size of its hash table very early around 2 items, while GNU's does it around 16 items.

Maybe we don't even want to allow any edit to a `dict` while it's iterated. I don't see any advantage of allowing iteration on a data structure that might get invalidated on undefined behavior.

PR #630, opened 2026-03-24
Fix #629: conflicting directives on newer Clang-format versions.

### LINKED ISSUE #629: `.clang-format` is conflicting on newer versions of Clang-format
On newer versions of Clang-format, this parameter seems to conflict with the older versions:

```
```

The newer versions format short function like this:

```c++
size_t ObjectManager::size() const noexcept
{ return objectIDToHash.size(); }
```

But these will be declined by older versions (like on the CI) which expect this format:

```c++
size_t ObjectManager::size() const noexcept
{
    return objectIDToHash.size();
}
```

My local clang-format is v22.1.1:

```bash
$ clang-format --version
Homebrew clang-format version 22.1.1
```

rafa-be:
Fixes #629.


sharpener6:
close this one as #719 will address it

PR #666, opened 2026-04-02
fix: tear down all subprocesses when any one exits in scaler all

### LINKED ISSUE #657: scaler all-in-one entry point should crash if any subprocess crashes
## Problem

In `src/scaler/entry_points/scaler.py`, the `main()` function starts subprocesses (object storage, scheduler, worker managers, GUI) and then waits for them with a sequential `process.join()` loop:

```python
try:
    for process in processes:
        process.join()
except KeyboardInterrupt:
    for process in processes:
        process.terminate()
    for process in processes:
        process.join()
```

This has a critical flaw: if any subprocess crashes unexpectedly, the rest keep running. For example:
- The scheduler crashes → worker managers continue running, connecting to nothing.
- A worker manager crashes → the scheduler keeps running, waiting for workers that will never come.

There is also no check of `process.exitcode` after joining, so even the process that finishes first going undetected as a crash.

## Expected Behavior

If any subprocess exits (whether cleanly or with an error), the `scaler all` entry point should:
1. Detect the exited subprocess.
2. Terminate all remaining subprocesses.
3. Exit with a non-zero exit code if the subprocess that exited did so with an error.

## Suggested Fix

Poll all processes in a loop (e.g. using `process.is_alive()` or `multiprocessing.connection.wait`) and tear everything down as soon as one exits. Something like:

```python
import time

try:
    while True:
        for process in processes:
            if not process.is_alive():
                # A subprocess exited — tear everything down
                for other in processes:
                    if other.is_alive():
                        other.terminate()
                for other in processes:
                    other.join()
                sys.exit(process.exitcode or 0)
        time.sleep(1)
except KeyboardInterrupt:
    for process in processes:
        process.terminate()
    for process in processes:
        process.join()
```

magniloquency:
## Summary

- Replace the sequential `process.join()` loop with sentinel-based wait using `multiprocessing.connection.wait()`
- When any subprocess exits (crash or clean exit), terminate all remaining subprocesses and exit with the exited process's exit code
- This avoids polling with `time.sleep()` — `multiprocessing.connection.wait()` blocks on process sentinels and wakes immediately when any process exits

## Related Issue

Closes #657


src/scaler/entry_points/scaler.py
```diff
@@ -111,8 +112,15 @@ def main() -> None:
         processes.append(gui_process)  # type: ignore[arg-type]
 
     try:
-        for process in processes:
-            process.join()
+        sentinel_to_process = {p.sentinel: p for p in processes}
+        done = multiprocessing.connection.wait(sentinel_to_process)
+        exited = sentinel_to_process[cast(int, done[0])]
+        for other in processes:
+            if other.is_alive():
```
gxuu: I think terminating a dead process is no-op, so we don't have to test against aliveness.

src/scaler/entry_points/scaler.py
```diff
@@ -111,8 +112,15 @@ def main() -> None:
...
+                other.terminate()
```
magniloquency:
```suggestion
            other.terminate()
```

src/scaler/entry_points/scaler.py
magniloquency: I think you're right. Done

PR #667, opened 2026-04-02
fix: retry ObjectStorageServer connection on worker startup

### LINKED ISSUE #656: Worker crashes with ConnectionRefusedError on ObjectStorageServer connect in test_future tests
## Summary

7 tests in `tests.client.test_future.TestFuture` produce unhandled `ConnectionRefusedError` tracebacks from worker subprocesses on every CI run. The tests still pass, but the worker processes crash before doing any useful work, and the errors pollute test output.

Observed in: https://github.com/finos/opengris-scaler/actions/runs/23777912073/job/69283726105

## Root Cause

There is a race condition between worker subprocess startup and `ObjectStorageServer` readiness. The worker tries to connect to the object storage server immediately on startup, but the server hasn't started listening yet. There is no retry or backoff — the first connection attempt fails with `ConnectionRefusedError` and the worker process crashes.

Stack trace (repeated for each affected test):

```
File "scaler/worker/worker.py", line 249, in __get_loops
    await self._connector_storage.connect(self._object_storage_address.host, self._object_storage_address.port)
  File "scaler/io/async_object_storage_connector.py", line 44, in connect
    self._reader, self._writer = await asyncio.open_connection(self._host, self._port)
  ...
ConnectionRefusedError: [Errno 111] Connect call failed ('127.0.0.1', <port>)
```

## Affected Tests

All 7 `test_mocked_*` tests and `test_state` in `tests.client.test_future.TestFuture`:

- `test_mocked_cancel`
- `test_mocked_cancel_as_completed`
- `test_mocked_cancel_concurrent`
- `test_mocked_success`
- `test_mocked_task_failure`
- `test_mocked_worker_failure`
- `test_state`

## Expected Behavior

The worker should either:
- Wait/retry connecting to `ObjectStorageServer` until it is ready, or
- The test setup should ensure the `ObjectStorageServer` is fully started before spawning worker processes.

## Possible Fix

Add retry logic with backoff in `async_object_storage_connector.py`'s `connect()` method, or ensure proper startup ordering in the test fixtures so the server is ready before workers are spawned.

magniloquency:
## Summary

- Workers connecting to `ObjectStorageServer` with a manually-configured address would crash immediately on startup with `ConnectionRefusedError` if the server process hadn't finished binding its socket by the time the worker's first connection attempt ran.
- Added retry-with-backoff (0.1 s interval, 30 s timeout) to `PyAsyncObjectStorageConnector.connect()` so transient `ECONNREFUSED` errors during startup are handled gracefully rather than propagating and killing the worker process.
- The retry window matches the existing `ObjectStorageServerProcess.wait_until_ready()` timeout, providing consistent startup tolerance throughout the system.
- Also fixes a pre-existing bug in `PySyncObjectStorageConnector.__init__`: `_socket_lock` was initialized **after** `socket.create_connection()`, so if the connection raised (e.g. `ConnectionRefusedError`) the object was left partially constructed. When GC later collected it, `__del__` → `destroy()` → `with self._socket_lock` crashed with `AttributeError`. Fix: initialize `_socket = None` and `_socket_lock` before the connection attempt so `destroy()` is always safe to call.

## Related Issue

Closes #656


magniloquency:
stale

PR #721, opened 2026-04-15
Enable windows wheel release

sharpener6:


.github/workflows/publish-artifact.yml
```diff
@@ -1,6 +1,8 @@
 name: Publish Python package to PyPI
 
 on:
+  pull_request:
+    branches: [ "main" ]
```
gxuu: does this mean run publish for pr?

src/cpp/scaler/object_storage/io_helper_windows.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "scaler/object_storage/io_helper.h"
+
+namespace scaler {
+namespace object_storage {
+
+int getAvailableTCPPort()
+{
+    static bool winsockInitialized = [] {
+        WSADATA data {};
+        return WSAStartup(MAKEWORD(2, 2), &data) == 0;
```
gxuu: You need to call `WSACleanup` as well, which should be done right before `return`.

For each start up there needs to be a clean up.

pyproject.toml
```diff
@@ -115,6 +115,8 @@ build.targets = [
 [tool.scikit-build.cmake.define]
 CMAKE_C_COMPILER = { env = "CMAKE_C_COMPILER", default = "gcc" }
 CMAKE_CXX_COMPILER = { env = "CMAKE_CXX_COMPILER", default = "g++" }
+CMAKE_PREFIX_PATH = { env = "CMAKE_PREFIX_PATH", default = "" }
+CapnProto_DIR = { env = "CapnProto_DIR", default = "" }
```
gxuu, line 118: do we need to define these two variables here?

.github/workflows/publish-artifact.yml
sharpener6: for testing, will get removed once everything else is fine

src/cpp/scaler/object_storage/io_helper_windows.cpp
sharpener6: Done

gxuu (dismissed):

sharpener6:
@gxuu please review again

src/cpp/scaler/object_storage/io_helper_windows.cpp
```diff
@@ -0,0 +1,54 @@
...
+namespace object_storage {
+
+int getAvailableTCPPort()
+{
+    WSADATA data {};
+    if (WSAStartup(MAKEWORD(2, 2), &data) != 0) {
+        throw std::runtime_error("WSAStartup() failed");
+    }
+
+    struct WSACleanupGuard {
```
gxuu: Let's have a constructor and make it calls `WSAStartup`.

Should `WSAStartup` failed and throw, the destructor will not be called.

src/cpp/scaler/object_storage/io_helper_windows.cpp
```diff
@@ -0,0 +1,54 @@
...
+        ~WSACleanupGuard()
+        {
+            WSACleanup();
+        }
+    } wsaCleanupGuard;
```
gxuu: Add a line right below it:
```c++
(void) wsaCleanupGuard;
```
so we express the intent that this variable is not meant to be used.

src/cpp/scaler/object_storage/object_storage_server.cpp
```diff
@@ -103,45 +86,26 @@ void ObjectStorageServer::shutdown()
...
-        _logger.log(
-            scaler::ymq::Logger::LoggingLevel::error,
-            "ObjectStorageServer: create on server ready FDs failed, errno=",
-            errno);
-        std::terminate();
-    }
-
-    onServerReadyReader = pipeFds[0];
-    onServerReadyWriter = pipeFds[1];
+    std::lock_guard<std::mutex> lock {this->_serverReadyMutex};
```
gxuu: An advice: rename this `lock` to `guard` as that's what's commonly being named in public code base.

src/cpp/scaler/object_storage/object_storage_server.cpp
```diff
@@ -103,45 +86,26 @@ void ObjectStorageServer::shutdown()
...
-    ssize_t ret    = write(onServerReadyWriter, &value, sizeof(uint64_t));
-
-    if (ret != sizeof(uint64_t)) {
-        _logger.log(
-            scaler::ymq::Logger::LoggingLevel::error,
-            "ObjectStorageServer: write to onServerReadyWriter failed, errno=",
-            errno);
-        std::terminate();
+    {
+        std::lock_guard<std::mutex> lock {this->_serverReadyMutex};
```
gxuu: `guard`.

gxuu (approved):

PR #723, opened 2026-04-16
Add scaler_orb_ec2_config entry point for ORB EC2 setup

magniloquency:
## Summary

- Adds a new `scaler_orb_ec2_config` CLI command (`src/scaler/entry_points/orb_ec2_config.py`) that generates a ready-to-use `config.toml` for the ORB AWS EC2 worker manager
- Uses `ec2-metadata` to auto-detect the instance's public and private IPs, replacing the manual `<EC2_PUBLIC_IP>` / `<EC2_PRIVATE_IP>` substitution step from the quickstart docs
- Prints a post-generation message with the `scaler config.toml` launch command and an exact `Client(address=...)` snippet for connecting from a local machine
- Registered as a package entry point in `pyproject.toml`

## Related Issue

N/A


magniloquency:
TODO: We also want to run the gui in the ec2 instance. The gui has two addresses, the monitor address that it uses to connect to the scheduler for data, and the gui address that the user connects their web browser to view the webpage

magniloquency:
This is going to be replaced by the better solution proposed in https://github.com/finos/opengris-scaler/pull/757

PR #724, opened 2026-04-16
Refactor connectors and enable YMQ on client and worker

rafa-be:
This PR prepares Scaler for the removal of ZMQ with these changes:

- YMQ are now used by the client's agent and worker's agent;
- Refactor of the connector APIs:
    - Replace context object with the new `NetworkBackend` interface;
    - Standardize the connector APIs and remove any ZMQ specific gimmicks;
    - Simplify the `name` / `identity` logic with the new `generate_identity_from_name()` function;
- Replace `ZMQConfig` by a new `AddressConfig` class;
- Replace `ObjectStorageAddressConfig` by `AddressConfig`; 
- Fix various bugs:
    - YMQ wasn't correctly calling `sendMessage()` callbacks on shutdown, triggering an occasional error message when shutting down the OSS;
    - When using YMQ, the processor graceful shutdown wasn't working properly;
    - the worker agent had a concurrency issue that emerged with the use of YMQ;
    - YMQ reconnect timeout (2s) was too large and adding unnecessary delays to some tests.


src/cpp/scaler/ymq/configuration.h
```diff
@@ -11,8 +11,8 @@ namespace ymq {
 // Expect all connections to start with this string.
 constexpr std::array<uint8_t, 4> magicString {'Y', 'M', 'Q', 1};
 
-constexpr size_t defaultClientMaxRetryTimes = 4;
-constexpr std::chrono::milliseconds defaultClientInitRetryDelay {2000};
+constexpr size_t defaultClientMaxRetryTimes = 8;
+constexpr std::chrono::milliseconds defaultClientInitRetryDelay {100};
```
rafa-be, line 15: Some Python unit tests were taking more time with YMQ than expected because the client try to connect to the scheduler before it was booted up.

That triggered the 2 secs reconnect delay and added 2 seconds to the test.

src/cpp/scaler/object_storage/object_storage_server.cpp
```diff
@@ -156,7 +156,8 @@ void ObjectStorageServer::processRequests(std::function<bool()> running)
             for (auto& fut: _pendingSendMessageFuts) {
                 if (fut.wait_for(std::chrono::seconds(0)) == std::future_status::ready) {
                     auto result = fut.get();
-                    if (!result.has_value()) {
+                    if (!result.has_value() &&
+                        result.error()._errorCode != scaler::ymq::Error::ErrorCode::SocketStopRequested) {
                         _logger.log(
                             scaler::ymq::Logger::LoggingLevel::error,
                             "ObjectStorageServer: send message failed: ",
```
rafa-be, line 163: The OSS was occasionally printing error messages on some tests when shutting down, because some `send()` calls didn't completed yet.

src/cpp/scaler/ymq/pymod/binder_socket.h
rafa-be, line 1: These changes remove occasional SIGKILL warnings on the processor when running with YMQ:

```
Processor[xxxx] does not terminate in time, send SIGKILL.
```

The C++ `shutdown()` methods weren't not exposed to Python, as the wrapper was relying on `__del__()` to shutdown the socket.

However, [there is a location in the `Processor` SIGTERM handler](https://github.com/finos/opengris-scaler/blob/cef7933be31bb6c4d9bad8c63d8e256b5af9af75/src/scaler/worker/agent/processor/processor.py#L135-L137) where we explicitly shutdown the socket without de-allocating it, to stop any blocking `recv()` call, and allow the processor to gracefully exit.

The old behavior just deleted the reference to the socket, but that didn't trigger an immediate shutdown, as the blocking thread held another reference to the socket. Thus the processor soft interrupt didn't work and the worker agent hard killed it after a few seconds.

src/scaler/worker/agent/processor_manager.py
```diff
@@ -271,7 +270,12 @@ async def on_task_result(self, processor_id: ProcessorID, task_result: TaskResul
...
 
     async def on_external_object_instruction(self, instruction: ObjectInstruction):
-        for processor_id in self._holders_by_processor_id.keys():
+        processor_ids = list(self._holders_by_processor_id.keys())
+
+        for processor_id in processor_ids:
+            if processor_id not in self._holders_by_processor_id:
+                continue  # processor got killed while we were iterating over the list
+
             await self._binder_internal.send(processor_id, instruction)
```
rafa-be, line 279: For some reason, this bug only emerges when I switched the processor to use YMQ.

Maybe it only occasionally occurred on ZMQ because of timing.

src/scaler/worker/worker.py
```diff
@@ -276,23 +273,24 @@ async def __get_loops(self):
...
             await self.__graceful_shutdown()
 
         self._connector_external.destroy()
         self._processor_manager.destroy("quit")
         self._binder_internal.destroy()
         self._connector_storage.destroy()
-        os.remove(self._address_path_internal)
+
+        if self._address_internal.type == SocketType.ipc:
+            pathlib.Path(self._address_internal.host).unlink(missing_ok=True)
```
rafa-be, line 283: YMQ automatically removes the IPC socket on close, but ZMQ does not.

src/scaler/worker_manager_adapter/symphony/task_manager.py
```diff
@@ -56,8 +55,6 @@ def __init__(self, base_concurrency: int, service_name: str):
         self._processing_task_ids: Set[TaskID] = set()
         self._canceled_task_ids: Set[TaskID] = set()
 
-        self._object_storage_address: Optional[ObjectStorageAddress] = None
-
```
rafa-be, line 60: This variable was never used.

sharpener6 (approved):

PR #730, opened 2026-04-17
Refactor worker managers to use common components

### LINKED ISSUE #722: Refactor worker manager implementations to extract shared base classes
## Summary

The five worker manager implementations contain significant duplicated code across heartbeat management, task management, worker lifecycle, event loop setup, and signal handling. This refactor would centralize shared logic into base classes and mixins, reducing duplication by an estimated 50–65% and making each implementation focused on only its backend-specific behaviour.

## Affected Files

| File | Lines | Est. duplication |
|---|---|---|
| `worker_manager_adapter/symphony/worker_manager.py` | 177 | ~85% |
| `worker_manager_adapter/baremetal/native.py` | 237 | ~80% |
| `worker_manager_adapter/aws_raw/ecs.py` | 299 | ~75% |
| `worker_manager_adapter/orb_aws_ec2/worker_manager.py` | 552 | ~45% |
| `worker_manager_adapter/aws_hpc/worker_manager.py` | 37 | ~90% |
| `worker_manager_adapter/symphony/worker.py` | 235 | ~85% |
| `worker_manager_adapter/aws_hpc/worker.py` | 270 | ~85% |
| `worker_manager_adapter/symphony/heartbeat_manager.py` | 87 | ~95% |
| `worker_manager_adapter/aws_hpc/heartbeat_manager.py` | 117 | ~95% |
| `worker_manager_adapter/symphony/task_manager.py` | 293 | ~75% |
| `worker_manager_adapter/aws_hpc/task_manager.py` | 747 | ~75% |

### 1. Heartbeat managers (`symphony/` and `aws_hpc/`)

`SymphonyHeartbeatManager` and `AWSBatchHeartbeatManager` are ~95% identical. They share:

- `__init__` signature and field initialization
- `register()` — identical
- `on_heartbeat_echo()` — identical (latency calculation, timeout update, storage address init)
- `get_object_storage_address()` — identical
- `routine()` — identical structure; only `AWSBatchHeartbeatManager` adds a synthetic `ProcessorStatus` entry for the simulated processor

**Proposed fix:** Extract a `BaseHeartbeatManager` with a `_get_processor_statuses() -> list` hook (default `[]`).

### 2. Task managers (`symphony/` and `aws_hpc/`)

`SymphonyTaskManager` and `AWSHPCTaskManager` duplicate the full priority queue, semaphore, and task-state machinery (~75% of each file). Shared logic includes:

- `AsyncPriorityQueue` + `Semaphore` setup
- `on_task_new()` — identical priority bypass logic
- `on_cancel_task()` — identical
- `on_task_result()` — identical cleanup
- `get_queued_size()` / `can_accept_task()` — identical
- `process_task()` — identical semaphore-acquire flow
- `resolve_tasks()` — identical: serializer caching, object storage write, `TaskResult` construction

**Unique to `AWSHPCTaskManager`:** array-job batching (`_flush_pending_batch`, `_submit_array_job`, `_monitor_array_job`, etc.) and all boto3 job submission.

**Proposed fix:** Extract a `BaseTaskManager` with an abstract `_execute_task(task)` method.

### 3. Worker processes (`symphony/worker.py` and `aws_hpc/worker.py`)

`SymphonyWorker` and `AWSBatchWorker` are ~85% identical. Shared:

- `__init__`, `identity` property
- `run()` → `_run()` → `__initialize()` pipeline
- `__on_receive_external()` — identical dispatch and backoff queue logic
- `__get_loops()` — identical `asyncio.gather` pattern with `CancelledError` / `ClientShutdownException` handling
- `__register_signal()` / `__graceful_shutdown()` / `__destroy()` — identical

**Only difference:** which `HeartbeatManager` and `TaskManager` subclass is instantiated inside `__initialize()`.

**Proposed fix:** Extract a `BaseWorker(multiprocessing.Process)` with abstract `_create_heartbeat_manager()` and `_create_task_manager()` factory methods.

### 4. Worker manager event-loop boilerplate (all five managers)

Every manager duplicates:

- `run()` → `asyncio.new_event_loop()` + `run_task_forever`
- `__register_signal()` + `__destroy()` — identical signal handler setup
- `__send_heartbeat()` → `WorkerManagerHeartbeat(...)` — identical
- `__get_loops()` with `asyncio.gather` + `CancelledError` + `ymq.YMQException` handling
- `_handle_command()` template: dispatch on `startWorkers` / `shutdownWorkers`, send `WorkerManagerCommandResponse`

**Only difference per implementation:** `start_worker()` and `shutdown_workers()` logic.

**Proposed fix:** Extract a `BaseWorkerManager(ABC)` with abstract `start_worker()` and `shutdown_workers()`.

### 5. Worker-ID validation / shutdown pattern

The "validate all IDs exist, then kill and join" block is copy-pasted verbatim between `NativeWorkerManager` and `SymphonyWorkerManager`. This could be a small shared helper.

## Proposed Structure

```
worker_manager_adapter/
├── base_manager.py          # BaseWorkerManager (ABC)
├── base_worker.py           # BaseWorker (ABC, multiprocessing.Process)
├── base_heartbeat_manager.py # BaseHeartbeatManager (ABC)
├── base_task_manager.py     # BaseTaskManager (ABC)
├── baremetal/native.py      # NativeWorkerManager(BaseWorkerManager) — only spawn logic
├── symphony/
│   ├── worker_manager.py    # SymphonyWorkerManager(BaseWorkerManager) — SOAM session
│   ├── worker.py            # SymphonyWorker(BaseWorker) — factory methods only
│   ├── heartbeat_manager.py # SymphonyHeartbeatManager(BaseHeartbeatManager)
│   └── task_manager.py      # SymphonyTaskManager(BaseTaskManager) — SOAM execution
├── aws_hpc/
│   ├── worker_manager.py    # AWSHPCWorkerManager(BaseWorkerManager)
│   ├── worker.py            # AWSBatchWorker(BaseWorker) — factory methods only
│   ├── heartbeat_manager.py # AWSBatchHeartbeatManager(BaseHeartbeatManager)
│   └── task_manager.py      # AWSHPCTaskManager(BaseTaskManager) — array job batching
├── aws_raw/ecs.py           # ECSWorkerManager(BaseWorkerManager) — ECS API
└── orb_aws_ec2/
    └── worker_manager.py    # ORBAWSEC2WorkerAdapter(BaseWorkerManager) — ORB + EC2 API
```

## Expected Outcome

- Each implementation file contains only backend-specific logic
- Shared bugs (e.g. in heartbeat latency calculation or priority queue handling) are fixed once
- New worker manager backends require only implementing the abstract methods
- Test coverage of shared logic can be written against the base classes

magniloquency:
## Summary

- Replaces duplicated logic across the five worker manager adapter implementations (symphony, aws_hpc, aws_raw/ecs, baremetal, orb_aws_ec2) with a composition pattern, following the project guideline of preferring composition over inheritance
- Introduces three ABCs in `mixins.py` that backends implement:
  - **`ProcessorStatusProvider`**: backends supply processor status entries injected into heartbeats
  - **`ExecutionBackend`**: backends implement `execute()`, `on_cancel()`, `on_cleanup()`, `routine()`, and `register()` for task dispatch; `TaskInputLoaderMixin` provides a default `register()` implementation that stores the `load_task_inputs` callable injected by `TaskManager`
  - **`WorkerProvisioner`**: backends implement `start_worker()` and `shutdown_workers()` for dynamic scaling
- Introduces four concrete shared classes that wire the above together:
  - **`HeartbeatManager`**: handles heartbeat send/receive logic using a `ProcessorStatusProvider`
  - **`TaskManager`**: owns the priority queue, asyncio semaphore, and full task-state machinery; delegates execution to an `ExecutionBackend`; accepts `idle_sleep_seconds` to tune the resolve-loop yield when idle (default 0 s for Symphony, 0.1 s for AWS Batch, matching each backend's original behaviour)
  - **`WorkerProcess`**: multiprocessing.Process lifecycle — connector setup, message dispatch, signal handling, graceful shutdown; constructed with a `ProcessorStatusProvider` and `ExecutionBackend`
  - **`WorkerManagerRunner`**: async event-loop boilerplate — heartbeat, command handling, signal handling, cleanup; constructed with a `WorkerProvisioner`; accepts `heartbeat_concurrency_multiplier` for backends (e.g. ECS) that scale the advertised concurrency by a per-task CPU factor
- Each backend now only implements the small ABC(s) relevant to it rather than subclassing shared state machinery; `ORBAWSEC2WorkerAdapter` composes a `WorkerManagerRunner` and drives it via `run_in_loop()` inside its ORB context manager
- Net reduction of ~680 lines across 18 source files
- Adapts all shared classes and backends to the new networking API from #724 (`AddressConfig`, `NetworkBackend`, `backend.create_async_connector`, `generate_identity_from_name`, etc.)
- Updates four test files and `examples/task_capabilities.py` to access worker manager config via `manager.config` (public property on `NativeWorkerManager`) instead of the private attributes removed by the refactor; also fixes `task_capabilities.py` to use `multiprocessing.get_context("spawn")` (unsafe fork with ZMQ/asyncio, same fix as #705)
- Fixes inverted `taskLock` in `HeartbeatManager` (was sending `True` when the worker could accept tasks, i.e. when unlocked)
- Fixes `AWSBatchHeartbeatManager` sending `Resource.cpu` values 10× too small — the field encodes CPU as a fixed-point integer (99.2% → 992 per the schema comment), but the old code omitted the `* 10` scaling that `SymphonyHeartbeatManager` already applied correctly; the new shared `HeartbeatManager` is consistent
- Fixes `WorkerProcess._cleanup` to own both connector destroys rather than splitting responsibility with `__get_loops`; previously `_connector_external.destroy()` was called at the end of `__get_loops` and again in `_cleanup`, causing a double-destroy on the normal exit path and a use-after-destroy race on the YMQ graceful-shutdown future
- Forwards `config.worker_config.io_threads` to `WorkerManagerRunner` in `NativeWorkerManager`, `SymphonyWorkerManager`, `ECSWorkerManager`, and `ORBAWSEC2WorkerAdapter`, restoring the pre-refactor behaviour (previously each manager passed its configured io_threads to the network backend)
- Adds 51 unit tests across three files: 30 for `TaskManager` in `test_task_manager.py` (init validation including `idle_sleep_seconds`, queuing, priority bypass, all cancel paths, `resolve_tasks`, `on_task_result`, and sentinel behaviour of `_load_task_inputs` before/after `register()`), 3 for `HeartbeatManager` in `test_heartbeat_manager.py` (verifying `taskLock` value vs semaphore state), and 16 for the shared runner/process/pool classes in `test_worker_manager_core.py` (`WorkerManagerRunner._handle_command` command routing, `WorkerProcess.__on_receive_external` backoff queue and message dispatch, `NativeWorkerPool` capacity and shutdown logic)
- `ExecutionBackend._load_task_inputs` now initialises to a sentinel coroutine that raises `RuntimeError` with a clear message if `execute()` is called before `register()`, rather than crashing with an opaque `TypeError`
- `AWSProcessorStatusProvider.get_processor_statuses` and `ORBWorkerPool.start_worker` now assert their preconditions (task manager and template ID set, respectively) instead of silently returning empty/wrong results
- Collapses identical ZMQ/YMQ signal handler branches in `WorkerManagerRunner._register_signal` (both did the same thing — no graceful-shutdown distinction needed at the manager level)
- Updates `docs/source/tutorials/worker_managers/aws_hpc_batch.rst` architecture diagram and components table to reflect the refactored class names (`WorkerProcess`, `AWSBatchExecutionBackend`) replacing the old `AWSBatchWorker` and `AWSHPCTaskManager`
- Documents `TaskManager.on_task_result` as satisfying the `TaskManagerMixin` interface but not dispatched by `WorkerProcess` (result handling goes via `resolve_tasks`); notes the known pre-existing `_queued_task_ids` accumulation for completed tasks

## Related Issue

Closes #722


src/scaler/worker_manager_adapter/aws_hpc/heartbeat_manager.py
```diff
@@ -1,116 +1,24 @@
...
-            latencyUS=self._latency_us,
-            taskLock=task_lock,
-            processors=[processor_status],
-            capabilities=self._capabilities,
-            workerManagerID=self._worker_manager_id,
-        )
-
-        await self._connector_external.send(heartbeat)
+    def get_processor_statuses(self) -> List[ProcessorStatus]:
+        assert self._task_manager is not None, "set_task_manager() must be called before get_processor_statuses()"
```
magniloquency, line 18: This assert replaces the original `if self._task_manager is None: return []` guard. The guard was unreachable:

- `get_processor_statuses` is only called from `HeartbeatManager.routine()`
- `routine()` is only scheduled after `WorkerProcess.__initialize()` completes
- `__initialize()` calls `heartbeat_manager.register()`, which calls `set_task_manager()`

So `_task_manager` is always set by the time this runs. The assert keeps the invariant explicit rather than silently masking a broken init order.

examples/task_capabilities.py
```diff
@@ -27,36 +27,36 @@ def main():
...
-                per_worker_task_queue_size=base_manager._task_queue_size,
-                heartbeat_interval_seconds=base_manager._heartbeat_interval_seconds,
-                task_timeout_seconds=base_manager._task_timeout_seconds,
-                death_timeout_seconds=base_manager._death_timeout_seconds,
-                garbage_collect_interval_seconds=base_manager._garbage_collect_interval_seconds,
-                trim_memory_threshold_bytes=base_manager._trim_memory_threshold_bytes,
-                hard_processor_suspend=base_manager._hard_processor_suspend,
-                io_threads=base_manager._io_threads,
-                event_loop=base_manager._event_loop,
+                per_worker_task_queue_size=base_config.worker_config.per_worker_task_queue_size,
```
gxuu: The original config was more flat, now it's nested for variables like this: `cluster._worker_manager.config.worker_config.hearbeat_interval_seconds`

any reason to make this change?

src/scaler/worker_manager_adapter/aws_hpc/heartbeat_manager.py
```diff
@@ -1,116 +1,24 @@
...
+from typing import TYPE_CHECKING, List, Optional
 
-Sends periodic heartbeats to the scheduler with worker status.
-"""
+from scaler.protocol.capnp import ProcessorStatus, Resource
+from scaler.worker_manager_adapter.mixins import ProcessorStatusProvider
 
-import time
-from typing import Any, Dict, Optional
+if TYPE_CHECKING:
```
gxuu, line 6: Existing code uses `from __future__ import annotations`?

src/scaler/worker_manager_adapter/aws_hpc/heartbeat_manager.py
```diff
@@ -1,116 +1,24 @@
...
-            workerManagerID=self._worker_manager_id,
-        )
-
-        await self._connector_external.send(heartbeat)
+    def get_processor_statuses(self) -> List[ProcessorStatus]:
+        assert self._task_manager is not None, "set_task_manager() must be called before get_processor_statuses()"
+        processing_tasks = self._task_manager.processing_task_count
+        return [
+            ProcessorStatus(
+                pid=0, initialized=True, hasTask=processing_tasks > 0, suspended=False, resource=Resource(cpu=0, rss=0)
```
gxuu, line 22: This functions returns dummy `ProcessorStatus`? 

Where will it be replaced by actual values?

src/scaler/worker_manager_adapter/aws_hpc/task_manager.py
```diff
@@ -359,7 +105,6 @@ async def _flush_pending_batch(self) -> None:
         await self._flush_batch(batch, futures_map)
 
     async def _flush_batch(self, batch: List[Tuple[Task, Any, List[Any]]], futures_map: Dict[TaskID, Future]) -> None:
-        """Submit a batch of tasks as either an array job or a single job."""
```
gxuu, line 362: Why are these docstirngs removed?

src/scaler/worker_manager_adapter/aws_hpc/task_manager.py
```diff
@@ -616,10 +349,9 @@ async def _submit_single_batch_job(self, task: Task, function: Any, arguments: L
         return response["jobId"]
 
     async def _monitor_batch_job(self, job_id: str, future: Future, task_id: TaskID) -> None:
-        """Monitor AWS Batch job and resolve future when complete."""
         import gzip
 
-        poll_interval = 2.0  # seconds
+        poll_interval = 2.0
```
gxuu: Perhaps we should use `poll_interval_seconds` instead.

src/scaler/worker_manager_adapter/aws_hpc/task_manager.py
gxuu: Some comments should be removed but docstrings are worth keeping.

src/scaler/worker_manager_adapter/aws_hpc/worker.py
```diff
@@ -1,261 +1,59 @@
...
-            pass
-
-    def __destroy(self) -> None:
-        self._task.cancel()
+from scaler.worker_manager_adapter.aws_hpc.heartbeat_manager import AWSProcessorStatusProvider
+from scaler.worker_manager_adapter.aws_hpc.task_manager import AWSBatchExecutionBackend
+from scaler.worker_manager_adapter.worker_process import WorkerProcess
+
+
+def create_aws_batch_worker(
```
gxuu, line 17: Is the common arguments similar enough so we should have `create_xxx_worker`?
(I see we have `create_symphony_worker` below).

tests/worker_manager_adapter/test_worker_manager_core.py
```diff
@@ -0,0 +1,309 @@
...
+from scaler.worker_manager_adapter.baremetal.native import NativeWorkerPool
+from scaler.worker_manager_adapter.mixins import WorkerPool
+from scaler.worker_manager_adapter.worker_manager_runner import WorkerManagerRunner
+from scaler.worker_manager_adapter.worker_process import WorkerProcess
+from tests.utility.utility import logging_test_name
+
+Status = WorkerManagerCommandResponse.Status
+
+
+# ─── helpers ─────────────────────────────────────────────────────────────────
```
gxuu: remove this line

also, do we put helpers at the bottom of the file or in current place? if the first is true, please move it to the bottom

src/scaler/worker_manager_adapter/mixins.py
```diff
@@ -0,0 +1,50 @@
...
+from scaler.utility.identifiers import TaskID
+
+if TYPE_CHECKING:
+    from scaler.worker_manager_adapter.task_manager import TaskManager
+
+Status = WorkerManagerCommandResponse.Status
+
+
+class ProcessorStatusProvider(ABC):
+    def set_task_manager(self, task_manager: "TaskManager") -> None:
```
sharpener6: what is this? why abstract class have a method is not abstract?

src/scaler/worker_manager_adapter/mixins.py
```diff
@@ -0,0 +1,50 @@
...
+        pass
+
+    @abstractmethod
+    def get_processor_statuses(self) -> List[ProcessorStatus]: ...
+
+
+class ExecutionBackend(ABC):
+    def __init__(self) -> None:
+        async def _not_registered(task: Task) -> Tuple[Any, List[Any]]:
```
sharpener6: why abstract class has implmentation

src/scaler/worker_manager_adapter/mixins.py
```diff
@@ -0,0 +1,50 @@
...
+        pass
+
+    def on_cleanup(self, task_id: TaskID) -> None:
+        pass
+
+    async def routine(self) -> None:
+        pass
+
+
+class WorkerPool(ABC):
```
sharpener6: WorkerProvisioner

src/scaler/worker_manager_adapter/mixins.py
```diff
@@ -0,0 +1,50 @@
...
+
+        self._load_task_inputs: Callable[[Task], Awaitable[Tuple[Any, List[Any]]]] = _not_registered
+
+    def register(self, load_task_inputs: Callable[[Task], Awaitable[Tuple[Any, List[Any]]]]) -> None:
+        self._load_task_inputs = load_task_inputs
+
+    @abstractmethod
+    async def execute(self, task: Task) -> asyncio.Future: ...
+
+    async def on_cancel(self, task_cancel: TaskCancel) -> None:
```
sharpener6: what are those implementation for?

examples/task_capabilities.py
magniloquency: The old monolithic worker manager config was split into multiple config classes. I've added a local variable in the example and some similar files to clean it up a bit.

src/scaler/worker_manager_adapter/aws_hpc/heartbeat_manager.py
magniloquency: Other production files use the string-quote approach for forward references (`processor.py`, `worker_manager_adapter/heartbeat_manager.py`), so keeping it consistent with those.

src/scaler/worker_manager_adapter/aws_hpc/heartbeat_manager.py
magniloquency: hasTask is a real value, but pid, cpu, and res are intentionally set to 0 because AWS Batch doesn't give us per-process metrics. This is existing behaviour copied over from before the refactor.

src/scaler/worker_manager_adapter/aws_hpc/task_manager.py
magniloquency: Good catch — restored docstrings on the non-trivial methods (`routine`, `_run_in_executor`, `_submit_array_job`, `_monitor_array_job`, `_monitor_batch_job`) in the latest commit. Skipped the trivial one-liners that just restate the method name.

src/scaler/worker_manager_adapter/aws_hpc/task_manager.py
magniloquency: Done, renamed in both `_monitor_batch_job` and `_monitor_array_job`.

src/scaler/worker_manager_adapter/aws_hpc/worker.py
magniloquency: The two functions share the common `WorkerProcess` parameters, but `create_aws_batch_worker` has several AWS-specific extras (`job_queue`, `job_definition`, `aws_region`, `s3_bucket`, `s3_prefix`, `job_timeout_seconds`, `name`) with no Symphony equivalent. Extracting a shared base would add indirection without meaningful reuse, so I'd prefer to keep them separate for now.

tests/worker_manager_adapter/test_worker_manager_core.py
magniloquency: Done — removed all separator comments and moved the helpers to the bottom of the file.

src/scaler/worker_manager_adapter/mixins.py
magniloquency: `set_task_manager` is an optional hook with a no-op default — not every `ProcessorStatusProvider` needs a `TaskManager`. For example, a provider that returns fixed or system-level metrics wouldn't need one. Making it `@abstractmethod` would force all subclasses to implement something they may not need. The same pattern is used for `on_cancel`, `on_cleanup`, and `routine` in `ExecutionBackend`.

src/scaler/worker_manager_adapter/mixins.py
magniloquency: Done — renamed to `WorkerProvisioner` across all implementations (`NativeWorkerPool`, `SymphonyWorkerPool`, `ECSWorkerPool`, `ORBWorkerPool`) and usages.

magniloquency (commented):
Good catch — replaced the `__init__` with a class-level type annotation. `register()` still sets the attribute at runtime, and `task_manager.py` always calls it before any `execute()` is possible, so the sentinel was unnecessary defensive code. Fixed in 091aad89.

src/scaler/worker_manager_adapter/mixins.py
magniloquency: Good catch — replaced the `__init__` with a class-level type annotation. `register()` still sets the attribute at runtime, and `task_manager.py` always calls it before any `execute()` is possible, so the sentinel was unnecessary defensive code. Fixed in 091aad89.

src/scaler/worker_manager_adapter/mixins.py
magniloquency: These are optional lifecycle hooks with no-op defaults. Not every `ExecutionBackend` needs cancellation handling, cleanup logic, or a background routine — making them `@abstractmethod` would force all subclasses to implement something they may not need. The same pattern is used for `set_task_manager` in `ProcessorStatusProvider`.

src/scaler/worker_manager_adapter/aws_hpc/heartbeat_manager.py
gxuu: I am a little worry about `pid=0` in particular. Would it make sense to set `pid` to be different for different AWS Batch? Just asking a question here, not saying that this should be done. As a general rule different processor should have something like an id in their status.

src/scaler/worker_manager_adapter/mixins.py
sharpener6: Check coding guideline, abstract class should not have any implementation

src/scaler/worker_manager_adapter/mixins.py
sharpener6: if the provider doesn't need a task manager, then they need implement such method to explicitly "do nothing"

src/scaler/worker_manager_adapter/mixins.py
magniloquency: I couldn't find anything about this in the contributing guidelines or agents.md, so i added it here: [5ce20e7](https://github.com/finos/opengris-scaler/pull/730/commits/5ce20e7842a9fbb4878ea58bf7f5f27e588b5e3c)

magniloquency: Done — extracted `register`/`_load_task_inputs` into a concrete `TaskInputLoaderMixin`, and made `on_cancel`, `on_cleanup`, `routine`, and `set_task_manager` all `@abstractmethod`. Subclasses that don't need a hook now implement an explicit `pass` body. `ExecutionBackend` and `ProcessorStatusProvider` are now purely abstract.

src/scaler/worker_manager_adapter/mixins.py
magniloquency: Done — `set_task_manager` is now `@abstractmethod` in `ProcessorStatusProvider`. `SymphonyProcessorStatusProvider`, which doesn't need a task manager, now implements it explicitly with a `pass` body.

src/scaler/worker_manager_adapter/mixins.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @abstractmethod
+    def get_processor_statuses(self) -> List[ProcessorStatus]: ...
+
+
+class TaskInputLoaderMixin:
+    async def _load_task_inputs(self, task: Task) -> Tuple[Any, List[Any]]:
+        raise RuntimeError(f"{type(self).__name__}.register() must be called before execute()")
+
+    def register(self, load_task_inputs: Callable[[Task], Awaitable[Tuple[Any, List[Any]]]]) -> None:
```
sharpener6: this should also be abstractmethod

if you have something is commonly used, it can be refactored as normal function, or duplicate 1 line of code is fine

src/scaler/worker_manager_adapter/baremetal/native.py
```diff
@@ -96,135 +67,76 @@ def _create_worker(self) -> Worker:
...
-            WorkerManagerCommandResponse(
-                command=cmd_type, status=response_status, workerIDs=worker_ids, capabilities=self._capabilities
-            )
-        )
+        for worker in self._workers.values():
+            worker.join()
 
-    async def start_worker(self) -> Tuple[WorkerID, Status]:
+    async def start_worker(self) -> Tuple[List[bytes], Status]:
         if len(self._workers) >= self._max_task_concurrency != -1:
```
gxuu: I think writing it as `self._max_task_concurrency != 1 && len(self._workers) >= self._max_task_concurrency` is better:
- easier to understand
- Short circuit on faster-to-judge condition

This should apply to other places that used chain comparison in the form of `a >= b != c` as well.

src/scaler/worker_manager_adapter/mixins.py
magniloquency: Done — made `TaskInputLoaderMixin` an `ABC` and both `register` and `_load_task_inputs` are now `@abstractmethod`. Each concrete backend (`AWSBatchExecutionBackend`, `SymphonyExecutionBackend`) stores the callback under a `_loader` annotation and implements both methods explicitly.

src/scaler/worker_manager_adapter/baremetal/native.py
magniloquency: while this is technically out of scope for this pr (it's unrelated existing code) i completely agree and have snuck it into here. done.

on a related note claude has also identified a few gaps in the hpc wm. i addressed a few but im going to leave the rest (they're mostly minor) because the pr is getting too far off-track

src/scaler/worker_manager_adapter/symphony/worker_manager.py
```diff
@@ -1,172 +1,89 @@
...
+from scaler.worker_manager_adapter.worker_manager_runner import WorkerManagerRunner
+from scaler.worker_manager_adapter.worker_process import WorkerProcess
 
 Status = WorkerManagerCommandResponse.Status
 
 
-class SymphonyWorkerManager:
-    def __init__(self, config: SymphonyWorkerManagerConfig):
-        self._address = config.worker_manager_config.scheduler_address
+class SymphonyWorkerPool(WorkerProvisioner):
```
sharpener6: please change all the terms, not just base class term, I know you are using AI, but you need review it and be very picky on AI generated code, especially the names, any new terminology will misleading reviewer,

If you are using AI to write code, and not review it, then it become AI slop.

You have way more responsibility to review instead of passing review burden all the way to final reviewer

magniloquency: done

e117649:
Should we also rename the folder to worker_managers since we no longer use the term adapter?

src/scaler/worker_manager_adapter/aws_hpc/heartbeat_manager.py
sharpener6: @gxuu for Batch, this is fake one, batch worker manager use batch client internally, like scaler client, which doesn't really know actual processor status

sharpener6 (dismissed):

src/scaler/worker_manager_adapter/aws_hpc/worker.py
```diff
@@ -1,261 +1,59 @@
 """
-AWS HPC Worker.
+AWS Batch Worker factory.
```
sharpener6: AWS HPC Batch Worker factory

src/scaler/worker_manager_adapter/mixins.py
```diff
@@ -0,0 +1,52 @@
...
+
+class ProcessorStatusProvider(ABC):
+    @abstractmethod
+    def set_task_manager(self, task_manager: "TaskManager") -> None: ...
+
+    @abstractmethod
+    def get_processor_statuses(self) -> List[ProcessorStatus]: ...
+
+
+class TaskInputLoaderMixin(ABC):
```
sharpener6: rename to `TaskInputLoader` instead of `TaskInputLoaderMixin`

src/scaler/worker_manager_adapter/mixins.py
```diff
@@ -0,0 +1,52 @@
...
+    @abstractmethod
+    async def _load_task_inputs(self, task: Task) -> Tuple[Any, List[Any]]: ...
```
sharpener6: rename to `load_task_inputs`

src/scaler/worker_manager_adapter/aws_hpc/worker.py
magniloquency: done

src/scaler/worker_manager_adapter/mixins.py
magniloquency: done

src/scaler/worker_manager_adapter/mixins.py
magniloquency: done

magniloquency:
@e117649 

> Should we also rename the folder to worker_managers since we no longer use the term adapter?

I think we should, but not in this PR

sharpener6 (approved):

PR #733, opened 2026-04-17
fix: derive worker IDs deterministically from worker_manager_id

magniloquency:
## Summary

- Added an optional `unique_worker_tag` parameter to `WorkerID.generate_worker_id()` (as suggested by @rafa-be); when provided it is used as the unique suffix instead of a random UUID, preserving existing random behaviour when omitted
- Removed the now-redundant `deterministic_worker_ids` flag from `Worker.__init__`, which was unreachable dead code
- In the native worker manager, when `worker_manager_id` is set, each worker now receives a deterministic ID of the form `Worker|{prefix}|{manager_id}|{index}` rather than a random one
- In the ORB AWS EC2 adapter, replaced `get_orb_aws_ec2_worker_name` with `WorkerID.generate_worker_id("ORB", unique_worker_tag=f"{instance_id}|0")` so the pre-computed ID matches what the native worker manager will generate on the EC2 instance

## Related Issue

Closes #700


magniloquency:
see #700

PR #738, opened 2026-04-21
worker manager controller emits new command

gxuu:
# READY FOR REVIEW.

- After this PR gets merged, the scheduler will emits old and new commands at the same time. Worker managers that migrates to the new command will have to drop old command at the same time. Otherwise, there will be a duplicate-start problem.
- start worker / shutdown workers command support will soon be dropped.
- worker managers should take `setWrokerCount` as the ONLY command from scheduler.
    - current approach is to simplify message exchange, based on the fact that scheduler reroute tasks from dead workers.
    - this is slower than the old approach `shutdownWorkers`, but cleaner on structure
    - old ways still work, so it needs your decision.


src/protocol/message.capnp
```diff
@@ -101,19 +101,27 @@ struct WorkerManagerHeartbeatEcho {
 enum WorkerManagerCommandType {
     startWorkers @0;
     shutdownWorkers @1;
+    setWorkerCount @2;
```
gxuu: `setWorkerCount` gives `[{numWorkers1:[cap1, cap2]}, {numWorkers2:[cap3]}]`.

This works for both `startWorkers` and `shutdownWorkers`. The worker managers are responsible for adjusting worker count based on this command.

src/scaler/protocol/helpers.py
sharpener6, line 30: can we also add `List[TaskCapability]` as annotation here?

gxuu: The helpers accept both a dictionary of `str:int` and a list of `TaskCapabilities`. I have made the annotation as a `Union` of those.

src/scaler/protocol/helpers.py
```diff
@@ -27,13 +28,17 @@ def from_capnp_object_id(capnp_object_id: CapnpObjectID) -> ScalerObjectID:
     )
 
 
-def capabilities_to_dict(capabilities) -> Dict[str, int]:
+def capabilities_to_dict(capabilities: Union[Dict[str, int], List[TaskCapability]]) -> Dict[str, int]:
```
sharpener6: if it's `Dict[str, int]`, then it should not calling this helper function at all

gxuu: this function is removed entirely.

sharpener6 (approved):

PR #739, opened 2026-04-22
feat: add WebSocket transport to YMQ

### LINKED ISSUE #728: Add WebSocket transport support to YMQ
> **Note:** This is an exploratory issue — it is not currently planned or scheduled. It documents a potential direction for discussion purposes only.

## Summary

Add WebSocket (`ws://` / `wss://`) as a first-class transport in YMQ alongside the existing TCP and IPC transports.

This would allow YMQ-based clients to connect to schedulers and workers through WebSocket-aware infrastructure (load balancers, proxies, firewalls) without requiring raw TCP access.

## Motivation

Raw TCP is not always viable in cloud/browser environments:
- Firewalls and corporate proxies often block arbitrary TCP ports but allow port 80/443
- Browser-based monitoring or debugging tools cannot use raw TCP sockets
- Cloud load balancers (ALB, GCP HTTPS LB) terminate WebSocket connections natively, enabling TLS offload and routing

## Architecture

YMQ's layered design makes this tractable without touching the protocol core. `MessageConnection` (identity exchange + 8-byte frame header) sits *above* the transport layer and is already transport-agnostic.

**Changes required:**

| Component | Change |
|---|---|
| `internal/client.h` | Add `WebSocketClient` as third variant (alongside `TCPSocket`, `Pipe`) — implement `write`, `readStart`, `readStop`, `shutdown`, `closeReset` |
| `internal/accept_server.h` | Add `WebSocketServer` variant — HTTP upgrade handshake, then yield `Client(WebSocketClient)` |
| `internal/connect_client.h` | Add `WebSocketConnectClient` variant — issue HTTP GET upgrade, handle `101 Switching Protocols`, then hand off |
| `address.h` | Parse `ws://host:port/path` and `wss://` in `Address::fromString()` |
| `CMakeLists.txt` | Find and link WebSocket library |

`BinderSocket`, `ConnectorSocket`, `IOContext`, `EventLoopThread`, and `MessageConnection` require **no changes**.

**Wire protocol:** the YMQ magic bytes `[Y, M, Q, 1]` and message framing are sent *after* the WebSocket handshake completes — no conflict.

## Recommended library

[`libwebsockets`](https://libwebsockets.org/) — integrates natively with libuv via `lws_set_loop_ops()`, so it plugs directly into `EventLoopThread`. Alternatives: `boost::beast` (header-only, no new dep) or `uWebSockets`.

## Scope

- [ ] `ws://` (plain WebSocket over TCP)
- [ ] Address parsing and `BinderSocket`/`ConnectorSocket` wiring
- [ ] Unit/integration tests mirroring the existing TCP test suite
- [ ] `wss://` (TLS) — separate follow-up; requires OpenSSL or mbedTLS

magniloquency:
## Summary

Implements WebSocket (`ws://`) transport support end-to-end, resolving #728. `wss://` (TLS) is parsed and accepted as a valid address scheme but is not yet functional.

**YMQ C++ transport layer**
- **Address parsing**: add `WebSocketAddress` struct and `Type::WebSocket` enum value to `Address`; `ws://` and `wss://` URLs are parsed and round-trip correctly alongside existing `tcp://` and `ipc://` addresses
- **`WebSocketStream`**: new class implementing the RFC 6455 HTTP/1.1 Upgrade handshake (client and server sides), binary frame framing/deframing, and the SHA-1 + Base64 computation for `Sec-WebSocket-Accept` — no external library required for `ws://`
- **`Client` variant**: extended to include `WebSocketStream` alongside `TCPStream` and `IPCStream`, so upper layers remain transport-agnostic
- **`AcceptServer`**: WebSocket accept path wired in — binds a TCP listener, upgrades accepted connections via `WebSocketStream` before delivering a ready `Client` to the caller's callback
- **`ConnectClient`**: WebSocket connect path wired in — TCP connect followed by HTTP Upgrade, with the same retry/disconnect support as TCP

**Python configuration layer**
- **`AddressConfig`**: extended to parse and round-trip `ws://` and `wss://` addresses; `from_string` now handles all four schemes (`tcp://`, `ipc://`, `ws://`, `wss://`) in a unified branch

**Object storage server**
- **C++ object storage server**: accepts a `ws://` bind address in addition to `tcp://`; the active scheme is propagated to Python via the Cap'n Proto `ObjectStorageAddress` message so callers always receive the correct URL
- **Python cluster integration**: `ObjectStorageServer` reads the scheme from the capnp message and constructs the correct `ws://` or `tcp://` address when reporting the server endpoint

**Tests**
- Address round-trip unit tests and an end-to-end `ClientServerHandshake` test over `ws://` in `tests/cpp/ymq/`
- `AddressConfig` round-trip tests for `ws://` and `wss://` in `tests/config/test_config_types.py`
- **`tests/io/uv_ymq/test_websocket_interop.py`**: two new Python tests that validate RFC 6455 compliance by connecting a standard `websockets` client to a YMQ `BinderSocket` (and vice versa), speaking the YMQ wire protocol (magic bytes + length-prefixed frames) directly over WebSocket frames

**New API**
- **`ConnectorSocket.async_connect()`** added to `src/scaler/io/ymq/sockets.py`: async variant of `connect()` that uses `call_async` so the asyncio event loop is not blocked during the YMQ handshake

**Dependencies**
- `websockets>=12.0` added to the `dev` dependency group in `pyproject.toml` and installed explicitly in the `run-test` CI action

**Bug fix**
- **`ConnectorSocket::recvMessage()`**: reordered `_pendingRecvMessages` check before `_disconnected` check, so messages buffered before a graceful remote close are delivered instead of silently discarded with `ConnectorSocketClosedByRemoteEnd`

## Related Issue

Closes #728


scripts/library_tool.sh
```diff
@@ -2,7 +2,7 @@
 # This script builds and installs the required 3rd party C++ libraries.
 #
 # Usage:
-#    	./scripts/library_tool.sh [boost|capnp|uv] [compile|install] [--prefix=PREFIX]
+#    	./scripts/library_tool.sh [boost|capnp|libuv|libwebsockets] [compile|install] [--prefix=PREFIX]
```
magniloquency: Changed `uv` to `libuv` so that the help text is accurate

magniloquency (commented):
## `websocket_utils` — why hand-rolled, and what the alternatives are

The SHA-1 and Base64 implementations in `websocket_utils.h/.cpp` are only used for the RFC 6455 opening handshake (`Sec-WebSocket-Key` / `Sec-WebSocket-Accept`). They are not on the data path.

### Security

Hand-rolling these presents no security concern for this use case:

- **SHA-1**: SHA-1 is collision-broken, but any implementation — including OpenSSL's — would be equally so, because RFC 6455 §4.1 mandates SHA-1 for the accept header computation. The question of implementation quality is separate from the protocol's choice of hash. In practice the known attacks are collision-finding attacks; what matters here is a specific key derivation, which is unaffected. SHA-1 collision weaknesses cannot be exploited against this construction.
- **`std::mt19937` for key generation**: `mt19937` is not a CSPRNG, but the WebSocket handshake key is not a security primitive. RFC 6455 §4.1 explicitly states that the nonce exists only to prevent non-WebSocket servers from accidentally accepting WebSocket upgrades (cross-protocol cache poisoning by proxies), not to provide confidentiality or authentication. Predicting the key gains an attacker nothing useful.
- **`base64Encode`**: Encoding only — no keys or secrets involved.
- **`extractHeader`**: Not constant-time, but timing attacks on WebSocket handshake header matching are not a meaningful threat.

Real security for WebSocket connections comes from TLS (`wss://`), which this transport does not implement. If `wss://` support is added in the future, a TLS library (e.g. OpenSSL, mbedTLS) would be introduced, and it would provide SHA-1 and Base64 as well — making this file redundant at that point.

### Alternatives considered

**OpenSSL**
Provides `SHA1()` and `EVP_EncodeBlock()`. Not currently a project dependency — pulling it in solely for two handshake-only helpers would be heavyweight.

**Cap'n Proto KJ** (`kj::encodeBase64`)
Already a project dependency and provides Base64 encoding. However, the `CapnProto::kj` CMake target is not linked into `ymq` (only into `object_storage` and `protocol`). Adding it would couple a low-level network layer to a serialisation library, which feels wrong.

**A single-header SHA-1 library**
Would add a new vendored dependency of comparable size to this file with no net reduction in code.

Still open to any of these, or other approaches — if there's a preferred solution let me know.

magniloquency:
## Potential simplification: wslay for WebSocket framing

[wslay](https://tatsuhiro-t.github.io/wslay/) is a C library that handles WebSocket frame encoding and decoding with no I/O or event loop of its own — you feed it bytes and it gives you frames. It could replace the hand-rolled framing code in `websocket_stream.cpp` (frame header construction, masking, length encoding, fragmentation reassembly, PING/PONG/CLOSE handling).

Unlike libwebsockets, wslay doesn't take ownership of the connection or the event loop, so it would fit naturally on top of the existing libuv `TCPSocket` without any architectural changes.

Not proposing to do this now, but worth keeping in mind if the framing code needs to grow (e.g. permessage-deflate compression).

src/cpp/scaler/ymq/connector_socket.cpp
rafa-be, line 1: That might actually solve the unit tests failing from time to time.

src/scaler/client/agent/heartbeat_manager.py
```diff
@@ -53,8 +53,14 @@ async def on_heartbeat_echo(self, heartbeat: ClientHeartbeatEcho):
             return
 
         object_storage_address_message = heartbeat.objectStorageAddress
+        # scheme is empty when talking to a scheduler that predates this field; fall back to tcp
+        scheme = (
+            SocketType(object_storage_address_message.scheme)
+            if object_storage_address_message.scheme
+            else SocketType.tcp
+        )
```
rafa-be: I'm not sure we want to be backward compatible.

@sharpener6 ?

src/cpp/scaler/ymq/internal/websocket_stream.h
```diff
@@ -0,0 +1,85 @@
...
+    WebSocketStream(const WebSocketStream&)            = delete;
+    WebSocketStream& operator=(const WebSocketStream&) = delete;
+
+    WebSocketStream(WebSocketStream&&) noexcept            = default;
+    WebSocketStream& operator=(WebSocketStream&&) noexcept = default;
+
+    // Create a ready-to-use WebSocketStream from an already-upgraded TCP socket.
+    // Used internally by the upgrade helpers; not part of the public transport API.
+    static WebSocketStream fromUpgradedSocket(
+        scaler::wrapper::uv::TCPSocket socket, bool isServer, std::vector<uint8_t> leftover = {}) noexcept;
```
rafa-be: Make that one private

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -0,0 +1,671 @@
...
+    if (!requestLine.starts_with("GET ") || !requestLine.ends_with(" HTTP/1.1")) {
+        ctx->callback(std::unexpected(scaler::wrapper::uv::Error {UV_EPROTO}));
+        return;
+    }
+
+    const auto upgradeHeader    = extractHeader(headers, "Upgrade");
+    const auto keyHeader        = extractHeader(headers, "Sec-WebSocket-Key");
+    const auto connectionHeader = extractHeader(headers, "Connection");
+    const auto versionHeader    = extractHeader(headers, "Sec-WebSocket-Version");
+
```
rafa-be: Unless I'm wrong, `extractHeader()` is O(n). Calling it every time for each header string is somewhat computation heavy.

Would be better to have a `extractHeaders()` function that builds and returns a `map<string, string>`, only parsing the headers once.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -0,0 +1,671 @@
...
+    if (!upgradeHeader.has_value() || !keyHeader.has_value() || !connectionHeader.has_value() ||
+        !versionHeader.has_value()) {
+        ctx->callback(std::unexpected(scaler::wrapper::uv::Error {UV_EPROTO}));
+        return;
+    }
+
+    // Verify Upgrade: websocket (case-insensitive)
+    std::string upgradeValue = *upgradeHeader;
+    for (char& c: upgradeValue)
+        c = static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
```
rafa-be: They are several places where a string's case is manually lowered. Would be better to have a `stringToLower()` function instead.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -0,0 +1,671 @@
...
+    scaler::utility::MoveOnlyFunction<void(std::expected<WebSocketStream, scaler::wrapper::uv::Error>)>
+        callback) noexcept
+{
+    auto ctx      = std::make_shared<ClientUpgradeContext>();
+    ctx->socket   = std::move(socket);
+    ctx->key      = generateWebSocketKey();
+    ctx->callback = std::move(callback);
+
+    const std::string request = buildClientUpgradeRequest(address.host, address.port, address.path, ctx->key);
+    auto requestData          = std::make_shared<std::string>(request);
```
rafa-be:
```suggestion
    auto requestData          = std::make_shared<std::string>(buildClientUpgradeRequest(address.host, address.port, address.path, ctx->key));
```

This would avoid 1 unnecessary copy, I think.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -0,0 +1,671 @@
...
+                        ctx->socket->readStop();
+                        ctx->callback(std::unexpected(scaler::wrapper::uv::Error {UV_EPROTO}));
+                        return;
+                    }
+
+                    const std::string_view view(
+                        reinterpret_cast<const char*>(ctx->recvBuffer.data()), ctx->recvBuffer.size());
+                    if (view.find("\r\n\r\n") != std::string_view::npos) {
+                        finishClientUpgrade(std::move(ctx));
+                    }
```
rafa-be, line 409: You'll have to locally copy the `ctx` value because `readStop()` will destroy the lambda and the captured `ctx`, this invalidating the `callback`.

```suggestion
                   // Copy ctx to the stack before readStop() — readStop() destroys this lambda (and the
                   // captured ctx) via setData({}), so ctx must outlive that call.
                   auto safeCtx = ctx;

                    if (!readResult.has_value()) {
                        safeCtx->socket->readStop();
                        safeCtx->callback(std::unexpected(readResult.error()));
                        return;
                    }

                    const auto& data = readResult.value();
                    safeCtx->recvBuffer.insert(safeCtx->recvBuffer.end(), data.begin(), data.end());

                    if (safeCtx->recvBuffer.size() > kMaxUpgradeHeaderSize) {
                        safeCtx->socket->readStop();
                        safeCtx->callback(std::unexpected(scaler::wrapper::uv::Error {UV_EPROTO}));
                        return;
                    }

                    const std::string_view view(
                        reinterpret_cast<const char*>(safeCtx->recvBuffer.data()), safeCtx->recvBuffer.size());
                    if (view.find("\r\n\r\n") != std::string_view::npos) {
                        finishClientUpgrade(std::move(safeCtx));
                    }
```

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -0,0 +1,671 @@
...
+    ctx->key      = generateWebSocketKey();
+    ctx->callback = std::move(callback);
+
+    const std::string request = buildClientUpgradeRequest(address.host, address.port, address.path, ctx->key);
+    auto requestData          = std::make_shared<std::string>(request);
+    const std::span<const uint8_t> requestSpan(
+        reinterpret_cast<const uint8_t*>(requestData->data()), requestData->size());
+
+    auto writeResult = ctx->socket->write(
+        std::span<const std::span<const uint8_t>>(&requestSpan, 1),
```
rafa-be:
```suggestion
        requestSpan,
```

I think that should work as `TCPSocket` has a single buffer overload.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -0,0 +1,671 @@
...
+                    }
+                }
+                // Reserved opcodes are silently ignored.
+            }
+        });
+}
+
+void WebSocketStream::readStop() noexcept
+{
+    _state->_readActive = false;
```
rafa-be, line 609: I'm not sure this flag is required.

Libuv guarantee that the read callback will never be called again once readStop() is called once. And readStop() always succeed.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -0,0 +1,671 @@
...
+
+std::expected<void, scaler::wrapper::uv::Error> WebSocketStream::readStart(
+    scaler::wrapper::uv::ReadCallback callback) noexcept
+{
+    auto state           = _state;
+    state->_readCallback = std::move(callback);
+    state->_readActive   = true;
+
+    return state->_socket.readStart(
+        [state](std::expected<std::span<const uint8_t>, scaler::wrapper::uv::Error> result) mutable {
```
rafa-be, line 602: Maybe move this callback to a `static void onRead(state, result)` method as it becomes quite complex.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -0,0 +1,671 @@
...
+    auto closeFrame = buildControlFrame(kOpcodeClose, !_state->_isServer, {});
+    auto frameData  = std::make_shared<std::vector<uint8_t>>(std::move(closeFrame));
+    const std::span<const uint8_t> frameSpan(*frameData);
+    auto state = _state;
+
+    // Share ownership so both the synchronous error path and the async write callback can invoke it.
+    auto callbackPtr = std::make_shared<scaler::wrapper::uv::ShutdownCallback>(std::move(callback));
+
+    auto result = _state->_socket.write(
+        std::span<const std::span<const uint8_t>>(&frameSpan, 1),
```
rafa-be:
```suggestion
        frameSpan,
```

`TCPSocket::write()` has a single buffer overload.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -0,0 +1,671 @@
...
+
+#include "scaler/ymq/internal/websocket_utils.h"
+
+namespace scaler {
+namespace ymq {
+namespace internal {
+
+namespace {
+
+static constexpr uint64_t kMaxWebSocketPayloadSize = 64ULL * 1024 * 1024;  // 64 MB
```
rafa-be: Why this 64 MB limit?

As far as I understand, WebSockets should support up to `2^64 - 1` byte payloads.

64 MB is very small for some of Scaler's use cases.

tests/cpp/ymq/internal/test_websocket_stream.cpp
```diff
@@ -0,0 +1,540 @@
...
+{
+    const auto addr = scaler::ymq::Address::fromString("wss://127.0.0.1:443/");
+    ASSERT_TRUE(addr.has_value());
+    ASSERT_EQ(addr->type(), scaler::ymq::Address::Type::WebSocket);
+    ASSERT_TRUE(addr->asWebSocket().secure);
+
+    const auto str = addr->toString();
+    ASSERT_TRUE(str.has_value());
+    ASSERT_TRUE(str->starts_with("wss://"));
+}
```
rafa-be: These are not related to `WebsocketStream`, and are already kind of tested in `test_ymq.cpp`

rafa-be (dismissed):
LGTM except for a few minor improvements.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -0,0 +1,671 @@
...
+static constexpr uint8_t kMaskOpcode = 0x0F;
+static constexpr uint8_t kMaskLength = 0x7F;
+
+// WebSocket opcodes (RFC 6455 §5.2)
+static constexpr uint8_t kOpcodeContinuation = 0x0;
+static constexpr uint8_t kOpcodeText         = 0x1;
+static constexpr uint8_t kOpcodeBinary       = 0x2;
+static constexpr uint8_t kOpcodeClose        = 0x8;
+static constexpr uint8_t kOpcodePing         = 0x9;
+static constexpr uint8_t kOpcodePong         = 0xA;
```
rafa-be: We usually don't use `k`-prefixed constants.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
magniloquency: Done. Created a helper

src/cpp/scaler/ymq/internal/websocket_stream.cpp
magniloquency: Indeed, but the while loop `while (_readActive && !_recvBuffer.empty())` in `onRead` needs it. If the user calls `readStop()` from within their `readCallback`, `_readActive` goes false mid-loop and stops further frame delivery. Without it, buffered frames would still be delivered after `readStop()` — which might be acceptable, what do you think?

src/cpp/scaler/ymq/internal/websocket_stream.cpp
magniloquency: changed to upper snake case

tests/cpp/ymq/internal/test_websocket_stream.cpp
magniloquency: removed these tests

src/cpp/scaler/ymq/internal/websocket_stream.cpp
magniloquency: removed the limit

src/cpp/scaler/ymq/internal/websocket_stream.cpp
magniloquency: done

src/cpp/scaler/ymq/internal/websocket_stream.cpp
magniloquency: done

src/cpp/scaler/ymq/internal/websocket_stream.cpp
magniloquency: done

src/cpp/scaler/ymq/internal/websocket_stream.cpp
magniloquency: done

src/cpp/scaler/ymq/internal/websocket_stream.cpp
magniloquency: done

src/cpp/scaler/ymq/internal/websocket_stream.cpp
magniloquency: implemented `extractHeaders()` and removed `extractHeader()`

src/cpp/scaler/ymq/internal/websocket_stream.h
magniloquency: This function is used in the unit tests to bypass certain parts of the flow in unit tests. Could make it private and then `friend` it in the test class?

rafa-be: Oh I see.

I think interfaces should be as clean as possible, so if you can keep the user-facing header file clean, and move the boilerplate stuff to the test file, that would be awesome.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
rafa-be: Yeah I think you're right, it's safer to do it that way.

The upper YMQ layer (`MessageConnection`) does not expect new `onRead` callbacks after calling `readStop()`, so we keep the same libuv guarantee by having this flag.

rafa-be (dismissed):

src/scaler/client/agent/heartbeat_manager.py
sharpener6: no need, I am pretty sure it's vibe coded, llm loves back compatibility

magniloquency: I kept this in here just in case. With the ORB WM there can be differences in the versions being run by the client, scheduler + oss, and workers, because the workers can only install versions of the scaler that are on pypi. This might only be relevant for dev/test though, I assume our users will always install from pypi, rather than from source.

magniloquency: I've removed this

rafa-be (dismissed):

sharpener6 (approved):

PR #743, opened 2026-04-23
fix: treat --max-task-concurrency as total worker count in ORB AWS EC2 and ECS managers

magniloquency:
## Summary

- **ORB AWS EC2**: `--max-task-concurrency` was silently interpreted as the maximum number of EC2 instances, not the total number of workers. With an instance type with 4 vCPUs, setting `MTC=4` would launch 4 instances × 4 workers = 16 workers, but the scheduler heartbeat only reported capacity 4, causing severe under-utilization. This PR queries `DescribeInstanceTypes` to get the vCPU count for the configured `--instance-type`, then computes `max_instances = ceil(MTC / vcpus)` so that MTC truly caps total workers.
- **AWS Raw ECS**: `--max-task-concurrency` was treated as maximum Fargate task count (instance count), not total workers. The heartbeat already multiplied by `ecs_task_cpu`, but the capacity check did not account for workers-per-task. The task cap is now `ceil(MTC / ecs_task_cpu)` and the runner receives `max_instances` so the heartbeat accurately reports `ceil(MTC/ecs_task_cpu) * ecs_task_cpu` actual workers.
- `ORBWorkerProvisioner` and `WorkerManagerRunner` are now constructed inside `_setup(sdk)` where all required information (vCPU count, SDK handle) is available, eliminating post-init mutation via setters.
- Updated documentation to describe the **resolution** concept: MTC is rounded up to the nearest multiple of workers-per-instance, so the actual worker count may exceed MTC by up to `resolution - 1`.
- `WorkerManagerRunner`'s `max_task_concurrency` parameter is renamed to `max_provisioner_units` and `heartbeat_concurrency_multiplier` to `workers_per_provisioner_unit`. A *provisioner unit* is what `WorkerProvisioner.start_worker()` spawns: a worker process for native/symphony, or a cloud instance for ORB/ECS. The heartbeat formula `max_provisioner_units × workers_per_provisioner_unit = maxTaskConcurrency` is now self-documenting.
- Bumped version to 2.1.1 (behavioral correctness fix).

## Related Issue

Closes #698


src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
```diff
@@ -212,35 +188,56 @@ def _build_app_config(self) -> dict:
...
-        sdk = self._orb_pool.sdk
+        self._orb_pool = ORBWorkerProvisioner(
+            config=self._config, max_instances=max_instances, sdk=sdk, template_id=template_id
+        )
+        self._runner = WorkerManagerRunner(
+            address=self._config.worker_manager_config.scheduler_address,
+            name="worker_manager_orb_aws_ec2",
+            heartbeat_interval_seconds=self._config.worker_config.heartbeat_interval_seconds,
+            capabilities=self._config.worker_config.per_worker_capabilities.capabilities,
+            max_task_concurrency=max_instances,
```
magniloquency: I think we should rename this `max_task_concurrency` to something else. Looking into it now.

magniloquency: Done — renamed both parameters. `max_provisioner_units` bounds how many times `start_worker()` may be called; `workers_per_provisioner_unit` is how many workers each spawned unit runs. The heartbeat formula is now self-documenting:

```
maxTaskConcurrency = max_provisioner_units × workers_per_provisioner_unit
```

| Manager | Provisioner unit | `workers_per_provisioner_unit` |
|---|---|---|
| native / symphony | worker process | 1 |
| ORB | EC2 instance | `workers_per_instance` (vCPU count) |
| ECS | Fargate task | `ecs_task_cpu` |

sharpener6 (approved):

PR #745, opened 2026-04-24
feat: migrate ORB worker manager to declarative scaling

magniloquency:
## Summary

- Renames `WorkerProvisioner` to `ImperativeWorkerProvisioner` and introduces `DeclarativeWorkerProvisioner` as an independent ABC; the two interfaces share no inheritance
- Routes `startWorkers`/`shutdownWorkers` commands in `WorkerManagerRunner` only to `ImperativeWorkerProvisioner` instances; routes `setDesiredTaskConcurrency` only to `DeclarativeWorkerProvisioner` instances — unrecognised commands are silently dropped
- Migrates `ORBWorkerProvisioner` to `DeclarativeWorkerProvisioner`: replaces `start_worker`/`shutdown_workers` with `start_units`/`stop_units` and reconciles the EC2 pool to the desired count on each `setDesiredTaskConcurrency` command
- Extracts `extract_desired_count` into `common.py` for shared use across declarative provisioners; uses capability subset matching to sum taskConcurrency across all matching requests
- Tightens `ImperativeWorkerProvisioner` method signatures from `List[bytes]` to `List[WorkerID]`

## Related Issue

#699


src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
```diff
@@ -26,41 +25,45 @@
...
+        self._units: List[str] = []  # EC2 instance IDs of active units
+        self._desired_count: int = 0
+        self._reconcile_lock: asyncio.Lock = asyncio.Lock()
+
+    async def set_desired_task_concurrency(
+        self, requests: List[WorkerManagerCommand.DesiredTaskConcurrencyRequest]
+    ) -> None:
+        own_capabilities = self._config.worker_config.per_worker_capabilities.capabilities
+        self._desired_count = extract_desired_count(requests, own_capabilities)
+        asyncio.create_task(self._reconcile())
```
magniloquency:
```suggestion

        # reconciling the ec2 instances can take some time
        # so we launch a task to handle it in the background
        # `_reconcile_lock` ensures only one runs at a time
        asyncio.create_task(self._reconcile())
```

src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
```diff
@@ -26,41 +25,57 @@
...
-    async def start_worker(self) -> Tuple[List[bytes], Status]:
-        if self._max_instances != -1 and len(self._workers) >= self._max_instances:
-            logging.warning(
-                f"Worker start rejected: at capacity ({len(self._workers)}/{self._max_instances} instances)"
-            )
-            return [], Status.tooManyWorkers
+        self._units: List[str] = []  # EC2 instance IDs of active units
+        self._desired_count: int = 0
+        self._reconcile_lock: asyncio.Lock = asyncio.Lock()
+        self._active_reconcile_task: Optional[asyncio.Task] = None  # strong ref: asyncio holds only weak refs to tasks
```
magniloquency:
```suggestion

        # keeps a strong reference to the running task
        self._active_reconcile_task: Optional[asyncio.Task] = None
```

src/scaler/worker_manager_adapter/common.py
```diff
@@ -9,6 +14,25 @@ class WorkerNotFoundError(Exception):
...
+    Returns 0 if no request matches.
+    """
+    best: Optional[Tuple[int, int]] = None  # (specificity, count)
+    for request in requests:
+        request_capabilities = {entry.key: entry.value for entry in request.capabilities}
+        if request_capabilities.items() <= own_capabilities.items():
+            specificity = len(request_capabilities)
+            if best is None or specificity > best[0]:
+                best = (specificity, request.taskConcurrency)
+    return best[1] if best is not None else 0
```
magniloquency: This function tries to find the "best fit" for capabilities in the request list. For example if we have `own_capabilities={"A", "B"}` and we get the following request: `{{}: 5, {"A"}: 5, {"B"}: 5, {"A", "B"}: 5}` it will find `({"A","B"}, 5)` to be the best match and launch 5 units. Is this what we want?

src/scaler/worker_manager_adapter/mixins.py
```diff
@@ -44,9 +47,34 @@ async def routine(self) -> None: ...
...
+
+    @abstractmethod
+    async def set_desired_task_concurrency(
+        self, requests: List[WorkerManagerCommand.DesiredTaskConcurrencyRequest]
+    ) -> None: ...
+
     @abstractmethod
-    async def start_worker(self) -> Tuple[List[bytes], Status]: ...
+    async def start_units(self, count: int) -> None:
+        """Launch `count` new units and register them internally."""
```
magniloquency:
```suggestion
        """Launch `count` new units."""
```

src/scaler/worker_manager_adapter/common.py
gxuu: No. Because those are concurrency numbers, so you would want (say you don't have fancy grid machine) 5+5+5 workers.

magniloquency: I think you mean 20 workers?

magniloquency: done

sharpener6 (approved):

PR #748, opened 2026-04-24
Add YMQ multicast sockets

rafa-be:
- Add a new `BinderSocker::sendMulticastMessage()` method;
- Add new `SyncSubscriber` and `AsyncPublisher` connectors;
- Fixes a lot of issues on Scaler Top and Scaler WebUI;
- Fixes a segfault that could occur while the initialization of the Python Capnp wrapper.


src/cpp/scaler/protocol/pymod/module_state.cpp
rafa-be, line 1: This function wasn't behaving correctly and caused a `SEGFAULT` when the module wasn't initialized yet.

src/scaler/entry_points/top.py
```diff
@@ -69,19 +74,35 @@ def show_status(status: BaseMessage, screen):
 
     task_manager_table = __generate_keyword_data(
         "task_manager",
-        dict(sorted((k._as_str(), v) for k, v in status.taskManager.stateToCount.items())),
+        dict(
+            sorted(
+                (TaskState(pair.state).name, pair.count)  # type: ignore[attr-defined]
+                for pair in status.taskManager.stateToCount
```
rafa-be, line 80: There is an issue with the Python Capnp wrapper with enum-like value.

When created from Python, these are of type `enum.Enum`, but [capnp.pyi](https://github.com/finos/opengris-scaler/blob/main/src/scaler/protocol/capnp.pyi) defines them as `EnumFieldValue`.

Because of this inconsistency, Mypy complains type-checking these.

src/scaler/protocol/__init__.py
rafa-be, line 1: AI garbage. Absolutely not required.

magniloquency:
IMO, we should rename these to "broadcast" sockets or similar. [Multicast](https://en.wikipedia.org/wiki/Multicast) is the name of a well-known technology in a similar space to us and might confuse people (it confuses me).

rafa-be:
> IMO, we should rename these to "broadcast" sockets or similar. [Multicast](https://en.wikipedia.org/wiki/Multicast) is the name of a well-known technology in a similar space to us and might confuse people (it confuses me).

It's actually multicast messages. We just happen to use it for broadcast messages :/

magniloquency:
> It's actually multicast messages

Hmm, yeah, you're right. Calling it broadcast would be inaccurate. Maybe there isn't a good alternative

sharpener6:
> > IMO, we should rename these to "broadcast" sockets or similar. [Multicast](https://en.wikipedia.org/wiki/Multicast) is the name of a well-known technology in a similar space to us and might confuse people (it confuses me).
> 
> It's actually multicast messages. We just happen to use it for broadcast messages :/

Unicast, multicast, broadcast are all old and still valid technology terms for network, and multicast is accurate in this case

sharpener6 (approved):

PR #766, opened 2026-05-01
Make `scaler` run on Windows

### LINKED ISSUE #755: Enable worker working on Windows
Current we release windows wheel, but the worker doesn't work, so we need make windows worker works on windows

gxuu:
# READY FOR REVIEW.

Closes #755

Many things are being done in this PR:
- Turn on the Windows unit test on CI.
- Make `scaler` be able to run on Windows.
- Fixed a few cmake build bugs on Windows (preset wouldn't build).
- Daemon thread on each worker on Windows to soft suspend.
- Daemon thread on Scheduler on Windows to mimic SIGTERM behavior.

### DIFF
--- a/.github/actions/run-test/action.yml
+++ b/.github/actions/run-test/action.yml
@@ -15,7 +15,7 @@ runs:
 
     - name: Install pydivert and pywin32 for MITM tests (Windows)
       if: inputs.os == 'Windows'
-      shell: pwsh
+      shell: bash
       run: uv pip install --system pydivert pywin32
 
     - name: Run C++ Tests (Linux/macOS)
@@ -29,14 +29,12 @@ runs:
       run: ./scripts/test.ps1 -C Debug
 
     - name: Build Wheel
-      if: inputs.os != 'Windows'
       shell: bash
       run: |
         uv pip install --system build
         python -m build --wheel
 
     - name: Install Wheel
-      if: inputs.os != 'Windows'
       shell: bash
       run: |
         wheel_file=$(ls dist/*.whl | head -n 1)
@@ -44,7 +42,6 @@ runs:
         uv pip install --system "$wheel_file[all]"
 
     - name: Run Unittests
-      if: inputs.os != 'Windows'
       shell: bash
       run: |
         python -m unittest discover -v tests -t .
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -106,7 +106,7 @@
       "name": "windows-x64",
       "displayName": "Windows x64 build",
       "configurePreset": "windows-x64",
-      "configuration": "Release",
+      "configuration": "Debug",
       "condition": {
         "type": "equals",
         "lhs": "${hostSystemName}",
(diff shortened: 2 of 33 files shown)


src/cpp/scaler/ymq/configuration.h
```diff
@@ -12,7 +12,7 @@ namespace ymq {
 constexpr std::array<uint8_t, 4> magicString {'Y', 'M', 'Q', 1};
 
 constexpr size_t defaultClientMaxRetryTimes = 8;
-constexpr std::chrono::milliseconds defaultClientInitRetryDelay {100};
+constexpr std::chrono::milliseconds defaultClientInitRetryDelay {200};
```
gxuu: This value is doubled as 100ms seems too aggressive - Windows CI would fail sometimes due to connect retry exceeding limit.

CMakePresets.json
```diff
@@ -106,7 +106,7 @@
       "name": "windows-x64",
       "displayName": "Windows x64 build",
       "configurePreset": "windows-x64",
-      "configuration": "Release",
+      "configuration": "Debug",
```
gxuu, line 109: This preset, when first added, was meant to make the developers life easier. We always want to build debug when develop locally. Hence the revertion.

.github/actions/run-test/action.yml
```diff
@@ -15,7 +15,7 @@ runs:
 
     - name: Install pydivert and pywin32 for MITM tests (Windows)
       if: inputs.os == 'Windows'
-      shell: pwsh
+      shell: bash
```
sharpener6, line 18: windows use bash?

gxuu: Yes. Windows can use bash indeed!

src/cpp/scaler/ymq/internal/accept_server.cpp
```diff
@@ -109,14 +109,20 @@ void AcceptServer::disconnect() noexcept
 
     std::optional<std::string> pipeName {};
     if (auto* pipeServer = std::get_if<scaler::wrapper::uv::PipeServer>(&_state->_server.value())) {
-        *pipeName = UV_EXIT_ON_ERROR(pipeServer->getSockName());
+        // Assign to the optional itself; `*pipeName = ...` would dereference an empty optional (UB).
+        // libstdc++ in debug mode traps this as an _M_is_engaged() assertion (SIGABRT); MSVC reports
+        // it as an access violation during socket shutdown.
+        pipeName = UV_EXIT_ON_ERROR(pipeServer->getSockName());
```
gxuu, line 112: On Linux, it seems like this branch was never taken. This is understandable because the async util under the hood works differently.

gxuu (commented):
In general, the approach making scaler be able to run on Windows is to set up a bunch of `if sys.platform == "win32"`.

This is not ideal. But getting rid of signal disposition (i.e. unified async note point in userspace) in this PR is FOR SURE out of scope.

sharpener6 (approved):

PR #792, opened 2026-05-26
feat: zero-copy Python buffer support in YMQ (#789)

### LINKED ISSUE #789: Make constructing YMQ `Bytes` from Python zero-copy
When a `Bytes` object is created from Python we use the buffer API to get a pointer to the data which we then copy into the object. We could instead store this pointer directly to enable creating `Bytes` from Python that are zero-copy

https://github.com/finos/opengris-scaler/blob/99388df239fb9dc70e9fa2d4182e03a78b103393/src/cpp/scaler/ymq/pymod/bytes.h#L18-L38

magniloquency:
## Summary

- Split the monolithic concrete `Bytes` class into an abstract base (`Bytes`) and two concrete implementations: `BufferedBytes` (C++-owned heap memory) and `PyBufferBytes` (zero-copy Python `Py_buffer` view)
- All YMQ socket and message APIs updated to use `std::unique_ptr<Bytes>` for ownership throughout
- `PyBytes_init` in the Python C module now transfers `Py_buffer` ownership to `PyBufferBytes` instead of copying — eliminating the O(message-size) allocation that occurred on every Python→YMQ send
- `PyBufferBytes` acquires the GIL only in its destructor before calling `PyBuffer_Release`, so C++ threads can safely hold and destroy Python buffers
- `Bytes` special members updated: `~Bytes()` is now a non-pure virtual default; move and copy constructor/assignment are defaulted; `BufferedBytes` and `PyBufferBytes` drop their redundant custom move implementations in favour of `= default`
- `asString` promoted from a free function to a pure virtual method on `Bytes`, with implementations in both concrete classes; all call sites updated

## Related Issue

Closes #789

## Follow-up: Eliminate pointer overhead in the object storage receive path

The object storage server only ever handles `BufferedBytes` in practice — the YMQ receive path unconditionally allocates `BufferedBytes` in `MessageConnection::recv()`, but erases the type to `unique_ptr<Bytes>` at the `Message` API boundary. This causes three heap allocations per stored object (the `BufferedBytes` object, its data buffer, and a separate `shared_ptr` control block when converting `unique_ptr` to `shared_ptr` in `ObjectManager::setObject`), plus virtual dispatch on every `data()`/`size()` call through the stored pointer.

The fix is a chain of three changes:

1. **Change `Message` to hold `BufferedBytes` by value** — `Message` is exclusively a receive-side type (the send API takes `unique_ptr<Bytes>` directly, never a `Message`), so no `BufferedMessage` split is needed. `RecvOperation::_buffer` in `MessageConnection` becomes `unique_ptr<BufferedBytes>` to preserve the nullable sentinel without introducing `std::optional` overhead.

2. **Change `ObjectPayload` to `ymq::BufferedBytes` and `setObject` to take it by value** — with `Message::payload` now a concrete type, the server can move it directly into `setObject` with no cast needed.

3. **Use `make_shared<const BufferedBytes>` in `ObjectManager`** — collapses the `BufferedBytes` object and its `shared_ptr` control block into a single allocation, reducing per-object allocations from three to two (control block + object together, data buffer separately).

The `pymod` receive handlers (`connector_socket.h`, `binder_socket.h`) and `PyBytes::bytes` field will also need updating to match the new `Message` field types.


src/cpp/scaler/ymq/bytes.h
```diff
@@ -1,151 +1,27 @@
...
-    }
+    virtual ~Bytes() noexcept = 0;
 
-    friend std::strong_ordering operator<=>(const Bytes& x, const Bytes& y) noexcept
-    {
-        return std::lexicographical_compare_three_way(x._data, x._data + x._len, y._data, y._data + y._len);
-    }
+    Bytes()                            = default;
+    Bytes(Bytes&&) noexcept            = default;
+    Bytes& operator=(Bytes&&) noexcept = default;
```
rafa-be: Shouldn't this be `virtual = 0`?

src/cpp/scaler/ymq/buffered_bytes.h
```diff
@@ -0,0 +1,75 @@
...
+    std::unique_ptr<uint8_t[]> _data;
+    size_t _size {0};
+};
+
+inline std::optional<std::string> as_string(const Bytes& b) noexcept
+{
+    if (!b.data())
+        return std::nullopt;
+    return std::string(reinterpret_cast<const char*>(b.data()), b.size());
+}
```
rafa-be: Non longer required if `asString()` is part of `Bytes`.

src/cpp/scaler/ymq/binder_socket.cpp
```diff
@@ -43,7 +45,9 @@ void BinderSocket::shutdown(ShutdownCallback onShutdownCallback) noexcept
         // Fail all pending send callbacks
         for (auto& [_, pendingMessages]: state->_pendingSendMessages) {
             for (auto& pendingMessage: pendingMessages) {
-                pendingMessage.onMessageSent(std::unexpected {Error {Error::ErrorCode::SocketStopRequested}});
+                pendingMessage.onMessageSent(
+                    std::unexpected {Error {Error::ErrorCode::SocketStopRequested}},
+                    std::move(pendingMessage.messagePayload));
```
rafa-be, line 52: Why this new argument?

src/cpp/scaler/object_storage/object_manager.cpp
```diff
@@ -2,31 +2,29 @@
...
-    const ObjectID& objectID, ObjectPayload&& payload) noexcept
+    const ObjectID& objectID, std::unique_ptr<ObjectPayload> payload) noexcept
 {
     if (hasObject(objectID)) {
         // Overriding object: delete old first
         deleteObject(objectID);
     }
 
-    ObjectHash hash = std::hash<ObjectPayload> {}(payload);
+    ObjectHash hash = computePayloadHash(*payload);
```
gxuu: Should this be `const`?

src/cpp/scaler/object_storage/object_storage_server.cpp
```diff
@@ -290,7 +291,7 @@ void ObjectStorageServer::processDuplicateRequest(
             "payload length should be size_of(ObjectID)=" + std::to_string(ObjectID::bufferSize()));
     }
 
-    ObjectID originalObjectID = ObjectID::fromBuffer(payload);
+    ObjectID originalObjectID = ObjectID::fromBuffer(*payload);
```
gxuu: Should this be `const`?

src/cpp/scaler/ymq/bytes.h
```diff
@@ -1,151 +1,36 @@
...
 #include <cstddef>
 #include <cstdint>
-#include <cstring>
-
-// C++
 #include <optional>
 #include <string>
 
-// First-party
-#include "scaler/ymq/common.h"
```
gxuu, line 17: If this file is no longer referenced by any other files, consider removing it.

src/cpp/scaler/object_storage/object_manager.cpp
magniloquency: done

src/cpp/scaler/object_storage/object_storage_server.cpp
magniloquency: done

src/cpp/scaler/ymq/bytes.h
magniloquency: removed

src/cpp/scaler/ymq/common.h
```diff
@@ -1,35 +0,0 @@
...
-
-using Errno = int;
-
-[[nodiscard("Memory is allocated but not used, likely causing a memory leak")]]
-inline uint8_t* datadup(const uint8_t* data, size_t len) noexcept
-{
-    uint8_t* dup = new uint8_t[len];  // we just assume allocation will succeed
-    std::memcpy(dup, data, len);
-    return dup;
-}
```
rafa-be, line 19: Nice. We know it's 0 copy because you removed the function that did the copies! 😅

src/cpp/scaler/ymq/bytes.h
```diff
@@ -1,151 +1,36 @@
...
-    {
-        return _data;
-    }
-};
+inline std::optional<std::string> asString(const Bytes& b)
+{
+    if (!b.data())
+        return std::nullopt;
+    return std::string(reinterpret_cast<const char*>(b.data()), b.size());
+}
```
rafa-be: I think this could be a non-virtual method of `Bytes`.

That would simplify a lot some of these calls:

```c++
// Old
socket.sendMessage(scaler::ymq::asString(*receivedMsg.address).value(), std::move(receivedMsg.payload));

// New
socket.sendMessage(receivedMsg.address->asString().value(), std::move(receivedMsg.payload));
```

src/cpp/scaler/ymq/buffered_bytes.h
```diff
@@ -0,0 +1,75 @@
...
+    {
+        return _data == nullptr;
+    }
+
+    std::optional<std::string> as_string() const
+    {
+        if (is_null())
+            return std::nullopt;
+        return std::string(reinterpret_cast<const char*>(_data.get()), _size);
+    }
```
rafa-be, line 56: I think these two methods could be added to the `Bytes` class as non-virtual methods, relying on the `data()` and `size()`.

Also, these should be `camelCase()`.

src/cpp/scaler/ymq/internal/message_connection.cpp
```diff
@@ -303,29 +308,31 @@ void MessageConnection::onRead(std::expected<std::span<const uint8_t>, scaler::w
...
 }
 
-void MessageConnection::onRemoteIdentity(Bytes payload) noexcept
+void MessageConnection::onRemoteIdentity(std::unique_ptr<Bytes> payload) noexcept
 {
     assert(connected());
     assert(!established());
 
-    Identity receivedIdentity = payload.as_string().value();
+    Identity receivedIdentity = std::string(reinterpret_cast<const char*>(payload->data()), payload->size());
```
rafa-be: Why not using ` asString(const Bytes& b)`?

src/cpp/scaler/ymq/bytes.h
rafa-be: or `deleted`. You don't want `Bytes` instances to be created.

src/cpp/scaler/ymq/bytes.h
```diff
@@ -1,151 +1,36 @@
...
+    virtual ~Bytes() noexcept = 0;
 
-    Bytes(Bytes&& other) noexcept: _data(other._data), _len(other._len)
-    {
-        other._data = nullptr;
-        other._len  = 0;
-    }
+    Bytes()                            = default;
+    Bytes(Bytes&&) noexcept            = default;
+    Bytes& operator=(Bytes&&) noexcept = default;
```
rafa-be: I think these should be `virtual`

src/cpp/scaler/ymq/bytes.h
```diff
@@ -1,151 +1,36 @@
...
+    virtual size_t size() const noexcept         = 0;
+};
 
-    [[nodiscard("Allocated Bytes is not used, likely causing a memory leak")]]
-    static Bytes alloc(size_t len)
-    {
-        auto ptr = new uint8_t[len];  // we just assume the allocation will succeed
-        return Bytes {ptr, len};
-    }
+inline Bytes::~Bytes() noexcept = default;
```
rafa-be: What is this?

src/cpp/scaler/ymq/bytes.h
rafa-be: I think it would be fine to have that one as a member function, relying on the `virtual` `data()` and `size()` methods.

That would make `Bytes` a true mixin class, and I think that's fine.

src/cpp/scaler/ymq/bytes.h
magniloquency: done. i had to leave `Bytes()` as default though.

rafa-be (approved):
LGTM

gxuu (dismissed):

rafa-be (dismissed):

e117649 (dismissed):

e117649 (approved):

PR #793, opened 2026-05-27
YMQ sockets support OpenSSL

rafa-be:
YMQ sockets can now support TLS sockets, using a `tls://` address.

This PR also fixes a lot of issues related to how the `SecureSocket` shutdown work that came up when enabling YMQ unit test on these sockets.

Scaler **does not use TLS sockets yet**. A following PR will add the CLI config (certificates, keys) and addressing changes required to configure these.


src/cpp/scaler/wrapper/openssl/secure_socket.cpp
rafa-be, line 1: Most of the changes in this file are related to:

- store the internal state of the `SecureSocket` class inside a `std::share_ptr<State>`. That is required for the socket to outlive its own class, and complete the TLS/SSL shutdown handshake correctly;
- handling additional corner cases related to the shutdown handshake, in particular the handling of pending read and write buffers;
- making the `SecureSocket` behavior match the `TCPSocket` behavior, especially in relation to error codes.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
rafa-be, line 1: @magniloquency these fix a few bugs I had while enabling additional YMQ tests for WebSockets. All related to shutdown of sockets.

sharpener6:
@rafa-be, please resolve the conflict

rafa-be:
> @rafa-be, please resolve the conflict

Done.

e117649 (dismissed):

src/cpp/scaler/ymq/address.cpp
```diff
@@ -139,38 +156,68 @@ std::expected<std::string, Error> Address::toString() const noexcept
...
+    if (!_secure) {
+        return std::nullopt;
     }
 
-    if (address.starts_with(_wssPrefix)) {
-        return details::fromWSString(address.substr(_wssPrefix.size()), true);
+    if (_tlsConfig.has_value()) {
+        auto context = _tlsConfig->getSSLContext();
+        if (!context.has_value()) {
+            unrecoverableError(context.error());
```
magniloquency: Is it possible for us to propagate an error here? I think this could result from e.g. a missing cert, that's something the user could act on.

tests/cpp/ymq/test_binder_socket.cpp
```diff
@@ -468,3 +477,21 @@ TEST_F(YMQBinderSocketTest, StopRequested)
...
+{
+    std::vector<std::string> transports;
+    transports.push_back("tcp");
+    transports.push_back("tls");
+    transports.push_back("ws");
+#ifdef __linux__
+    transports.push_back("ipc");
+#endif
+    return transports;
+}
```
magniloquency: Should we add `wss` here?

magniloquency: This seems identical to the other `Get*Transports(...)` functions, maybe we should have just one?

src/cpp/scaler/ymq/address.cpp
```diff
@@ -139,38 +156,68 @@ std::expected<std::string, Error> Address::toString() const noexcept
...
+        }
+        return std::move(context.value());
     }
 
-    return std::unexpected {Error {
-        Error::ErrorCode::InvalidAddressFormat, "Address must start with 'tcp://', 'ipc://', 'ws://', or 'wss://'"}};
+    // No TLS config provided, use default SSL context.
+    return UV_EXIT_ON_ERROR(scaler::wrapper::openssl::SSLContext::init());
```
magniloquency: Is this desirable default behaviour? It's possible to create and bind a TLS socket without any context, but then if this default path creates a TLS context without certs, we won't get a failure until the handshake fails. Not necessarily a bad thing, it's just that we could detect that kind of issue earlier. Thoughts?

rafa-be: For binding TLS sockets, these usually need certificates by default.

But you can configure them to not use certificates. In that case, TLS only provides encryption and no authentication.

Not sure if we will use this, but at least this implementation will allow it.

tests/cpp/ymq/test_binder_socket.cpp
rafa-be: WebSockets don't support TLS yet. That's a future PR.

rafa-be: Good catch. These use to be different while I was developing, but are the same now.

src/cpp/scaler/ymq/address.cpp
rafa-be: Done

magniloquency (approved):

PR #795, opened 2026-05-28
Fix #729 OSS exiting on SIGPIPE events on Linux

### LINKED ISSUE #729: ObjectStorageLoggingTest.TestServerLogsStartToFile crashes with SIGPIPE
## Summary

`test_object_storage_server` crashes with SIGPIPE in `ObjectStorageLoggingTest.TestServerLogsStartToFile` on ubuntu-latest. This is a pre-existing race condition surfaced in [CI run 24531890568](https://github.com/finos/opengris-scaler/actions/runs/24531890568/job/71716951372).

```
3 - test_object_storage_server (SIGPIPE)
```

## Root Cause

`ObjectStorageServer::setupSignalHandling()` (`src/cpp/scaler/object_storage/object_storage_server.cpp`) only installs a handler for `SIGTERM`. `SIGPIPE` is left with its default disposition: **terminate the process**.

`TestServerLogsStartToFile` is the only test that creates a client that immediately disconnects without completing the YMQ handshake. This triggers a race:

1. The server's `MessageConnection` queues its handshake writes (`sendHandshake()` → `processSendQueue()` → `uv_write()`).
2. The client disconnects before the handshake completes.
3. During socket teardown, a write to the now-closed client connection generates `SIGPIPE` — killing the entire test binary.

Although libuv uses `MSG_NOSIGNAL` for established TCP socket writes, there is a window during the teardown sequence (handshake write in-flight + `shutdownClient()` called) where a write can slip through without that protection.

## Fix

Add `SIG_IGN` for `SIGPIPE` in `setupSignalHandling()`:

```cpp
void setupSignalHandling()
{
    // Ignore SIGPIPE — broken client connections should be handled via error codes, not signals
    signal(SIGPIPE, SIG_IGN);

    struct sigaction sa {};
    sa.sa_handler = handleSigTerm;
    sigemptyset(&sa.sa_mask);
    sa.sa_flags = 0;

    if (sigaction(SIGTERM, &sa, nullptr) == -1) {
        perror("sigaction");
    }
}
```

This is standard practice for server applications. Write errors from disconnected clients are already handled gracefully via `UV_EPIPE` in `onWriteDone()`.

rafa-be:
The OSS occasionally prematurely crashes with a `SIGPIPE` error on Linux (see #729 ). 

This is because, on Linux, a [`SIGPIPE` signal is raised](https://pod.tst.eu/http://cvs.schmorp.de/libev/ev.pod#The_special_problem_of_SIGPIPE) when a remote socket is destroyed while writing to it.

macOS/BSDdon't have that issue and just have their `write()` syscall return the `EPIPE` error.

This PR makes the OSS ignore these `SIGPIPE` signals. This is safe the corresponding YMQ's `write()` will receive and handle a a `EPIPE` error.


gxuu:
https://github.com/finos/opengris-scaler/issues/729 is related and there were discussion about this problem. Please consider my comment in that issue, and see if that is a more proper way to tackle this.

sharpener6 (approved):

PR #737, opened 2026-04-21
WASM Scaler client and demo notebooks with Pyodide/JupyterLite

e117649:
Requires websocket support and should be merged after https://github.com/finos/opengris-scaler/pull/739

PR #796, opened 2026-05-28
fix(worker): handle deserialize failure in ObjectCache.add_object

e117649:
This change was part of https://github.com/finos/opengris-scaler/pull/737

When the worker's serializer raised (e.g. the worker venv is missing a module that the client side has, like numpy), the ObjectCache silently swallowed the exception via 'except Exception' and then immediately referenced the unbound 'deserialized' local, raising UnboundLocalError and crashing the entire processor.

Store the failure as a DeserializeObjectError sentinel and re-raise it on get_object so the dependent task fails with a meaningful error instead of taking down the processor.

Adds tests/worker/agent/processor/test_object_cache.py covering the unbound-local regression and the new DeserializeObjectError surfacing.


src/scaler/worker/agent/processor/object_cache.py
```diff
@@ -65,9 +65,12 @@ def add_object(self, client: ClientID, object_id: ObjectID, object_bytes: bytes)
             self.add_serializer(client, cloudpickle.loads(object_bytes))
         else:
             try:
-                deserialized = self.deserialize(client, object_bytes)
-            except Exception:  # noqa
+                deserialized: Any = self.deserialize(client, object_bytes)
+            except Exception as exc:  # noqa: BLE001
```
sharpener6, line 69: I think this case we should response with TaskFailed

e117649: It's already handled lazily at the get_object stage before the function runs, unless you want an early return but then we would also need more logic to skip __process_task and add a similar check to has_object resulting in multiple paths for the same error.

sharpener6 (approved):

PR #802, opened 2026-06-01
Fix timeout error that happens indeterminately

gxuu:
# READY FOR REVIEW.

Binder side needs to address connection abort/disconnect rather than ignoring the reason.


src/cpp/scaler/ymq/binder_socket.cpp
```diff
@@ -144,6 +154,17 @@ void BinderSocket::closeConnection(Identity remoteIdentity) noexcept
...
+        // Symmetric with onRemoteDisconnect: any subsequent send to this identity should fail
+        // fast rather than wait for a reconnect that may never come.
+        state->_disconnectedIdentities.insert(node.key());
+        auto pendingIt = state->_pendingSendMessages.find(node.key());
+        if (pendingIt != state->_pendingSendMessages.end()) {
+            for (auto& pending: pendingIt->second) {
+                pending.onMessageSent(std::unexpected {Error {Error::ErrorCode::ConnectorSocketClosedByRemoteEnd}});
+            }
+            state->_pendingSendMessages.erase(pendingIt);
+        }
```
rafa-be: I think this *might* be simplified with


```suggestion
ConnectionID connectionID = node.mapped();

onRemoteDisconnect(std::move(state), connectionID, DisconnectReason::Disconnected);
```

src/scaler/io/ymq_async_binder.py
```diff
@@ -62,7 +62,14 @@ async def routine(self):
...
         self.__count_sent(message.__class__.__name__)
-        await self._socket.send_message(to.decode(), Bytes(serialize(message)))
+        try:
+            await self._socket.send_message(to.decode(), Bytes(serialize(message)))
+        except ConnectorSocketClosedByRemoteEndError:
+            # Peer has disconnected. Treat the send as a no-op: heartbeat-echo / task-dispatch
+            # callers expect peer-gone to be a routine event handled by their own timeout/cleanup
+            # paths. Re-raising here would bubble up through binder.routine and asyncio.gather and
+            # tear down the whole scheduler for what is a normal peer departure.
+            pass
```
rafa-be, line 68: Maybe we actually want to propagate this exception, and catch it inside the Scaler's connector classes (e.g. `YMQAsyncBinder`), instead of doing it in the Python YMQ wrapper.

That's because I think it's better if the YMQ Python wrapper's behavior maps 1:1 to the C++ class.

rafa-be (dismissed):

src/cpp/scaler/ymq/binder_socket.h
```diff
@@ -97,6 +98,11 @@ class BinderSocket {
         std::map<ConnectionID, std::unique_ptr<internal::MessageConnection>> _connections {};
         std::map<Identity, ConnectionID> _identityToConnectionID {};
 
+        // Identities that completed identity exchange and then disconnected. Sends to these are
+        // failed immediately (instead of queued in _pendingSendMessages) to avoid a hang when the
+        // peer has gracefully gone away. Cleared on reconnect.
+        std::set<Identity> _disconnectedIdentities {};
```
rafa-be: We might want to cleanup that set from time to time, or else it might growth quite huge.

When users use nested tasks, a lot of Scaler clients might be created and then destroyed, adding a lot of entries to this set.

I'd suggest we add a expiry date for these entries (e.g. 15 minutes) and remove the expired entries from time to time (e.g. at each user-exposed method call).

If you don't fix it in this PR, maybe create an issue so that we keep track of the problem.

src/scaler/io/ymq_async_binder.py
gxuu: You mean at each and all call site?

src/cpp/scaler/ymq/binder_socket.cpp
gxuu: good catch

src/cpp/scaler/ymq/binder_socket.h
gxuu: done, but with a shorter expiry date.

src/cpp/scaler/ymq/binder_socket.cpp
```diff
@@ -177,14 +190,55 @@ void BinderSocket::onRemoteIdentity(
...
+
+    // Remember the identity so any sendMessage call that lands *after* this disconnect (because
+    // Python is just catching up on messages libuv already buffered) fails fast instead of
+    // queueing forever in _pendingSendMessages and hanging the asyncio loop. The set is
+    // bounded by a TTL purge driven from here (insert-time) so it cannot grow without bound
+    // even under workloads that churn many short-lived peers (e.g. nested-task clients).
+    constexpr auto kDisconnectedIdentityTTL = std::chrono::seconds {60};
+    const auto now                          = std::chrono::steady_clock::now();
+    while (!state->_disconnectedIdentityInsertions.empty() &&
+           now - state->_disconnectedIdentityInsertions.front().first > kDisconnectedIdentityTTL) {
```
rafa-be: `kDisconnectedIdentityTTL` should go into [configuration.h](https://github.com/finos/opengris-scaler/blob/main/src/cpp/scaler/ymq/configuration.h) I think.

src/cpp/scaler/ymq/binder_socket.cpp
```diff
@@ -177,14 +190,55 @@ void BinderSocket::onRemoteIdentity(
...
+        const auto& [expiredTs, expiredId] = state->_disconnectedIdentityInsertions.front();
+        // Only drop the map entry if its timestamp matches; otherwise the peer re-disconnected
+        // and the current entry is fresher than this expired deque slot.
+        if (auto it = state->_disconnectedIdentities.find(expiredId);
+            it != state->_disconnectedIdentities.end() && it->second == expiredTs) {
+            state->_disconnectedIdentities.erase(it);
+        }
+        state->_disconnectedIdentityInsertions.pop_front();
+    }
+    state->_disconnectedIdentities[remoteIdentity] = now;
```
rafa-be, line 223: Maybe move this to a new method (`purgeExpiredIdentities()` ?).

src/scaler/io/ymq_async_binder.py
rafa-be: Oh, nevermind, I missread and tought this was in the Python wrapper!

rafa-be (dismissed):

sharpener6 (approved):

PR #807, opened 2026-06-04
Fix stack overflow when calling __getattr__ on a Python constructed Capnp object

rafa-be:
Fixes a C++ segfault/stack overflow that occurred in the Capnp Python wrapper:

image


src/cpp/scaler/protocol/pymod/utility.cpp
```diff
@@ -197,12 +208,14 @@ bool load_lazy_struct_metadata(
...
-    if (!source) {
-        PyErr_Clear();
+    // We cannot call get_attr() until we are sure CAPNP_SOURCE_ATTR is defined, or else we will trigger a recursive
+    // call to load_lazy_struct_metadata().
+    if (!has_dict_attr(self, CAPNP_SOURCE_ATTR)) {
         PyErr_Format(PyExc_AttributeError, "%s not found: object is not a lazy Cap'n Proto struct", CAPNP_SOURCE_ATTR);
         return false;
     }
+
+    source = get_attr(self, CAPNP_SOURCE_ATTR);
```
rafa-be, line 218: The module registers `capnp_struct_get_attr()` as a handler for `__getattr__`, which ultimately calls this function (`load_lazy_struct_metadata()`) for **uninitialized fields**.

However, this function calls `__getattr__(CAPNP_SOURCE_ATTR)`, thus creating a recursive call if `CAPNP_SOURCE_ATTR` isn't defined. That's the case for Python constructed Capnp objects.

I tried to refactor the module to avoid this behavior, but it's a hard task. Instead, this small "hack" short-circuit recursive calls for Python constructed objects.

rafa-be: And why it happens on some platforms but not on others:

Python is checking the depth level of calls to `PyXXX()` functions, and will raise a `RecursionError` exception when reaching a 10,000 deep call.

That `RecursionError` exception would make `load_lazy_struct_metadata()` return `false`, and the calling function will correctly fallback as if the `CAPNP_SOURCE_ATTR` field didn't exist.

However, in some cases and on some platforms, the recursive call will stack overflow before the `RecursionError` limit is reached, crashing the program with `SIGSEGV` instead.

sharpener6:
@rafa-be please bump to 2.4.1, if I do it I will not able to approve

rafa-be:
> @rafa-be please bump to 2.4.1, if I do it I will not able to approve

Done

sharpener6 (approved):

PR #815, opened 2026-06-08
Use arm64 native runner instead of QEMU

sharpener6:


.github/workflows/publish-artifact.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         include:
           - os: ubuntu-latest
             cibw_archs: x86_64
-          - os: ubuntu-latest
```
magniloquency, line 19: Doesn't this mean we'll stop releasing x64?

gxuu (approved):

.github/workflows/publish-artifact.yml
sharpener6: look carefully

PR #824, opened 2026-06-10
Fix unclean shutdown leaking processes and bound ports

e117649:
The scaler CLI and scaler_scheduler ran the scheduler as a grandchild process that nothing signalled on shutdown: terminating the wrapper orphaned the scheduler, which kept its ports bound forever and wedged interpreter exit via the multiprocessing resource tracker. The CLI also had no SIGTERM handler (kill leaked the whole cluster), did not cover the startup window on interrupt, and joined children without a timeout.

- run the scheduler in the current process via run_scheduler(); SchedulerProcess delegates to it, no nested process
- scaler CLI: route SIGTERM through the Ctrl-C path, start children inside the try block, and tear down in reverse startup order with a join deadline, SIGKILL escalation, and orphaned-descendant reaping
- YMQ: AcceptServer::init() returns std::expected; bind/listen failures (e.g. EADDRINUSE) propagate to the bind callback and raise a catchable SysCallError in Python instead of exiting the process uncleanly
- tests: shutdown regression tests for the CLI; nested task tests use a random free port instead of a hardcoded one


rafa-be (dismissed):

sharpener6:
@rafa-be please review this PR

e117649:
> @rafa-be please review this PR

I don't know if you want him to review it more thoroughly, but the only change since he last approved it was fixing the CI test on macOS 32b210d029716cba26bc38112ce47f7c0e8bbdbc

rafa-be (approved):
All the C++ code looks good to me.

My only concern is about the `_shutdown_processes()` function: I'm not sure this fix isn't actually hiding an underlying issue/bug in the Scheduler process.

What I mean is that we should'nt expect Scaler processes to not properly exit.

But I didn't have much time to think about it. If you both think that this is the proper fix, then it's good to me.

sharpener6 (approved):

PR #833, opened 2026-06-15
Waterfall policy: make max_task_concurrency optional in rule format

magniloquency:
## Summary

- When `max_task_concurrency` is omitted from the CSV `policy_content`, the rule's field is `None` and the policy falls back to the value reported by the worker manager heartbeat.
- An explicit value still acts as a soft cap (`min` of rule cap and heartbeat cap).
- Includes a minor style fix: wrap long ternary in waterfall scaling policy.


rafa-be (dismissed):

sharpener6:
@magniloquency Bump the version

magniloquency:
@sharpener6, bumped.

gxuu (approved):

PR #835, opened 2026-06-15
fix: report coordinator desired count in heartbeat to prevent missed scale-down

magniloquency:
## Summary

- The scaling policies suppressed sending a `WorkerManagerCommand` when effective desired workers matched the number of connected workers. This caused scale-down to be skipped when a provisioner had `desired > 0` but workers hadn't connected yet (e.g. EC2 instance still booting), leaving orphaned instances indefinitely.
- Adds `currentDesiredWorkers` to `WorkerManagerHeartbeat` so the scheduler compares against the coordinator's actual desired state instead of the count of connected workers.
- All three policies (vanilla, capability, waterfall) and all seven provisioner implementations are updated.

Depends on: #834 (ORB EC2 max_task_concurrency)

## Test plan

- [ ] `test_scaling.py` and `test_waterfall_scaling.py` pass with updated assertions
- [ ] Verify scale-down is triggered when a provisioner reports `desired > 0` but no workers are connected

### DIFF
--- a/docs/html_extra/launchpad/provisioner.js
+++ b/docs/html_extra/launchpad/provisioner.js
@@ -147,6 +147,10 @@ worker_manager_id = "${wm.id}"
 
       if (wm.type === "orb_aws_ec2") {
         var req = (wm.requirements || "").trim();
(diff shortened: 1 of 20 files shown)


magniloquency:
superseded by https://github.com/finos/opengris-scaler/pull/837

PR #836, opened 2026-06-15
Launchpad: policy panel, config load/save, OCI improvements, and UX fixes

magniloquency:
## Summary

- **Policy panel**: interactive editor for `waterfall_v1` and `simple` policies in the Launchpad UI.
- **Config load/save**: download current config as TOML; load an existing TOML config back into the UI.
- **Draggable WM reordering**: worker managers can be reordered via drag-and-drop.
- **OCI raw improvements**: budget/instance cap selector; reduce default memory from 30 GB to 8 GB.
- **UX**: always show Download/Load config buttons regardless of provisioning phase; fetch full scheduler log instead of last 150 lines.
- **Dev mode (`?dev`)**: appending `?dev` to the URL enables sessionStorage-backed credential persistence across page reloads, supporting a hot-reload dev cycle without re-entering credentials each time. A DEV badge appears in the nav bar. Credentials are cleared when the tab closes; the existing "not stored by this application" disclaimer remains accurate for normal usage.
- **Scaling: remove effective-equals-current early-return**: all scaling policies now unconditionally emit a `setDesired` command so the worker manager always receives the authoritative desired count, even when it matches the current worker count.
- **Launchpad: unified TOML builder**: `buildConfigToml` and the EC2 user-data TOML were hand-rolled separately and had drifted — the user-data path was missing the `[scheduler.policy]` section entirely, silently ignoring waterfall policy selection on real EC2 deploys. Both now go through shared helpers and serialize via `TOML.stringify`.
- **EC2 worker manager user-data**: install `git` (required for pip git-URL installs); switch from `python -m venv` + `pip` to `uv`; build Cap'n Proto from source via `library_tool.sh` using `gcc14` for git-install mode.

### Bug fixes

- **Scheduler policy TOML structure**: policy keys (`policy_engine_type`, `policy_content`) were emitted under `[scheduler.policy]` but the Python scheduler reads them as flat keys under `[scheduler]` — Waterfall policy was silently dropped on every real deploy.
- **Incremental log fetching**: replaced `cat` (truncated from the head by SSM) with byte-offset–based fetching (`tail -c +$((offset+1))`); advances offset on new content and loops immediately to drain remaining chunks before resuming the normal poll interval.
- **Scheduler log state preservation**: split the log `useEffect` into separate reset and polling effects so returning to the Logs tab no longer re-fetches from byte 0.
- **`network_backend` round-trip**: previously omitted from TOML for zmq, making reload impossible; now always written explicitly as `"ymq"` or `"zmq"`, and `configFromToml` / `handleLoadConfig` restore it correctly. Also fixed a `tcp_zmq` typo (valid enum value is `zmq`).
- **Worker manager ID collisions**: ID generation now checks against the live list and increments until the candidate is free; a unique-ID check in pre-launch validation blocks launch with a visible error if duplicates slip through.
- **`_extract_git_url_and_branch` rewrite**: the old `rfind("@")` mis-split `ssh user@host` and token-auth URLs; replaced with PEP 508 `Requirement` parsing and `urlsplit` to locate the `@branch` separator only in the path component. Adds unit tests for https, ssh `user@host`, token auth, env markers, and comments.
- **OCI raw defaults alignment**: memory and OCPU defaults in the UI now match the provisioner's values.
- **ORB EC2 `max_task_concurrency`**: multiplied by vCPU count so the per-worker limit scales with instance size.
- **`aws_raw_ecs` requirements round-trip**: `ecs_python_requirements` was not read back by `configFromToml`, leaving the field empty after a load.
- **User-data quoting**: `clone_url` and `clone_branch` are now quoted in the root user-data script to handle URLs containing special characters.
- **`submit_tasks` example**: `--count 0` is now rejected upfront rather than silently submitting zero tasks.


magniloquency:
**Removed: oscillation fix for VanillaScalingPolicy low-ratio drain branch**

Commit 74c12694 added a `current == 0` guard in `VanillaScalingPolicy._compute_desired_worker_count` to prevent spurious scale-up cycles when a manager has no connected workers. It was removed before merging to keep this PR focused; leaving a note here for whoever picks it up.

**The bug:** With two worker managers (e.g. OCI and EC2), OCI workers can complete all tasks before EC2 finishes booting. At that point there might be 1 task left and 2 OCI workers connected — ratio = 0.5, below `lower_task_ratio = 1`. The low-ratio drain branch computes `desired = max(1, ceil(1/10)) = 1` for EC2, even though EC2 has zero managed workers. EC2 gets a start command, boots, finds nothing to do, drains, and the cycle repeats.

**The fix:**

```python
elif task_ratio < self._lower_task_ratio:
-    desired = 0 if task_count == 0 else max(1, ceil(task_count / self._upper_task_ratio))
+    if task_count == 0 or current == 0:
+        desired = 0
+    else:
+        desired = max(1, ceil(task_count / self._upper_task_ratio))
```

The `max(1, ...)` floor in the drain branch is meant to prevent a manager that *already has workers* from fully emptying while tasks remain — not to bootstrap a new manager. When `current == 0` there is nothing to keep, so the floor should not apply.

docs/html_extra/launchpad/provisioner.js
```diff
@@ -130,130 +135,309 @@ function downloadText(filename, content) {
...
-      }
+  var root = {
+    object_storage_server: TOML.Section({
+      bind_address: `${proto}://0.0.0.0:${op}${wsSlash}`,
+    }),
+    scheduler: TOML.Section({
+      bind_address: `${proto}://0.0.0.0:${sp}${wsSlash}`,
+      object_storage_address: `${proto}://127.0.0.1:${op}${wsSlash}`,
+      advertised_object_storage_address: `${proto}://${ctx.publicIp}:${op}${wsSlash}`,
+      policy: TOML.Section(policySection),
```
e117649: The policy is emitted as a nested policy table on the scheduler section (policy: TOML.Section(policySection)), which TOML.stringify renders as [scheduler.policy]. But the Python scheduler reads policy_engine_type / policy_content as flat keys directly inside [scheduler]. SchedulerConfig.policy is a PolicyConfig ConfigClass (config/section/scheduler.py:98) that _toml_section_defaults (config/loading.py:39-53) flattens into the same [scheduler] namespace, registering only policy_engine_type/policy_content. A nested [scheduler.policy] sub-table arrives as a single key policy that maps to no dest and is silently dropped. The canonical config confirms these keys live directly under [scheduler] (config/scheduler.toml:61,66). Net effect: selecting Waterfall produces a config where the policy is never applied — the scheduler falls back to the default simple. Fix: emit the policy keys flat on the scheduler table (spread policy_engine_type, and policy_content for waterfall, directly into the scheduler object) and update configFromToml (lines 407-409) to read them from scheduler, not scheduler.policy.

docs/html_extra/launchpad/provisioner.js
```diff
@@ -130,130 +135,309 @@ function downloadText(filename, content) {
...
+    var orbDerivedCount = wm.capMode === "instances"
+      ? Math.max(0, wm.instanceCap || 0)
+      : Math.max(0, Math.floor((wm.budgetCap || 0) / (inst.price || 1)));
+    Object.assign(table, {
+      worker_scheduler_address: `${ctx.proto}://${ctx.privateIp}:${ctx.sp}${ctx.wsSlash}`,
+      object_storage_address: `${ctx.proto}://${ctx.privateIp}:${ctx.op}${ctx.wsSlash}`,
+      python_version: cfg.pythonVersion,
+      requirements_txt: TOML.multiline.basic(req + "\n"),
+      instance_type: wm.instanceType,
+      max_task_concurrency: orbDerivedCount,
```
e117649: orb_aws_ec2 writes the instance count into max_task_concurrency, halving actual instances and the quoted cost

orbDerivedCount (lines 151-153) is a count of instances (instanceCap, or floor(budgetCap / inst.price)). But the ORB worker manager treats max_task_concurrency as total worker slots and derives instances itself: max_instances = ceil(max_task_concurrency / vCPUs_per_instance) (worker_manager.py:196-198, workers_per_instance = _discover_vcpu_count(...)). So requesting 4 instances of a 2-vCPU t3.medium writes max_task_concurrency = 4, yielding ceil(4/2) = 2 instances — half what the user asked for, and the Cost Summary (which quotes count × inst.price) is also doubled vs reality. The oci_raw branch already does this right by multiplying by OCPUs (line 208). Mirror it: SCALER_INSTANCES exposes a vcpu field, so write max_task_concurrency: orbDerivedCount * (inst.vcpu || 1), and divide back in configFromToml (line 330) instead of reading it straight into instanceCap.

docs/html_extra/launchpad/provisioner.js
```diff
@@ -130,130 +135,309 @@ function downloadText(filename, content) {
...
+    Object.assign(table, {
+      worker_scheduler_address: `${ctx.proto}://$PUBLIC_IP:${ctx.sp}${ctx.wsSlash}`,
+      oci_region: wm.ociRegion || "us-ashburn-1",
+      compartment_id: wm.ociCompartmentId || "",
+      availability_domain: wm.ociAvailabilityDomain || "",
+      subnet_id: wm.ociSubnetId || "",
+      container_image: wm.ociContainerImage || "",
+      instance_shape: wm.ociShape || "CI.Standard.E4.Flex",
+      instance_ocpus: wm.ociOcpus || 4,
+      instance_memory_gb: wm.ociMemoryGb || 30,
```
e117649: oci_raw memory default (30 GB) no longer matches the UI default (8 GB)

This diff lowered the UI memory default to 8 GB everywhere in app.jsx (wm.ociMemoryGb || 8: cost calc line 525, stepper 590, breakdown 679/682, costEstimate 1543, summary 2904/2911), and the oci_raw create path (app.jsx:242) never seeds ociMemoryGb. But provisioner.js still falls back to 30: the generated TOML (line 207), the cost-per-instance calc (line 194), and configFromToml (line 363) all use || 30. So a user who never touches the Memory stepper sees an 8 GB cost estimate but downloads/launches a config provisioning 30 GB (~3.75× more memory and bill; in budget mode the derived instance count is also computed against the wrong value). Fix: change lines 194/207/363 to || 8, or (better) seed ociMemoryGb: 8 and ociOcpus: 4 in the oci_raw onChange so display, build, and load all share one default.

src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
```diff
@@ -30,6 +30,20 @@
...
 
 
+def _extract_git_url_and_branch(requirements_content: str) -> Optional[Tuple[str, str]]:
+    """Return (clone_url, branch) for the first git+ line in a requirements file, or None."""
+    for line in requirements_content.splitlines():
+        idx = line.find("git+")
+        if idx < 0:
+            continue
+        raw_url = line[idx + 4 :]
+        at_idx = raw_url.rfind("@")
```
e117649: _extract_git_url_and_branch mis-splits URLs containing @ (ssh user, auth tokens, #egg)

raw_url.rfind("@") picks the last @, but the pip @branch separator only appears after the URL path, never the userinfo @ in scheme://user@host. The result is interpolated straight into git clone -b {branch} {url} (lines 340-344), so common forms break silently: git+ssh://git@github.com/org/repo.git → url=ssh://git, branch=github.com/org/repo.git; git+https://TOKEN@github.com/org/repo.git → url=https://TOKEN; ...repo.git@main#egg=foo → branch=main#egg=foo (invalid ref); trailing env markers (; python_version<"3.9") leak into the branch. Fix: strip any #… fragment and trailing ; marker first, then only treat @ as a branch separator if it occurs after the last / of the path. packaging.requirements.Requirement is already imported (line 13) and could parse this cleanly. Please add unit tests for ssh/token/#egg/marker/plain-@branch cases.

docs/html_extra/launchpad/app.jsx
```diff
@@ -1808,9 +1888,37 @@ function App() {
...
+        const cfg = configFromToml(toml);
+        if (cfg.transport) setTransport(cfg.transport);
+        if (cfg.schedulerPort) setSchedPort(cfg.schedulerPort);
+        if (cfg.objectStoragePort) setObjPort(cfg.objectStoragePort);
+        if (cfg.pythonVersion) setPyVer(cfg.pythonVersion);
+        if (cfg.region) setRegion(cfg.region);
+        if (cfg.policy) setPolicy(cfg.policy);
+        if (cfg.workerManagers && cfg.workerManagers.length) {
+          setWorkerManagers(cfg.workerManagers);
+          wmCounterRef.current = cfg.workerManagers.length;
```
e117649: wmCounterRef set to array length after load can mint a colliding worker-manager id

After load, wmCounterRef.current = cfg.workerManagers.length, but loaded ids come from worker_manager_id (provisioner.js:323) and aren't guaranteed to be wm-1..wm-N. Load a config with ids wm-1 and wm-3 (easy: create 3, delete the middle, save) → counter becomes 2 → next "Add" mints wm-3, colliding. Since React keys and selectedWmId are keyed on wm.id, this breaks selection, removal (filters out both), update, and the emitted config. Fix: seed the counter from the max numeric suffix among loaded ids, not the length: wmCounterRef.current = Math.max(0, ...cfg.workerManagers.map(w => { const m = /^wm-(\d+)$/.exec(w.id); return m ? +m[1] : 0; })) (and derive _uid consistently).

docs/html_extra/launchpad/app.jsx
```diff
@@ -1808,9 +1888,37 @@ function App() {
...
+    reader.onload = (ev) => {
+      try {
+        const toml = parseConfigToml(ev.target.result);
+        const cfg = configFromToml(toml);
+        if (cfg.transport) setTransport(cfg.transport);
+        if (cfg.schedulerPort) setSchedPort(cfg.schedulerPort);
+        if (cfg.objectStoragePort) setObjPort(cfg.objectStoragePort);
+        if (cfg.pythonVersion) setPyVer(cfg.pythonVersion);
+        if (cfg.region) setRegion(cfg.region);
+        if (cfg.policy) setPolicy(cfg.policy);
```
e117649, line 1921: Load config.toml drops networkBackend (zmq) on round-trip

handleLoadConfig restores transport/ports/python/region/policy/workerManagers but never networkBackend, and configFromToml (provisioner.js:429-437) doesn't return it. Since the orb builder writes network_backend = "ymq" (and omits it for zmq, passing zmq via the SCALER_NETWORK_BACKEND env var instead, provisioner.js:549), a zmq config can't round-trip: loading leaves the UI at the ymq default and re-downloading silently switches to ymq. Fix: read network_backend off a worker_manager entry in configFromToml, return it, and add if (cfg.networkBackend) setNetBack(cfg.networkBackend); here.

docs/html_extra/launchpad/components.jsx
```diff
@@ -1627,7 +1627,7 @@ function SchedulerLogTerminal({ instanceId, region, credentials, isActive }) {
             DocumentName: "AWS-RunShellScript",
             Parameters: {
               commands: [
-                "tail -n 150 /var/log/scaler.log 2>/dev/null || echo '(log not yet available)'",
+                "cat /var/log/scaler.log 2>/dev/null || echo '(log not yet available)'",
```
e117649: cat (vs tail -n 150) over an unbounded log shows the oldest lines, not recent activity

Switching tail -n 150 /var/log/scaler.log to cat is a regression. SSM GetCommandInvocation.StandardOutputContent is truncated to roughly the first ~2,500 characters of output (the head, with a ----- Result truncated ---- marker), not the tail. Once scaler.log grows past that — almost immediately after startup — the terminal shows only startup output and never the recent activity the user wants, the opposite of what tail gave. Revert to a bounded tail (tail -n 500 or tail -c 20000). (setLines ~line 1675 also rebuilds the whole array each poll with no bound.)

src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
```diff
@@ -320,15 +334,60 @@ def _create_user_data(self) -> str:
 
             # User data runs as root so no sudo is needed.
             # set -e ensures any install failure aborts the script rather than launching a broken worker.
-            script += f"""set -e
+            git_info = _extract_git_url_and_branch(requirements_content)
+            if git_info is not None:
+                clone_url, clone_branch = git_info
+                clone_cmd = (
+                    f"git clone -b {clone_branch} --depth 1 {clone_url} /opt/scaler-src"
```
e117649: clone_url/clone_branch interpolated unquoted into the root user-data script

f"git clone -b {clone_branch} --depth 1 {clone_url} /opt/scaler-src" interpolates unsanitized values into a script that runs as root at boot. Any value with whitespace breaks the clone, and shell metacharacters in the requirements line execute as root. Even though the requirements file is operator-supplied, please harden: import shlex and git clone -b {shlex.quote(clone_branch)} --depth 1 {shlex.quote(clone_url)} ... (and the no-branch branch), plus parse the branch cleanly so markers/fragments don't leak in.

src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
```diff
@@ -30,6 +30,20 @@
 ORB_AWS_EC2_MAX_POLLING_ATTEMPTS = 60
 
 
+def _extract_git_url_and_branch(requirements_content: str) -> Optional[Tuple[str, str]]:
+    """Return (clone_url, branch) for the first git+ line in a requirements file, or None."""
+    for line in requirements_content.splitlines():
+        idx = line.find("git+")
```
e117649: line.find("git+") matches commented-out / inline-comment lines

line.find("git+") matches the substring anywhere, including # install via git+https://... and inline trailing comments. A commented requirement therefore flips the entire user-data into the heavy from-source C++ build path, and raw_url = line[idx+4:] carries trailing comment text into the clone command. _validate_requirements (lines 487-499) already strips/skips comment lines correctly — mirror that parsing here so comments can't select the from-source path.

docs/html_extra/launchpad/provisioner.js
```diff
@@ -130,130 +135,309 @@ function downloadText(filename, content) {
...
+      return Object.assign(base, {
+        instanceType: wm.instance_type || "t3.medium",
+        capMode: "instances",
+        instanceCap: wm.max_task_concurrency != null ? wm.max_task_concurrency : 4,
+        budgetCap: 10,
+        requirements: wm.requirements_txt || "opengris-scaler[all]",
+      });
+    }
+    if (wm.type === "aws_raw_ecs") {
+      return Object.assign(base, {
```
e117649, line 342: aws_raw_ecs requirements lost on config.toml round-trip

On load, the aws_raw_ecs branch never reads back the ECS requirements. Save writes them as ecs_python_requirements (provisioner.js:180) and the UI binds the textarea to wm.requirements (app.jsx:639), but this Object.assign omits requirements, so after a round-trip the textarea is empty and a re-download drops the line entirely. Add requirements: wm.ecs_python_requirements || "opengris-scaler[all]" (mirroring the orb branch at line 332).

examples/submit_tasks.py
```diff
@@ -8,58 +8,103 @@
...
             results_starmap = client.starmap(add, [(5, 5), (10, 10)])
             print(f"Results of starmap(add, [(5,5), (10,10)]): {list(results_starmap)}")
+
+            print(f"\nSubmitting {args.count} heavy CPU tasks (prime count up to {args.workload}) via .map()...")
+            batch_start = time.monotonic()
+            heavy_results = list(client.map(cpu_bound, [args.workload] * args.count))
+            batch_elapsed = time.monotonic() - batch_start
+            print(
+                f"Completed {len(heavy_results)} heavy tasks in {batch_elapsed:.1f}s "
+                f"({batch_elapsed / len(heavy_results) * 1000:.1f} ms/task avg)"
```
e117649, line 95: --count 0 crashes with ZeroDivisionError / empty max()

[args.workload] * args.count is empty for --count 0, so batch_elapsed / len(heavy_results) (line 92) raises ZeroDivisionError and max(... for r in mixed_results) (line 99) raises on the empty sequence. Either reject it up front (if args.count < 1: parser.error("--count must be >= 1")) or guard with default=0.0 / a length check.

docs/html_extra/launchpad/app.jsx
```diff
@@ -1428,6 +1493,16 @@ function App() {
...
   }, [theme]);
 
+  useEffect(() => {
+    if (!IS_DEV) return;
+    sessionStorage.setItem('launchpad-dev-aki', accessKeyId);
+    sessionStorage.setItem('launchpad-dev-sk', secretKey);
+    sessionStorage.setItem('launchpad-dev-oci-uid', ociUserId);
+    sessionStorage.setItem('launchpad-dev-oci-tid', ociTenancyId);
+    sessionStorage.setItem('launchpad-dev-oci-fp', ociFingerprint);
+    sessionStorage.setItem('launchpad-dev-oci-pk', ociPrivateKey);
```
e117649, line 1508: Dev mode persists the AWS secret key and full OCI private key to sessionStorage in cleartext (security, dev-gated)

Under ?dev, this effect writes secretKey (line 1499) and the OCI private key (line 1503) to sessionStorage, rehydrated on reload via the initializers (lines 1375-1381). It's gated behind the ?dev flag and sessionStorage clears on tab close, but cleartext credential persistence is an easy-to-leave-on footgun. Recommend persisting only non-secret identifiers (access key id, region, OCI user/tenancy/fingerprint), clearing on beforeunload, and surfacing a visible warning in the DEV badge.

e117649 (changes requested):

docs/html_extra/launchpad/provisioner.js
magniloquency: Fixed in bd4b6680. Spread the policy keys flat into the scheduler object via `Object.assign` and updated `configFromToml` to read `policy_engine_type`/`policy_content` directly from `scheduler` instead of `scheduler.policy`.

docs/html_extra/launchpad/provisioner.js
magniloquency: Fixed in 1c4e09e1. Now writes `max_task_concurrency: orbDerivedCount * inst.vcpu` (mirroring the `oci_raw` branch) and divides back by `vcpu` in `configFromToml` to recover the instance count on load. If the instance type isn't in `SCALER_INSTANCES`, both paths throw immediately rather than silently defaulting.

docs/html_extra/launchpad/provisioner.js
magniloquency: Fixed in 7127519a. Seeded `ociMemoryGb: 8` and `ociOcpus: 4` explicitly in the `oci_raw` onChange so there's one source of truth. Removed the `|| 30` / `|| 4` fallbacks from the build and load paths in `provisioner.js` — `configFromToml` now throws if either field is absent rather than silently provisioning the wrong capacity.

src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
magniloquency: Fixed in e6e3cac0. Rewrote `_extract_git_url_and_branch` to use `packaging.requirements.Requirement` for PEP 508 parsing (handles comment skipping and env marker stripping) and `urlsplit` to locate the `@branch` separator only in the path component, so `user@host` in the authority is never mistaken for a branch. Note: only the PEP 508 URL form (`name @ git+url`) is supported; bare legacy VCS lines are not. Added unit tests covering https with branch, ssh `user@host`, token auth, env markers, and commented-out lines.

docs/html_extra/launchpad/app.jsx
magniloquency: Fixed in 424ed385. Moved ID generation into the `setWorkerManagers` updater so it sees the live list and bumps the counter until the candidate ID is free — no dependency on the counter being in sync with loaded IDs. Removed the `wmCounterRef` reset on load entirely. Also added a unique-ID check to the pre-launch `checks` array so that duplicate IDs (e.g. from a manually edited ID field) block launch with a visible error rather than producing a broken config.

docs/html_extra/launchpad/app.jsx
magniloquency: Fixed in f2b8b666. `network_backend` is now always written explicitly (`"ymq"` or `"zmq"`) into every `orb_aws_ec2` worker manager TOML section — the Python config loader accepts both via `NetworkBackendType` enum name parsing. The scheduler env var is now set unconditionally too, and the `tcp_zmq` typo is fixed to `zmq`. `configFromToml` reads `network_backend` from the first worker manager entry and returns it; `handleLoadConfig` restores it via `setNetBack`.

docs/html_extra/launchpad/components.jsx
magniloquency: Fixed in ee0dfe32. Replaced `cat` with a byte-offset-based incremental fetch: each poll runs `tail -c +$((offset+1))` and advances the offset by the bytes actually received. On a new polling session the offset resets to 0, so the full log is read from the top in chunks determined by SSM's output limit. When a fetch returns content the loop fires again immediately for the next chunk; once output is empty we're caught up and the normal interval timer resumes. This also removes the unbounded `setLines` rebuild — new content is appended rather than replacing the full array.

src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
magniloquency: Fixed in ef22e431. Added `shlex.quote` to both `clone_url` and `clone_branch`. Note that the rewrite of `_extract_git_url_and_branch` in e6e3cac0 (comment #4) already ensures these values are structurally clean via `Requirement` + `urlsplit`, but quoting is the right practice for values interpolated into a root shell script regardless.

src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
magniloquency: Fixed as a side-effect of e6e3cac0 (comment #4). The rewrite of `_extract_git_url_and_branch` now uses `packaging.requirements.Requirement` to parse each line, which skips blank lines and lines starting with `#` or `-` before attempting to parse. Commented-out lines no longer reach the VCS URL check.

docs/html_extra/launchpad/provisioner.js
magniloquency: Fixed in 4a09602e. Added `requirements: wm.ecs_python_requirements || "opengris-scaler[all]"` to the `aws_raw_ecs` branch in `configFromToml`.

examples/submit_tasks.py
magniloquency: Fixed by rejecting `--count 0` upfront via `parser.error('--count must be >= 1')` immediately after `parse_args()`.

docs/html_extra/launchpad/app.jsx
magniloquency: I want to keep this. One of the main barriers to testing is needing to re-enter all fields on page refresh. This stores it in session storage, which is cleared on tab close. I agree that this is insecure, but it's an explicit opt-in, this only applies if you put `?dev` in your url. I'll understand if that's still untenable.

e117649: Ok that's fine, let's just note or warn somewhere though that sessionStorage unlike in-memory state gets written to disk by the browser's session-restore/crash-recovery, so the plaintext secret key and OCI private key can land in the browser profile on disk and outlive a reload.

src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
```diff
@@ -30,6 +32,36 @@
...
+            req = Requirement(line.partition(";")[0].rstrip())
+        except Exception:
+            continue
+        if not req.url or not req.url.startswith("git+"):
+            continue
+        parsed = urlsplit(req.url)
+        at_idx = parsed.path.find("@")
+        if at_idx >= 0:
+            branch = parsed.path[at_idx + 1 :]
+            url = urlunsplit(parsed._replace(scheme=parsed.scheme[4:], path=parsed.path[:at_idx]))
```
e117649: urlunsplit re-appends the fragment, so pkg @ git+https://host/repo.git@main#egg=pkg → git clone https://host/repo.git#egg=pkg, which fails at boot. Add fragment="" to the parsed._replace(...) in both return paths. (#egg was on the original list.)

docs/html_extra/launchpad/app.jsx
```diff
@@ -1808,9 +1896,37 @@ function App() {
...
+        const toml = parseConfigToml(ev.target.result);
+        const cfg = configFromToml(toml);
+        if (cfg.transport) setTransport(cfg.transport);
+        if (cfg.schedulerPort) setSchedPort(cfg.schedulerPort);
+        if (cfg.objectStoragePort) setObjPort(cfg.objectStoragePort);
+        if (cfg.pythonVersion) setPyVer(cfg.pythonVersion);
+        if (cfg.region) setRegion(cfg.region);
+        if (cfg.policy) setPolicy(cfg.policy);
+        if (cfg.networkBackend) setNetBack(cfg.networkBackend);
+        if (cfg.workerManagers && cfg.workerManagers.length) {
```
e117649, line 1923: Dropping wmCounterRef.current = cfg.workerManagers.length fixes id collisions, but _uid is used as the React key ([app.jsx:2869](https://0k713hn8h8c4109gkss8nciuqa58jiulkdacnmcadn2coe0lqtel.vscode-cdn.net/stable/7e7950df89d055b5a378379db9ee14290772148a/out/vs/workbench/contrib/webview/browser/pre/docs/html_extra/launchpad/app.jsx#L2869), 2898, 2917, 2928) and the do/while only de-dupes ids, not _uids. Load a config whose ids aren't wm-N (e.g. you renamed them alpha/beta) → they get _uid 1,2 while the counter stays at 1 → clicking Add mints wm-2 (free id) with _uid: 2, colliding with beta's _uid: 2. Result: duplicate-key warning and possible mis-render of the cost-breakdown rows. Fix: restore a counter sync on load (wmCounterRef.current = Math.max(wmCounterRef.current, cfg.workerManagers.length)) or give _uid its own monotonic counter.

docs/html_extra/launchpad/components.jsx
```diff
@@ -1619,75 +1620,77 @@ function SchedulerLogTerminal({ instanceId, region, credentials, isActive }) {
...
-                .split("\n")
-                .map((text) => ({ text, cls: "info" })),
-            );
-            break;
+            const output = inv.StandardOutputContent || "";
+            if (output) {
+              byteOffsetRef.current += output.length;
+              setLines((prev) => [
+                ...prev,
+                ...output.split("\n").map((text) => ({ text, cls: "info" })),
```
e117649: The byte-offset tail correctly fixes the head-truncation problem. Two new issues from the append+split: (a) tail output ends in \n, so output.split("\n") appends a phantom "" line every poll — the terminal fills with blank gaps; and a log line straddling a poll/SSM-truncation boundary renders as two split lines (never rejoined). Keep a pendingPartialRef: prepend it, split, pop the last (partial) element back into the ref, append only complete lines. (b) setLines(prev => [...prev, ...]) now grows unbounded over a long session (the old cat replaced the array) — cap it: return next.length > MAX_LINES ? next.slice(-MAX_LINES) : next.

e117649 (changes requested):

src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
magniloquency: Fixed — added `fragment=""` to `parsed._replace(...)` in both return paths.

docs/html_extra/launchpad/app.jsx
magniloquency: Fixed — added a separate `uidCounterRef` that only ever increments. `addWorkerManager` uses `++uidCounterRef.current` for `_uid`, and `handleLoadConfig` re-maps each loaded WM to a fresh UID from the same counter, so there's no way for loaded and newly-added WMs to share a `_uid` regardless of what ids the config contains.

docs/html_extra/launchpad/components.jsx
magniloquency: Fixed the partial-line/phantom-blank issue with a pendingPartialRef that holds whatever follows the last \n between polls, prepended to the next batch before splitting. WRT to capping the number of log lines. I want the full log to always be accessible.

docs/html_extra/launchpad/app.jsx
magniloquency: Added a warning in the credentials pane noting this.

docs/html_extra/launchpad/components.jsx
e117649: How long do we expect a tab to be open for? after about a day of running it would start to become janky, and potentially crash the tab since every line is rendered in DOM

If we expect long runs, or logs that approach 100k lines, there's a few options:

> 1. Virtualize the render. Keep all lines in the JS array but only mount the visible window (~100s) via windowing (react-window-style, or manual: render a slice from scroll position with top/bottom spacer divs). A plain array of strings holds millions of lines in a few hundred MB with a constant-size DOM — full retention, no crash.
> 
> 2. Keep the full log outside React + a Download button. Accumulate the raw text in a ref (not state, not DOM), render only the last N lines in the terminal, and add "Download full log" that dumps the accumulated text as a Blob. Simpler than virtualization; you keep everything, just don't display it all.
> 
> 3. Fetch the file on demand. The complete log already lives at /var/log/scaler.log on the instance — the live terminal can show a bounded tail, and "full logs" is best served by a one-shot fetch of the whole file (SSM → download, or stream to S3) rather than mirroring it live in the browser.

e117649 (approved):

PR #837, opened 2026-06-17
scaling: always emit setDesired to fix missed scale-down of in-flight workers

magniloquency:
## Summary

- Removes the `if effective == len(managed_worker_ids): return []` early-return from `VanillaScalingPolicy`, `CapabilityScalingPolicy`, and `WaterfallScalingPolicy`
- Policies now unconditionally emit a `setDesired` command so worker managers always receive the authoritative desired count, even when it matches the number of currently *connected* workers

## Motivation

The early-return compared the computed desired count against the number of **connected** workers, but workers that are launching are not yet connected. This created a window where a scale-down could be silently skipped.

**Reproduction scenario:**

1. Start a cluster with both an OCI and an EC2 worker manager configured.
2. Submit a few tasks — the scheduler tells both WMs to scale to 1 worker.
3. OCI workers come up faster (~seconds); EC2 workers take ~2 minutes to boot.
4. An OCI worker executes all tasks before any EC2 worker is online.
5. The client disconnects; the scheduler begins scaling down.
6. The scheduler correctly tells the OCI WM to scale to 0.
7. For the EC2 WM, `len(managed_worker_ids) == 0` (no connected workers yet) and `effective == 0` (desired is also 0), so `effective == len(managed_worker_ids)` is `True` — the early-return fires and **no scale-down message is sent**.
8. Moments later the EC2 workers come online. Because the EC2 WM's desired count was never updated, those workers stay alive indefinitely.

By always emitting the `setDesired` command the worker manager receives the correct target regardless of how many workers have connected so far.


gxuu (dismissed):

sharpener6:
@magniloquency @gxugxu

We should not do any early returns and just populate desired number of concurrency to worker manager unconditionally for a heartbeat time.

magniloquency:
> We should not do any early returns and just populate desired number of concurrency to worker manager unconditionally for a heartbeat time.

@sharpener6, indeed that is what this PR does. Policies would try to not send "set desired" messages if they thought the worker manager already had the correct value. Unfortunately there was a bug in the check that meant it would sometimes not be sent when it should be. We investigated fixing the check, but it turned out to be difficult, so the easiest solution is just to send it every time, which is what we are doing here.

Unfortunately that did not fix all of our problems. I found that in multi-manager scenarios, scaling policies could produce oscillations in the worker counts. I think this could have two causes, which I have not yet directly confirmed:

- Policies don't understand worker managers that launch N workers at a time (e.g. orb ec2, oci, ...)
- Policies have a bootstrapping process that always tries to provision workers, but there is no guard against it scaling back down later and re-triggering the bootstrap

sharpener6:
> @sharpener6, indeed that is what this PR does. Policies would try to not send "set desired" messages if they thought the worker manager already had the correct value. Unfortunately there was a bug in the check that meant it would sometimes not be sent when it should be. We investigated fixing the check, but it turned out to be difficult, so the easiest solution is just to send it every time, which is what we are doing here.

I don't think it's good idea that policy to be smart to decide do this:

```
not send "set desired" message if they thought the worker manager already had the correct value
```

scheduler blindly send "set desired" message is THE correct way

make connection one direction only, means only scheduler -> worker manager, @gxugxu why scheduler's policy engine need to know how many workers is up now?

magniloquency:
> scheduler blindly send "set desired" message is THE correct way

@sharpener6, I agree. This PR implements that and is ready, feel free to review / merge whenever you want

gxugxu:
@sharpener6,
>  @gxugxu why scheduler's policy engine need to know how many workers is up now?

It doesn't _need_ to know. I want policy engine knows how many workers is up now so that policy engine won't send `setDesired` when the worker count the same as what we asked for (or reached the maximum). Apparently that turned out wrong. This PR remediates this.

sharpener6 (approved):

PR #839, opened 2026-06-17
Better logging

gxugxu:
# READY FOR REVIEW.

This PR changes the original logging method to named loggers. This way, we know which file is printing a particular log message.


src/cpp/scaler/logging/logging.h
```diff
@@ -1,5 +1,8 @@
 #pragma once
 
+#include <uv.h>
```
magniloquency, line 3: I think project style says this should be below std lib imports?

src/scaler/worker_manager_adapter/aws_raw/ecs.py
```diff
@@ -16,6 +16,10 @@
 if TYPE_CHECKING:
     from scaler.protocol.capnp import WorkerManagerCommand
 
+logger = logging.getLogger(__name__)
+
+_WorkerGroupID = bytes
```
magniloquency: Unintended change?

src/scaler/worker_manager_adapter/symphony/worker_manager.py
```diff
@@ -15,6 +15,7 @@
 
 if TYPE_CHECKING:
     from scaler.protocol.capnp import WorkerManagerCommand
+logger = logging.getLogger(__name__)
```
magniloquency, line 19: I think we should have a space between these lines. Same for `aws_hpc/worker_manager.py`.

```suggestion
from scaler.protocol.capnp import WorkerManagerCommand

logger = logging.getLogger(__name__)
```

src/scaler/worker_manager_adapter/aws_hpc/worker_manager.py
```diff
@@ -84,13 +84,16 @@ async def terminate(self) -> None:
         self._units.clear()
 
 
+logger = logging.getLogger(__name__)
```
magniloquency: This should be at the top of the file

src/scaler/utility/graph/topological_sorter.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import sys
 
+logger = logging.getLogger(__name__)
 # python-graphblas's native Matrix.__del__ has been observed to corrupt the heap on Windows
```
magniloquency, line 6:
```suggestion
logger = logging.getLogger(__name__)

# python-graphblas's native Matrix.__del__ has been observed to corrupt the heap on Windows
```

magniloquency (changes requested):

src/scaler/utility/logging/scoped_logger.py
```diff
@@ -3,6 +3,8 @@
 import time
 from typing import Optional
 
+logger = logging.getLogger(__name__)
```
magniloquency: I think this sets the logger name for the logger used by the `ScopedLogger` class, which means the wrong module name will be used wherever `ScopedLogger` is used

magniloquency (changes requested):

src/cpp/scaler/logging/logging.h
gxugxu: this is done automatically via `clang-format`. Otherwise pipeline will complain.

magniloquency (dismissed):

magniloquency (approved):

e117649 (approved):

PR #840, opened 2026-06-17
Object Storage Address should have pretty print name

gxugxu:
# READY FOR REVIEW.


src/cpp/scaler/protocol/pymod/bootstrap.cpp
```diff
@@ -192,6 +192,11 @@ PyObject* py_capnp_struct_to_bytes(PyObject* self, PyObject* /*unused*/)
     return ::scaler::protocol::pymod::capnp_struct_to_bytes(self).take();
 }
 
+PyObject* py_capnp_struct_repr(PyObject* self, PyObject* /*unused*/)
```
magniloquency, line 195: This needs to be registered in the class definition so that it's wired up

magniloquency (changes requested):

magniloquency (approved):

PR #841, opened 2026-06-17
Actor Model Initial Implementation

gxugxu:
# READY FOR REVIEW.

This PR only implements the basic Actor Model Layer API, while the RPC Layer API is left until this is merged. This PR enables the ability of bi-directional arbitrary message passing between client and worker (actors).

### DIFF
--- a/src/protocol/common.capnp
+++ b/src/protocol/common.capnp
@@ -70,3 +70,39 @@ struct ObjectStorageAddress {
     port   @1 :UInt16;
     scheme @2 :Text;
 }
+
+enum ActorState {
+    pending @0;     # accepted by the scheduler, no worker assigned yet
+    creating @1;    # actor process spawning / constructor running on the worker
+    alive @2;       # constructor finished, serving messages
+    stopping @3;    # graceful destroy in progress on the worker
+    dead @4;        # terminal; ActorStateUpdate.deathInfo carries the reason
+}
+
+struct ActorPayload {
+    type @0 :ActorPayloadType;
+    data @1 :Data;     # the serialized value itself (inline) or its object storage id (objectID)
+
+    enum ActorPayloadType {
+        inline @0;
+        objectID @1;
+    }
+}
+
+struct ActorArguments {
+    positional @0 :List(ActorPayload);
+    keyword @1 :List(KeywordArgument);
+
+    struct KeywordArgument {
+        name @0 :Text;
+        value @1 :ActorPayload;
+    }
+}
+
+struct ActorError {
+    errorType @0 :Text;     # fully qualified exception class name, e.g. "builtins.ValueError"
+    message @1 :Text;       # str(exception)
+    # Re-raisable detail (formatted traceback, pickled exception) is deferred to the RPC layer,
+    # which is where a failure is actually surfaced to user code; the Actor Layer only carries
+    # the human-readable, language-neutral type + message.
+}
--- a/src/protocol/message.capnp
+++ b/src/protocol/message.capnp
@@ -67,6 +67,8 @@ struct GraphTask {
 struct ClientHeartbeat {
     resource @0 :Status.Resource;
     latencyUS @1 :UInt32;
+    actorIds @2 :List(Data);   # complete set of actors this client claims to own; always the
+                               # full set, never a diff
 }
 
 struct ClientHeartbeatEcho {
@@ -83,6 +85,7 @@ struct WorkerHeartbeat {
     processors @6 :List(Status.ProcessorStatus);
     capabilities @7 :List(CommonType.TaskCapability);
     workerManagerID @8 :Data;
+    actors @9 :List(Status.ActorHostStatus);
 }
 
 struct WorkerHeartbeatEcho {
@@ -200,6 +203,76 @@ struct InformationResponse {
     response @0 :Data;
 }
 
+# Convention for every actor message:
+#   actorId: client-generated, globally unique, never reused across incarnations
+#   source:  identity of the owning client; sender on client->worker messages, destination on
+#            worker->client messages
+
+# Declares "this actor should exist". Idempotent per actorId: re-sending is answered with the
+# actor's current ActorStateUpdate, never a duplicate actor.
+struct ActorCreate {
+    actorId @0 :Data;
+    source @1 :Data;
+    classObjectId @2 :Data;    # object id of the serialized actor class
+    constructorArguments @3 :CommonType.ActorArguments;
+    capabilities @4 :List(CommonType.TaskCapability);
+}
+
+# Declares "this actor should not exist". Idempotent and escalatable (graceful then kill).
+struct ActorDestroy {
+    actorId @0 :Data;
+    source @1 :Data;
+    mode @2 :Mode;
+
+    enum Mode {
+        graceful @0;   # let the actor wind down cooperatively, then exit
+        kill @1;       # terminate the actor process immediately
+    }
+}
+
+# Declares the full current state of one actor; sent on every transition and re-sendable at
+# any time. Receivers treat it as absolute truth, so it is safe to duplicate and safe to lose
+# (heartbeat reconciliation re-derives it).
+struct ActorStateUpdate {
+    actorId @0 :Data;
+    source @1 :Data;
+    workerId @2 :Data;                  # assigned worker; empty while pending
+    state @3 :CommonType.ActorState;
+    deathInfo @4 :DeathInfo;            # only meaningful when state == dead
+
+    struct DeathInfo {
+        reason @0 :Reason;
+        error @1 :CommonType.ActorError;   # populated for constructorFailed/actorCrashed
+
+        enum Reason {
+            destroyed @0;           # client-requested destroy completed
+            constructorFailed @1;   # actor constructor raised
+            actorCrashed @2;        # actor process exited unexpectedly
+            workerDied @3;          # hosting worker lost (heartbeat timeout)
+            clientDisconnected @4;  # owning client lost; scheduler reclaimed the actor
+            placementFailed @5;     # no worker satisfies the requested capabilities
+            unknownActor @6;        # actorId unknown to the scheduler
+        }
+    }
+}
+
+# The actor data plane. payload is opaque to the scheduler and the worker agent; it is routed
+# by actorId (client->worker) and source (worker->client).
+struct ActorMessage {
+    actorId @0 :Data;
+    source @1 :Data;
+    payload @2 :Data;
+}
+
+struct StateActor {
+    actorId @0 :Data;
+    source @1 :Data;
+    workerId @2 :Data;
+    className @3 :Data;
+    state @4 :CommonType.ActorState;
+    capabilities @5 :List(CommonType.TaskCapability);
+}
+
 struct Message {
     union {
         task @0 :Task;
@@ -240,5 +313,11 @@ struct Message {
         workerManagerHeartbeat @25 :WorkerManagerHeartbeat;
         workerManagerHeartbeatEcho @26 :WorkerManagerHeartbeatEcho;
         workerManagerCommand @27 :WorkerManagerCommand;
+
+        actorCreate @28 :ActorCreate;
+        actorDestroy @29 :ActorDestroy;
+        actorStateUpdate @30 :ActorStateUpdate;
+        actorMessage @31 :ActorMessage;
+        stateActor @32 :StateActor;
     }
 }
--- a/src/protocol/status.capnp
+++ b/src/protocol/status.capnp
@@ -3,6 +3,8 @@
 using Cxx = import "/capnp/c++.capnp";
 $Cxx.namespace("scaler::protocol");
 
+using CommonType = import "common.capnp";
+
 struct Resource {
     cpu @0 :UInt16;   # 99.2% will be represented as 992 as integer
     rss @1 :UInt64;  # 32bit is capped to 4GB, so use 64bit to represent
@@ -85,4 +87,15 @@ struct BinderStatus {
         client @0 :Text;
         number @1 :UInt32;
     }
+}
+
+# Declared by the worker in every heartbeat, one entry per hosted actor.
+struct ActorHostStatus {
+    actorId @0 :Data;
+    source @1 :Data;                   # owning client; lets a restarted scheduler rebuild
+                                       # its actor routing table from heartbeats alone
+    state @2 :CommonType.ActorState;   # only the hosted states occur here: creating | alive |
+                                       # stopping (pending has no worker yet, dead is reported
+                                       # via ActorStateUpdate and removed from heartbeats)
+    processorPid @3 :UInt32;           # joins with ProcessorStatus for cpu/rss
 }
--- a/src/scaler/client/agent/actor_manager.py
+++ b/src/scaler/client/agent/actor_manager.py
@@ -33,7 +33,7 @@ class _Lifecycle:
 
 
 class _ActorRecord:
-    def __init__(self):
+    def __init__(self) -> None:
         # serializes lifecycle transitions and carries the wait/notify protocol; plain reads
         # of `lifecycle` never take it
         self.condition = threading.Condition()
@@ -50,7 +50,7 @@ class ClientActorManager(ActorManager):
     queue, and blocking waits go through the record's condition variable instead of polling.
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
         self._dict_lock = threading.Lock()
         self._actor_id_to_record: Dict[ActorID, _ActorRecord] = dict()
         # dead actor ids in death order, used to evict the oldest dead records once the retained
@@ -62,13 +62,13 @@ class ClientActorManager(ActorManager):
 
         self._connector_external: Optional[AsyncConnector] = None
 
-    def register(self, connector_external: AsyncConnector):
+    def register(self, connector_external: AsyncConnector) -> None:
         self._connector_external = connector_external
 
-    def track_actor(self, actor_id: ActorID):
+    def track_actor(self, actor_id: ActorID) -> None:
         self.__ensure_record(actor_id)
 
-    def untrack_actor(self, actor_id: ActorID):
+    def untrack_actor(self, actor_id: ActorID) -> None:
         """Drops a record again, e.g. when a create could not be sent after track_actor()."""
         with self._dict_lock:
             self._actor_id_to_record.pop(actor_id, None)
@@ -125,17 +125,17 @@ class ClientActorManager(ActorManager):
 
         return item
 
-    async def on_create_actor(self, actor_create: ActorCreate):
+    async def on_create_actor(self, actor_create: ActorCreate) -> None:
         self.track_actor(ActorID(bytes(actor_create.actorId)))
         await self._connector_external.send(actor_create)
 
-    async def on_destroy_actor(self, actor_destroy: ActorDestroy):
+    async def on_destroy_actor(self, actor_destroy: ActorDestroy) -> None:
         await self._connector_external.send(actor_destroy)
 
-    async def on_send_actor_message(self, actor_message: ActorMessage):
+    async def on_send_actor_message(self, actor_message: ActorMessage) -> None:
         await self._connector_external.send(actor_message)
 
-    async def on_actor_message(self, actor_message: ActorMessage):
+    async def on_actor_message(self, actor_message: ActorMessage) -> None:
         actor_id = ActorID(bytes(actor_message.actorId))
 
         record = self.__get_record(actor_id)
@@ -147,7 +147,7 @@ class ClientActorManager(ActorManager):
         # payload is a stable bytes object we can hand to the inbox without re-copying
         record.inbox.put(actor_message.payload)
 
-    async def on_actor_state_update(self, actor_state_update: ActorStateUpdate):
+    async def on_actor_state_update(self, actor_state_update: ActorStateUpdate) -> None:
         actor_id = ActorID(bytes(actor_state_update.actorId))
         record = self.__get_record(actor_id)
         if record is None:
@@ -166,7 +166,7 @@ class ClientActorManager(ActorManager):
 
     def set_all_actors_dead(
         self, reason: ActorStateUpdate.DeathInfo.Reason = ActorStateUpdate.DeathInfo.Reason.clientDisconnected
-    ):
+    ) -> None:
         with self._dict_lock:
             # latch inside the same locked section that snapshots the records, so a record added
             # by __ensure_record after this point sees the flag and is born dead instead of being
@@ -183,7 +183,7 @@ class ClientActorManager(ActorManager):
         record: _ActorRecord,
         new_state: ActorState,
         death_reason: Optional[ActorStateUpdate.DeathInfo.Reason],
-    ):
+    ) -> None:
         with record.condition:
             current = record.lifecycle
 
@@ -205,7 +205,7 @@ class ClientActorManager(ActorManager):
             # record lock is held would invert ordering); first death only, since dead is absorbing
             self.__retain_dead(actor_id)
 
-    def __retain_dead(self, actor_id: ActorID):
+    def __retain_dead(self, actor_id: ActorID) -> None:
         with self._dict_lock:
             self._dead_actor_ids[actor_id] = None
             while len(self._dead_actor_ids) > _MAX_RETAINED_DEAD_RECORDS:
--- a/src/scaler/client/agent/actor_manager.py
+++ b/src/scaler/client/agent/actor_manager.py
@@ -1,0 +1,229 @@
+import collections
+import dataclasses
+import logging
+import queue
+import threading
+from typing import Dict, Optional
+
+from scaler.client.agent.mixins import ActorManager
+from scaler.io.mixins import AsyncConnector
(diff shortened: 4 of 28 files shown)


src/protocol/message.capnp
```diff
@@ -200,6 +203,76 @@ struct InformationResponse {
...
+            unknownActor @6;        # actorId unknown to the scheduler
+        }
+    }
+}
+
+# The actor data plane. payload is opaque to the scheduler and the worker agent; it is routed
+# by actorId (client->worker) and source (worker->client).
+struct ActorMessage {
+    actorId @0 :Data;
+    source @1 :Data;
```
magniloquency, line 263: The direction of the message feels kind of ambiguous, can we renamed these `to` and `from`? The actual shape of the message doesn't change, but we can know where the message is going.

src/protocol/message.capnp
```diff
@@ -200,6 +203,76 @@ struct InformationResponse {
...
+
+# Declares the full current state of one actor; sent on every transition and re-sendable at
+# any time. Receivers treat it as absolute truth, so it is safe to duplicate and safe to lose
+# (heartbeat reconciliation re-derives it).
+struct ActorStateUpdate {
+    actorId @0 :Data;
+    source @1 :Data;
+    workerId @2 :Data;                  # assigned worker; empty while pending
+    state @3 :CommonType.ActorState;
+    deathInfo @4 :DeathInfo;            # only meaningful when state == dead
```
magniloquency, line 241: Does capnp support tagged union? This feels like it should be a tagged union?

src/protocol/message.capnp
```diff
@@ -200,6 +203,76 @@ struct InformationResponse {
...
+# Convention for every actor message:
+#   actorId: client-generated, globally unique, never reused across incarnations
+#   source:  identity of the owning client; sender on client->worker messages, destination on
+#            worker->client messages
+
+# Declares "this actor should exist". Idempotent per actorId: re-sending is answered with the
+# actor's current ActorStateUpdate, never a duplicate actor.
+struct ActorCreate {
+    actorId @0 :Data;
+    source @1 :Data;
```
magniloquency, line 215: `source` feels a bit ambiguous, can we call it `client` or `owner`?

src/protocol/common.capnp
```diff
@@ -70,3 +70,39 @@ struct ObjectStorageAddress {
     port   @1 :UInt16;
     scheme @2 :Text;
 }
+
+enum ActorState {
```
magniloquency, line 74: The word "state" is kind of ambiguous, perhaps ` ActorLifecycle` would be clearer?

src/protocol/status.capnp
```diff
@@ -86,3 +88,12 @@ struct BinderStatus {
         number @1 :UInt32;
     }
 }
+
+# Declared by the worker in every heartbeat, one entry per hosted actor.
+struct ActorHostStatus {
+    actorId @0 :Data;
+    source @1 :Data;                   # owning client; lets a restarted scheduler rebuild
+                                       # its actor routing table from heartbeats alone
+    state @2 :CommonType.ActorState;
```
magniloquency: In this situation, not every actor state is possible right?

src/scaler/client/agent/actor_manager.py
```diff
@@ -0,0 +1,219 @@
...
+    def register(self, connector_external: AsyncConnector):
+        self._connector_external = connector_external
+
+    def track_actor(self, actor_id: ActorID):
+        self.__ensure_record(actor_id)
+
+    def get_actor_state(self, actor_id: ActorID) -> ActorState:
+        record = self.__get_record(actor_id)
+        if record is None:
+            return ActorState.dead
```
magniloquency: We should push this up to the caller, let's return `None` if the actor isn't found, that could be meaningfully different to `dead` in some contexts?

src/scaler/client/agent/actor_manager.py
```diff
@@ -0,0 +1,219 @@
...
+        return record.lifecycle.state
+
+    def get_actor_death_reason(self, actor_id: ActorID) -> Optional[ActorStateUpdate.DeathInfo.Reason]:
+        record = self.__get_record(actor_id)
+        if record is None:
+            return None
+        return record.lifecycle.death_reason
```
magniloquency: Could we replace these with a single `get_actor_record()` method?

src/scaler/client/agent/actor_manager.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def receive_actor_message(self, actor_id: ActorID, timeout: Optional[float] = None) -> bytes:
+        """Blocks the calling (user) thread until the actor pushes a payload.
+
+        Payloads that arrived before the actor died remain consumable; once drained, a dead
+        actor's inbox always raises ActorDeadError.
+        """
+        record = self.__get_record(actor_id)
+        if record is None:
+            raise ActorDeadError(f"unknown actor {actor_id!r}")
```
magniloquency: We should have a separate exception class for this case.

src/scaler/client/agent/actor_manager.py
```diff
@@ -0,0 +1,219 @@
...
+
+    async def on_actor_message(self, actor_message: ActorMessage):
+        actor_id = ActorID(bytes(actor_message.actorId))
+
+        record = self.__get_record(actor_id)
+        if record is None:
+            logging.warning(f"{self.__class__.__name__}: dropping message for unknown actor {actor_id!r}")
+            return
+
+        record.inbox.put(bytes(actor_message.payload))
```
magniloquency: zero-copy?

src/scaler/client/agent/actor_manager.py
```diff
@@ -0,0 +1,219 @@
...
+    def set_all_actors_dead(self):
+        with self._dict_lock:
+            # latch inside the same locked section that snapshots the records, so a record added
+            # by __ensure_record after this point sees the flag and is born dead instead of being
+            # missed by the snapshot below
+            self._no_new_actors = True
+            items = list(self._actor_id_to_record.items())
+
+        for actor_id, record in items:
+            self.__transition(actor_id, record, ActorState.dead, ActorStateUpdate.DeathInfo.Reason.clientDisconnected)
```
magniloquency: Should the reason be a parameter?

src/scaler/client/client.py
```diff
@@ -484,6 +496,68 @@ def send_object(self, obj: Any, name: Optional[str] = None) -> ObjectReference:
...
+            source=self._identity,
+            classObjectId=class_object_id,
+            constructorArguments=ActorArguments(positional=positional_arguments, keyword=keyword_arguments),
+            capabilities=[],
+        )
+
+        self._actor_manager.track_actor(actor_id)
+
+        self._object_buffer.commit_send_objects()
+        self._connector_agent.send(actor_create)
```
magniloquency: If one of these throws an exc, is the actor untracked from the manager?

src/scaler/scheduler/controllers/actor_controller.py
```diff
@@ -0,0 +1,232 @@
...
+        self._binder = binder
+        self._worker_controller = worker_controller
+
+    def is_actor_worker(self, worker_id: WorkerID) -> bool:
+        return self._worker_to_actor_ids.has_key(worker_id)
+
+    async def on_actor_create(self, client_id: ClientID, actor_create: ActorCreate):
+        actor_id = ActorID(bytes(actor_create.actorId))
+
+        if bytes(actor_create.source) != bytes(client_id):
```
magniloquency: Can we compare without converting to bytes?

```suggestion
        if actor_create.source != client_id:
```

src/scaler/worker/agent/processor/processor.py
```diff
@@ -282,6 +310,132 @@ def __get_required_object_ids_for_task(task: Task) -> List[ObjectID]:
         ]
         return object_ids
 
+    def __on_received_actor_create(self, actor_create: ActorCreate):
+        if self._actor_instance is not None:
+            logging.error(f"Processor[{self.pid}]: already hosts an actor, ignoring ActorCreate")
+            return
```
magniloquency, line 329: I feel like we should communicate something back to the scheduler in this case?

src/scaler/worker/agent/processor/processor.py
```diff
@@ -282,6 +310,132 @@ def __get_required_object_ids_for_task(task: Task) -> List[ObjectID]:
...
+
+        payload = bytes(actor_message.payload)
+
+        receive_method = getattr(self._actor_instance, "__receive__", None)
+        if receive_method is None:
+            logging.error(
+                f"Processor[{self.pid}]: actor {type(self._actor_instance).__name__} has no __receive__ "
+                "method, dropping message"
+            )
+            return
```
magniloquency, line 401: Workers often run on remote machines (e.g. orb ec2, ecs, oci, etc.) I feel like this is a message we should surface to the scheduler maybe?

src/scaler/worker/agent/processor/processor.py
```diff
@@ -282,6 +310,132 @@ def __get_required_object_ids_for_task(task: Task) -> List[ObjectID]:
...
+
+        if reply is None:
+            return
+
+        if not isinstance(reply, (bytes, bytearray)):
+            logging.error(
+                f"Processor[{self.pid}]: __receive__ returned {type(reply).__name__}, expected bytes or "
+                "None, dropping reply"
+            )
+            return
```
magniloquency: Other types of objects can be successfully turned into bytes (e.g. anything implementing the buffer interface) it's more robust to call `bytes()` inside a `try ... catch`

src/scaler/worker/agent/processor/processor.py
```diff
@@ -282,6 +310,132 @@ def __get_required_object_ids_for_task(task: Task) -> List[ObjectID]:
...
+            ActorState.dead,
+            death_info=ActorStateUpdate.DeathInfo(reason=ActorStateUpdate.DeathInfo.Reason.destroyed),
+        )
+
+        # exiting the process is what actually releases the actor's resources; the agent
+        # detects the exit and starts a fresh processor
+        self._stop_requested = True
+
+    def __send_actor_state_update(
+        self, actor_create: ActorCreate, state: ActorState, death_info: Optional["ActorStateUpdate.DeathInfo"] = None
```
magniloquency: Why is the type a string? I think if you import annotations from `__future__` or use `TYPE_CHECKING` you should be able to put the actual type.

src/scaler/worker/agent/actor_manager.py
```diff
@@ -0,0 +1,285 @@
...
+    async def on_actor_state_update(self, host_identity: bytes, actor_state_update: ActorStateUpdate):
+        designation = self._designation
+        actor_id = ActorID(bytes(actor_state_update.actorId))
+        state = ActorState(actor_state_update.state.value)
+
+        if designation is not None and designation.actor_id == actor_id:
+            if state == ActorState.alive:
+                designation.alive = True
+            elif state == ActorState.dead:
+                self._designation = None  # the designation is released; heartbeats rejoin the pool
```
magniloquency: The representation of state seems kind of strange to me, is it possible for `designation.alive` to be `False`?

src/scaler/worker/agent/actor_manager.py
```diff
@@ -0,0 +1,285 @@
...
+
+        if designation.destroy_requested:
+            reason = ActorStateUpdate.DeathInfo.Reason.destroyed
+            detail = ""
+        else:
+            reason = ActorStateUpdate.DeathInfo.Reason.actorCrashed
+            detail = "actor processor exited unexpectedly"
+            logging.warning(f"{self.__class__.__name__}: actor {designation.actor_id!r}: {detail}")
+
+        await self.__send_dead(designation.actor_id, designation.source, reason, detail)
```
magniloquency: Can we make the happy path fall through?

src/scaler/worker/agent/actor_manager.py
```diff
@@ -0,0 +1,285 @@
...
+            )
+            return
+
+        if self._processor_manager.current_processor_id() is not None:
+            await self.__designate(actor_create)
+            return
+
+        # the processor initializes shortly after the worker starts; an ActorCreate can only
+        # beat it by a tiny window, so wait for it off the message loop
+        task = asyncio.get_event_loop().create_task(self.__designate_when_processor_ready(actor_create.to_bytes()))
```
magniloquency: this function is deprecated: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop

src/scaler/worker/agent/actor_manager.py
```diff
@@ -0,0 +1,285 @@
...
+        # manager detects the exit and starts a fresh processor
+        await self._binder_internal.send(designation.processor_id, actor_destroy)
+
+    def __forget_pending_create(self, actor_id: ActorID, task: asyncio.Task):
+        # only drop the entry if it is still this task: a destroy may have already removed it (and
+        # scheduled a fresh one is impossible here since actor ids are never reused)
+        if self._pending_creates.get(actor_id) is task:
+            del self._pending_creates[actor_id]
+
+    async def __designate_when_processor_ready(self, actor_create_payload: bytes):
```
magniloquency: Why not?


```suggestion
    async def __designate_when_processor_ready(self, actor_create_payload: ActorCreate):
```

magniloquency (changes requested):

src/protocol/message.capnp
gxugxu: This is used extensively (e.g. `Task.source`). So I think `source` is the right name.

src/protocol/common.capnp
gxugxu: This is the same as `WorkerState`, and they represent the same semantics.

src/protocol/message.capnp
gxugxu: This is used extensively (e.g. Task.source). So I think source is the right name.

src/protocol/status.capnp
gxugxu: added a comment

src/scaler/client/agent/actor_manager.py
gxugxu: done

src/scaler/client/agent/actor_manager.py
gxugxu: done

src/scaler/client/agent/actor_manager.py
gxugxu: Storing payload directly now.

src/scaler/client/agent/actor_manager.py
gxugxu: done

src/scaler/client/client.py
gxugxu: yes, now it does

src/scaler/worker/agent/processor/processor.py
gxugxu: sends `dead(UnknownActor)` now

src/scaler/worker/agent/processor/processor.py
gxugxu: This will be updated to send perhaps something like dead(IllegalActor).

src/scaler/worker/agent/actor_manager.py
gxugxu: done

src/scaler/worker/agent/actor_manager.py
gxugxu: done

src/scaler/worker/agent/actor_manager.py
gxugxu: yes

src/scaler/worker/agent/processor/processor.py
gxugxu: I re-read the code, this two errors should never happen. This should be removed once we polished the actor model. For now, it's imposed as a constraint to note other people developing the code.

src/scaler/worker/agent/processor/processor.py
```diff
@@ -282,6 +310,149 @@ def __get_required_object_ids_for_task(task: Task) -> List[ObjectID]:
...
+            logging.exception(f"Processor[{self.pid}]: actor raised while handling a message:")
+            return
+
+        if reply is None:
+            return
+
+        # accept anything that exposes the buffer protocol (bytes, bytearray, memoryview, numpy
+        # arrays, ...), not just bytes; bytes() raises TypeError for non-buffer objects
+        try:
+            payload = bytes(reply)
```
e117649: Should be `payload = bytes(memoryview(reply))` otherwise an actor whose `__receive__` returns a non-bytes-like value is silently delivered as a garbage payload instead of being rejected, e.g. `bytes(5) -> b'\x00\x00\x00\x00\x00'`

src/protocol/message.capnp
gxugxu: I don't really want to make it an unnamed union. Other messages typically uses enum

src/scaler/worker/agent/processor/processor.py
gxugxu: good catch!

e117649 (dismissed):

src/scaler/worker/agent/processor/processor.py
magniloquency: works for me

src/scaler/scheduler/controllers/actor_controller.py
```diff
@@ -0,0 +1,232 @@
...
+            return
+
+        actor_ids = self._client_to_actor_ids.get_values(client_id).copy()
+        logging.info(f"{self.__class__.__name__}: destroying {len(actor_ids)} actor(s) of client {client_id!r}")
+        for actor_id in actor_ids:
+            worker_id = self._worker_to_actor_ids.get_key(actor_id)
+            self.__forget_actor(actor_id)
+            await self._binder.send(
+                worker_id, ActorDestroy(actorId=actor_id, source=client_id, mode=ActorDestroy.Mode.kill)
+            )
```
magniloquency, line 199: There's potentially a subtle bug here, because the loop has an `await` it's possible for `self.actor_ids` to be mutated while we're iterating. We can guard by copying the list and guarding the `.get_key()`:


```suggestion
        for actor_id in actor_ids[:]:
            try:
                worker_id = self._worker_to_actor_ids.get_key(actor_id)
            except ValueError:
                continue # dict may have been modified during iteration
            self.__forget_actor(actor_id)
            await self._binder.send(
                worker_id, ActorDestroy(actorId=actor_id, source=client_id, mode=ActorDestroy.Mode.kill)
            )
```

Or we could add a lock

src/scaler/scheduler/controllers/actor_controller.py
```diff
@@ -0,0 +1,232 @@
...
+    async def on_worker_disconnect(self, worker_id: WorkerID):
+        if worker_id not in self._worker_to_actor_ids.keys():
+            return
+
+        actor_ids = self._worker_to_actor_ids.get_values(worker_id).copy()
+        logging.info(f"{self.__class__.__name__}: {len(actor_ids)} actor(s) died with worker {worker_id!r}")
+        for actor_id in actor_ids:
+            owner = self._client_to_actor_ids.get_key(actor_id)
+            self.__forget_actor(actor_id)
+            await self.__send_dead(owner, actor_id, ActorStateUpdate.DeathInfo.Reason.workerDied, worker_id=worker_id)
```
magniloquency, line 214: There's a similar issue here to above

src/scaler/worker/agent/actor_manager.py
```diff
@@ -0,0 +1,285 @@
...
+                logging.warning(f"{self.__class__.__name__}: actor {designation.actor_id!r}: {detail}")
+
+            await self.__send_dead(designation.actor_id, designation.source, reason, detail)
+
+    def destroy(self, reason: str):
+        for task in self._pending_creates.values():
+            task.cancel()
+        self._pending_creates.clear()
+
+        self._designation = None
```
magniloquency: Should we send dead notifications when we cancel creates?

src/scaler/scheduler/controllers/actor_controller.py
```diff
@@ -0,0 +1,232 @@
...
+        ) != bytes(worker_id):
+            logging.error(
+                f"{self.__class__.__name__}: worker {worker_id!r} sent a state update for actor {actor_id!r} it "
+                "does not host, dropping"
+            )
+            return
+
+        # route to the owner from our authoritative table, never the wire `source`
+        owner = self._client_to_actor_ids.get_key(actor_id)
+        self._actor_id_to_state[actor_id] = state
```
magniloquency, line 136: Should we check that `state >= self._actor_id_to_state[actor_id]` first?

src/protocol/message.capnp
```diff
@@ -67,6 +67,8 @@ struct GraphTask {
 struct ClientHeartbeat {
     resource @0 :Status.Resource;
     latencyUS @1 :UInt32;
+    actorIds @2 :List(Data);   # complete set of actors this client claims to own; always the
+                               # full set, never a diff
 }
```
magniloquency, line 72: Was the code that creates the heartbeat updated to populate this field?

src/scaler/scheduler/controllers/actor_controller.py
```diff
@@ -0,0 +1,232 @@
...
+    ):
+        await self._binder.send(
+            client_id, ActorStateUpdate(actorId=actor_id, source=client_id, workerId=worker_id or b"", state=state)
+        )
+
+    async def __send_dead(
+        self,
+        client_id: ClientID,
+        actor_id: ActorID,
+        reason: "ActorStateUpdate.DeathInfo.Reason",
```
magniloquency: String type

src/protocol/message.capnp
```diff
@@ -83,6 +85,7 @@ struct WorkerHeartbeat {
     processors @6 :List(Status.ProcessorStatus);
     capabilities @7 :List(CommonType.TaskCapability);
     workerManagerID @8 :Data;
+    actors @9 :List(Status.ActorHostStatus);
 }
```
magniloquency, line 89: Was the code that creates the heartbeat updated to populate this field?

magniloquency (changes requested):

src/scaler/scheduler/controllers/actor_controller.py
gxugxu: gc

src/scaler/worker/agent/actor_manager.py
gxugxu: no, planning to just let the scheduler tracks it. 

destroy cannot await also.

src/protocol/message.capnp
gxugxu: No, this is not updated. This is intended to be left as another pr. I didn't want to put those code in this pr.

src/protocol/message.capnp
gxugxu: same as above; No, as I intended to put those code in a separate PR.

src/scaler/scheduler/controllers/worker_controller.py
magniloquency, line 136: Does this shutdown actor workers?

magniloquency (changes requested):

src/scaler/scheduler/controllers/worker_controller.py
gxugxu: no, good catch. Fixed in the latest commit.

src/scaler/client/agent/mixins.py
```diff
@@ -87,3 +91,29 @@ async def on_client_disconnect(self, disconnect: ClientDisconnect):
...
+    async def on_send_actor_message(self, actor_message: ActorMessage):
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    async def on_actor_message(self, actor_message: ActorMessage):
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def set_all_actors_dead(self):
+        raise NotImplementedError()
```
magniloquency, line 119: Can we ensure that all fns have annotated return types please?

There's a few other files this applies to:
- `src/scaler/scheduler/controllers/mixins.py`
- `src/scaler/worker/agent/mixins.py`
- `src/scaler/client/agent/actor_manager.py`
- `src/scaler/scheduler/controllers/actor_controller.py`
- `src/scaler/worker/agent/actor_manager.py`

src/scaler/worker/agent/actor_manager.py
```diff
@@ -0,0 +1,285 @@
...
+                await self.__send_dead(
+                    actor_id,
+                    source,
+                    ActorStateUpdate.DeathInfo.Reason.constructorFailed,
+                    "worker has no initialized processor",
+                )
+                return
+
+            await asyncio.sleep(_PROCESSOR_READY_POLL_SECONDS)
+            waited_seconds += _PROCESSOR_READY_POLL_SECONDS
```
magniloquency: Can we replace this wait loop with an event primitive? Processor sends ProcessInitialized on startup, can we wait for that somehow?

magniloquency (changes requested):

src/scaler/worker/agent/actor_manager.py
gxugxu: resolved in the latest commit

src/scaler/client/agent/mixins.py
gxugxu: resolved in the latest commit

gxugxu:
This PR will be split to smaller chunks for easier review, as per @sharpener6's instruction.

PR #842, opened 2026-06-18
Scaler supports OpenSSL

rafa-be:
This PR adds initial TLS support for Scaler, through the use of `tls://` addresses, [see this tutorial](https://github.com/finos/opengris-scaler/pull/842/changes#diff-7c619875accc4b51419ae4dfae4c40b2c41a336b6900c9ecc48c51f7fb16ee1c).

🐞 **There is still an issue when the cluster is started through `SchedulerClusterCombo`**. It seems like OpenSSL has issues when ran inside multiple "spawn" Python processes.

Starting the Scaler components independently (`scaler_scheduler`, `scaler_object_storage_server`, `scaler_gui` ...) with TLS addresses works well.


src/scaler/cluster/object_storage_server.py
rafa-be, line 1: These changes are mostly about checking the OSS readiness using the YMQ connector instead of a raw TCP socket. That more resilient on the address type used by the OSS (e.g. TLS, WS, IPC ...)

gxugxu (approved):

src/cpp/scaler/ymq/pymod/tls_config.h
```diff
@@ -0,0 +1,94 @@
...
+    const char* kwlist[]     = {"cert_chain", "private_key", nullptr};
+
+    if (!PyArg_ParseTupleAndKeywords(
+            args, kwds, "s#s#", (char**)kwlist, &certChain, &certChainLen, &privateKey, &privateKeyLen)) {
+        return -1;
+    }
+
+    try {
+        new (&self->tlsConfig) scaler::ymq::TLSConfig(
+            std::string {certChain, static_cast<size_t>(certChainLen)},
```
gxugxu, line 39: Is this `Py_ssize_t` just a typedef?

rafa-be: Looks like it is: 
https://github.com/python/cpython/blob/87120d333dc54971ce3cdc1c170efab4798f5246/Include/pyport.h#L128-L143

It's guaranteed that `sizeof(Py_ssize_t) == sizeof(size_t)` so I think this cast is safe.

e117649:
> 🐞 **There is still an issue when the cluster is started through `SchedulerClusterCombo`**. It seems like OpenSSL has issues when ran inside multiple "spawn" Python processes.

It looks like this might be that `_ymq.so` and object_storage_server.so each statically link their own copy of OpenSSL and export the OpenSSL symbols with default (global) visibility. When both are loaded into one process, ELF symbol interposition makes one module's SSL_* calls resolve into the other module's OpenSSL build/global state, corrupting the server-side TLS state machine → EPROTO on handshake.

The OSS server process always hits this: ObjectStorageServerProcess (used by the combo and the scaler / scaler_object_storage_server entry points) imports scaler.cluster.object_storage_server → scaler.io.network_backends → scaler.io.ymq._ymq, so _ymq.so gets loaded alongside object_storage_server.so.

In `ScalerPythonModule.cmake`
```
@@ -1,3 +1,5 @@
+set(_SCALER_PYMOD_CMAKE_DIR "${CMAKE_CURRENT_LIST_DIR}")
+

@@ -93,6 +95,11 @@ function(scaler_add_python_module)
         target_link_libraries(${PYMOD_TARGET} PRIVATE ${PYMOD_LINK_LIBRARIES})
     endif()
 
+    if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+        target_link_options(${PYMOD_TARGET} PRIVATE
+            "LINKER:--version-script=${_SCALER_PYMOD_CMAKE_DIR}/scaler_pymod_exports.map")
+    endif()
+
...

e117649 (approved):

PR #845, opened 2026-06-19
feat: acquire TLS certificate via certbot before starting scaler

magniloquency:
Open port 80 in the scheduler security group for the ACME HTTP-01 challenge, then install certbot via `uv tool install certbot` and obtain a Let's Encrypt certificate for the instance's public IP address before the scaler process starts.

The certificate is written to `/etc/letsencrypt/live/$PUBLIC_IP/` and will be consumed when native `wss://` support is wired into the scaler.

Let's Encrypt has issued certificates for bare IP addresses since July 2025 (RFC 8738), so no domain name workaround is required.

### DIFF
--- a/src/scaler/scheduler/controllers/policies/simple_policy/scaling/vanilla.py
+++ b/src/scaler/scheduler/controllers/policies/simple_policy/scaling/vanilla.py
@@ -50,7 +50,10 @@ class VanillaScalingPolicy(ScalingPolicy):
             if task_ratio > self._upper_task_ratio:
                 desired = current + 1
             elif task_ratio < self._lower_task_ratio:
-                desired = 0 if task_count == 0 else max(1, ceil(task_count / self._upper_task_ratio))
+                if task_count == 0 or current == 0:
+                    desired = 0
+                else:
+                    desired = max(1, ceil(task_count / self._upper_task_ratio))
             else:
                 desired = current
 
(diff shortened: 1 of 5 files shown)


rafa-be:
> Let's Encrypt has issued certificates for bare IP addresses since July 2025 (RFC 8738), so no domain name workaround is required.

Nice.

Do you know if browsers complain about these IP-based certificates?

magniloquency:
@rafa-be, I don't know. Do you think they will?

rafa-be:
> @rafa-be, I don't know. Do you think they will?

Just checked, apparently it's supported with a few caveats:

- the user has to explicitly use the IP as hostname;
- these certificates expire after 6 days.

rafa-be (approved):

rafa-be (dismissed):

src/scaler/scheduler/controllers/policies/simple_policy/scaling/vanilla.py
rafa-be, line 1: I think these changes (with the ones in test_scaling.py) have been erroneously pushed to this PR.

magniloquency: This PR builds on work from other PRs some of which were not merged until after this one was opened. This PR is still a draft and this will be cleaned up before it's opened

magniloquency:
@rafa-be 

> the user has to explicitly use the IP as hostname;

This should be fine(?), is JS that's opening the connection and using the IP, so I think that should work.

> these certificates expire after 6 days.

I don't expect scaler to be used for long-lived deployments so I think this is also ok. If we have to, AWS gives each public ec2 instance a dns name, so we could use that

rafa-be:
Yes I think it's fine

rafa-be (approved):

PR #847, opened 2026-06-23
feat: Add "try it" tab to launchpad and embed worker monitor into page

magniloquency:
## Summary

- Adds a **Try it** tab to the Launchpad with a Monaco editor, Pyodide (Python in WebAssembly), and Jedi for autocompletion and hover docs
- Hover tooltips are rendered from parsed docstrings rather than raw text: `docstring_parser` (reST, Google and NumPy) supplies the summary and the Parameters/Returns/Raises lists as Markdown, and the signature is fenced as Python and led with `def`/`class` so Monaco has a keyword to colour regardless of which annotations a name happens to use. Names are resolved from `globals()` first, via `inspect.getdoc()` so an attribute that defines no docstring of its own picks up the inherited one, with Jedi as a secondary fallback for attributes and comprehension variables that never enter `globals()`. Keywords, keyword arguments at a call site, builtin instances, and names carrying neither a signature nor documentation get no tooltip at all
- **Scheduler Logs**, **Worker Monitor**, and **Try it** tabs are gated behind scheduler readiness (`scheduler_address` present); Worker Monitor is embedded in an iframe tab instead of opening externally, now served over the same TLS cert as the scheduler/object storage server so the browser doesn't block it as mixed content
- The client interpreter installs the union of every worker manager's `requirements.txt` (falling back to the default PyPI spec for any worker manager whose scaler requirement is a git/direct-reference install that micropip can't build), since a submitted task can land on any worker manager; this union is frozen for the lifetime of a deployment and only re-applied on redeploy
- The scaler wasm wheel defaults to installing from PyPI via micropip (falling back to a locally staged `scripts/build_wasm.sh` build, gated behind an explicit `LAUNCHPAD_TRYIT_LOCAL_WHEELS=1` opt-in, for wasm-client development), with an advanced-mode Pyodide version override for testing wheels built against a different Pyodide/Emscripten build
- An **Available packages** dropdown, anchored under the Packages button and floated over the editor via a portal so it can't push the editor down, shows each resolved package's actual installed version (via `importlib.metadata`) rather than the raw requirement spec
- Pyodide init is guarded at `window._pyodideReady` so component remounts reuse the existing interpreter; the scaler wheels and the tab's own infra packages (jedi, docstring_parser) are installed in a single `micropip.install` call
- The editor resets to the default snippet, and the output panel clears, whenever the scheduler address changes (e.g. after a redeploy), without touching in-progress edits otherwise


e117649:
Note: JupyterLite 0.8.0 was recently released, which includes Pyodide 314, upgraded it here https://github.com/finos/opengris-scaler/pull/854

rafa-be (approved):

PR #848, opened 2026-06-24
bump orb-py to 1.7.0

magniloquency:
Pins `orb-py` to `==1.7.0`. Verified API compatibility — `ORBClient`, `create_request`, `get_request_status`, `create_return_request`, `create_template`, and `validate_template` are all present and signature-compatible with our usage.

The new `ACQUIRING` status added in 1.7.0 is non-terminal, so the polling loop in `ORBWorkerProvisioner.start_units` handles it correctly without changes.


magniloquency:
Tested and good to go

e117649 (approved):

PR #854, opened 2026-06-29
Upgrade to Pyodide 314 and JupyterLite 0.8.0

e117649:
WASM build is also now able to be released as a wheel to pypi


docs/source/gallery/monte_carlo_pi.ipynb
```diff
@@ -10,6 +10,17 @@
...
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "scaler-install",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# One-time setup: install the Scaler client\n",
+    "%pip install opengris-scaler"
```
gxugxu, line 21: Is there a way to make a `requirements.txt` and put union of all dependencies in `docs/source/gallery/*.ipynb` in it? Since now `examples/application/notebooks/*.ipynb` uses `requirements.txt`, unless there's a good reason, we should try to make things consistent here.

gxugxu: Unless that these examples are meant to run from browser with one click, then please ignore this message.

scripts/build_wasm.sh
```diff
@@ -20,16 +20,16 @@ WASM_VENV_ROOT="${XDG_CACHE_HOME:-${HOME}/.cache}/opengris-scaler"
 WASM_VENV="${WASM_VENV_ROOT}/pyodide-build-venv"
 
 if ! command -v uv >/dev/null 2>&1; then
-    echo "uv is required to create the Python 3.13 wasm build environment."
```
gxugxu, line 23: Would it make sense to set variables on top of the file to store the version? How often you would expect these dependencies version be changed? 

I think it would make sense to have `PYODIDE_BUILD_VERSION`, not sure about `PYTHON_VERSION`.

docs/source/gallery/monte_carlo_pi.ipynb
e117649: yep, these are from the browser with 1 click, Jupyterlite doesn't let us bundle packages with the kernel so they need need to be installed this way. with the old version I used a hack to silently install it, and the hack doesn't work with this new version anymore so this is the only way

magniloquency (dismissed):

magniloquency (approved):

gxugxu (approved):

sharpener6 (approved):

PR #858, opened 2026-06-29
Fix ORB/EC2 auto-scaling: scale-down cooldown and worker manager ID mismatch

magniloquency:
## Summary

- **Scale-down cooldown**: `VanillaScalingPolicy` previously scaled down immediately on any heartbeat where `task_count == 0`, which caused instances to be terminated between task waves (e.g. while the client is submitting the next batch). A 30-second cooldown now holds the current worker count before applying any reduction.

- **Worker manager ID mismatch on EC2**: Workers launched via the ORB provisioner used the EC2 instance ID (`$INSTANCE_ID`) as their `--worker-manager-id`, but the `WorkerManagerRunner` heartbeated to the scheduler using the configured `worker_manager_id` (e.g. `wm-1`). The scheduler therefore never matched connected workers to their manager, so scaling commands had no effect. Fixed by passing the configured ID directly into the user-data script instead of looking it up from instance metadata.

### DIFF
--- a/src/scaler/scheduler/controllers/policies/simple_policy/scaling/capability_scaling.py
+++ b/src/scaler/scheduler/controllers/policies/simple_policy/scaling/capability_scaling.py
@@ -1,3 +1,4 @@
+import time
 from collections import defaultdict
 from math import ceil
 from typing import Dict, FrozenSet, List, Tuple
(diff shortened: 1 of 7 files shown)


magniloquency:
Splitting into two separate PRs.

PR #860, opened 2026-06-29
scaling policies: add scale-down cooldown to vanilla, capability, and waterfall

magniloquency:
## Summary

All three scaling policies scaled down instantly whenever the computed desired count
dropped below what's currently running -- including the generic zero-tasks case and,
for capability-aware policies, a capability set falling out of the declarative output
entirely once it has no more pending tasks. In practice this fires between waves of
work (e.g. while a client is submitting the next batch) or during brief lulls, causing
worker managers -- especially ones provisioning external resources with long startup
times -- to terminate instances only to re-provision them seconds later.

A 30s cooldown now holds the current count the first time a policy wants to reduce
below it, per worker manager (and per capability set for capability-aware policies).
Any observation that calls for scale-up resets the cooldown. This is implemented once,
via a shared `ScaleDownCooldown` helper used by composition, and applied consistently
across `VanillaScalingPolicy`, `CapabilityScalingPolicy`, and `WaterfallScalingPolicy`:

- `apply(desired, current)` gates a single value against a real, currently-connected
  worker count (Vanilla's whole-manager desired; Waterfall's generic/empty-capset entry).
- `reconcile(desired_per_capset)` gates a declarative per-capability-set list against
  each capset's own last-emitted value, treating a capset that drops out of the list
  entirely (no more pending tasks) the same as an explicit reduction to 0 -- so it's
  held at its last value during the cooldown window instead of draining instantly
  (Capability's per-capset entries; Waterfall's capability-specific entries).

Cooldown state is scoped per worker-manager-id everywhere, including Vanilla, which
also fixes a latent bug where a single unkeyed cooldown timer would be shared (and
bleed) across multiple worker managers in a multi-manager Simple-policy deployment.


src/scaler/scheduler/controllers/policies/simple_policy/scaling/vanilla.py
```diff
@@ -8,15 +9,23 @@
...
+    Scale-down is gated by a cooldown: once the policy first wants to reduce the desired
+    worker count below the current managed count, it holds at the current count for
+    _SCALE_DOWN_COOLDOWN_SECONDS before applying the reduction. Any observation that
+    calls for scale-up resets the cooldown.
     """
 
     def __init__(self):
         self._lower_task_ratio = 1
         self._upper_task_ratio = 10
+        self._scale_down_since: Optional[float] = None
```
gxugxu: If we are to do this cooldown systematically, can we have `self._scale_down_since` to have type `float`, and `=0` means no cooldown?

src/scaler/scheduler/controllers/policies/simple_policy/scaling/vanilla.py
```diff
@@ -26,9 +35,27 @@ def get_scaling_commands(
...
         desired_per_capset: List[Tuple[Dict[str, int], int]] = [({}, desired)]
         return [build_set_desired_command(desired_per_capset)]
 
+    def _apply_scale_down_cooldown(self, desired: int, current: int) -> int:
+        if desired >= current:
+            self._scale_down_since = None
+            return desired
+
+        now = time.time()
+        if self._scale_down_since is None:
```
gxugxu: and in this case we can have `self._scale_down_since=max(now, 0)`?

gxugxu (commented):
current code looks okay

magniloquency:
replaced by https://github.com/finos/opengris-scaler/pull/873

PR #867, opened 2026-07-02
Cleanup logging level names

gxugxu:
# READY FOR REVIEW.

Cleanup log output level name and format
- Logging.CTIC becomes Logging.CRITICAL
- Logging.EROR becomes Logging.ERROR
- Logging.DEBG becomes Logging.DEBUG
- Logging.WARN becomes Logging.WARNING

which are more conventional.

- Formatting an enum with trailing comma


e117649:
Is it easy to change the values to match Python's? https://docs.python.org/3/library/logging.html#logging-levels

gxugxu:
> Is it easy to change the values to match Python's? https://docs.python.org/3/library/logging.html#logging-levels

Edited other names to be aligned with what Python uses. The numeric values are still not aligned. Because it's not communicated to Python, nor does it need to. In C++ it's more common to have consecutive enum values starting from 0. So that's been kept.

e117649 (dismissed):

src/cpp/scaler/logging/logging.h
```diff
@@ -19,7 +18,14 @@ namespace ymq {
 
 class Logger {
 public:
-    enum LoggingLevel { critical = 0, error = 1, warning = 2, info = 3, debug = 4, notset = 5 };
+    enum LoggingLevel {
```
rafa-be: Not sure if you did it on purpose (to make its use shorter), but that might as well be defined as an `enum class`.

rafa-be (dismissed):

rafa-be (approved):

PR #868, opened 2026-07-02
tests: audit and fix all tests

e117649:
Uses `moto` for integration tests in the regular CI, and `localstack` for a complete E2E stress test that can be triggered manually

Created a separate pull request for the e2e tests here https://github.com/finos/opengris-scaler/pull/872


tests/cpp/ymq/common/utils_linux.cpp
```diff
@@ -11,3 +15,31 @@ void raiseSocketError(const char* msg)
 {
     throw std::system_error(errno, std::generic_category(), msg);
 }
+
+int getFreePort()
```
gxugxu: I think similar utility exists already.

tests/cpp/ymq/common/utils_windows.cpp
```diff
@@ -11,4 +12,49 @@ void raiseSystemError(const char* msg)
...
 {
     throw std::system_error(WSAGetLastError(), std::generic_category(), msg);
+}
+
+int getFreePort()
+{
+    // WSAStartup is reference-counted, so this is safe even if Winsock is already initialised by the
+    // socket tests; the matching WSACleanup below only decrements that count.
+    WSADATA wsaData;
+    bool initialised = WSAStartup(MAKEWORD(2, 2), &wsaData) == 0;
```
gxugxu: Somewhere in the codebase you can find an `RAII` style wrap of `WSAStartup`. If we have
```
struct foo {
  foo()  { WSAStartUp(...); set_initialize_to_true; }
  ~foo() { if(initialized) WSACleanup(); }
};
```
then we don't have to test against whether `WSA` is initialized every time the system calls goes wrong.

tests/cpp/ymq/common/utils_windows.cpp
```diff
@@ -11,4 +12,49 @@ void raiseSystemError(const char* msg)
...
+
+    SOCKET fd = ::socket(AF_INET, SOCK_STREAM, 0);
+    if (fd == INVALID_SOCKET) {
+        if (initialised) {
```
gxugxu: E.G. here.

tests/compat/test_ray_compat.py
```diff
@@ -1,16 +1,30 @@
...
+    import numpy as np
+    from numpy import random
+except ModuleNotFoundError:  # optional test dependency
+    np = None
+    random = None
+
+try:
+    import ray
+except ModuleNotFoundError:  # optional test dependency
+    ray = None
```
gxugxu: Why would it be a bad idea to make dependencies optional? We have full control over the testing environment?

tests/compat/test_ray_compat.py
```diff
@@ -79,6 +93,10 @@ def h():
 
     # https://docs.ray.io/en/latest/ray-core/patterns/nested-tasks.html#code-example
     def test_ray_example_nested_quicksort(self) -> None:
+        # Kept small (the docs example uses millions) so the test still drives several levels of
+        # distributed recursion and verifies correctness without multi-second sorts of 8M elements.
+        _QUICKSORT_INLINE_THRESHOLD = 1000
```
gxugxu: I think we don't need to lower the threshold here. The original threshold completes well under 1sec.

tests/core/test_death_timeout.py
```diff
@@ -25,14 +25,32 @@
...
+# Workers that cannot reach a scheduler self-terminate once they exhaust their connection-retry
+# backoff. The tests below bound every wait on that behaviour, so they cannot loop forever and are
+# safe under `unittest discover`.
+DEATH_TIMEOUT_SECONDS = 10
+# Generous upper bound (loaded CI box) on how long a cluster shutdown may take.
+TEARDOWN_TIMEOUT_SECONDS = 30
+# A worker manager that never reaches a scheduler exits only after its workers exhaust their
+# connection-retry backoff (~50s on an unloaded box), which dominates the death timeout. Bound the
+# wait well above that so a real regression fails the test instead of hanging CI, without flaking.
+NO_SCHEDULER_EXIT_TIMEOUT_SECONDS = 120
```
gxugxu, line 34: I think we should set it in `defaults.py` or use existing setting if we have one.

gxugxu: This variable is not referenced?

tests/core/test_death_timeout.py
```diff
@@ -73,15 +98,26 @@ def test_shutdown(self):
...
-        # this is combo cluster, client only shutdown clusters, not scheduler, so scheduler need be shutdown also
-        cluster.shutdown()
+            logging.info("Shutting down")
+            client.shutdown()
+
+        # client.shutdown() stops the workers connected to the scheduler, but not the combo's own scheduler.
+        # Poll until the worker manager process has actually exited instead of sleeping a fixed amount.
+        deadline = time.monotonic() + TEARDOWN_TIMEOUT_SECONDS
+        while cluster._worker_manager_process.is_alive() and time.monotonic() < deadline:
+            time.sleep(0.05)
```
gxugxu: Stay consistent between hard-coded numbers and `POLL_INTERVAL_SECONDS = 0.05`

PR #872, opened 2026-07-06
tests: add aws integration e2e tests

### LINKED ISSUE #946: Create E2E testing framework

e117649:
- adds `moto` for integration tests (with cloud APIs) that run everything short of booting real workers with our existing tests on every commit

- adds heavier full e2e mock with actual workers and multiple worker managers using `floci` and docker that can be manually triggered to run on any branch and can provide some basic config e.g. waterfall policy string

closes https://github.com/finos/opengris-scaler/issues/946


.github/actions/run-test/action.yml
```diff
@@ -52,6 +52,21 @@ runs:
...
+      # itself unless RUN_INTEGRATION_TESTS=1, and moto is not in [all]. This step is what actually
+      # runs the suite -- install moto, flip the gate, Linux-only to keep the win/mac lanes lean. The
+      # heavier LocalStack and scaling-stress runs are separate manual workflows.
+      if: inputs.os == 'Linux'
+      shell: bash
+      env:
+        RUN_INTEGRATION_TESTS: '1'
+      run: |
+        # Install the integration extra (moto) from the built wheel so the pin lives only in pyproject.
+        wheel_file=$(ls dist/*.whl | head -n 1)
```
gxugxu: I think we should not rely on how `ls` format its output. Also, we should express the intent of what we are trying to do.

Is it possible to use `find(1)` together with `sort` instead of `ls` here? I think something like
```
find ./dist -maxdepth=1 -name '*.whl' | sort -n | head -n 1
```
would be equivalent, assuming on this machine `ls` output its result sorted lexicographically.

gxugxu: This command tells we are finding files and what we are looking for, although possibly more complex.

Hardcoded string, in retrospect, is not ideal. Because then we would have to update this file in every PR.

.github/workflows/integration-localstack.yml
```diff
@@ -0,0 +1,73 @@
...
+  pull_request:
+    types: [ labeled ]
+
+permissions:
+  contents: read
+
+jobs:
+  localstack-integration:
+    name: LocalStack Integration (Linux)
+    # Manual dispatch always runs; on a PR only when it carries the "localstack" label.
```
gxugxu: We should populate this information to README.md or some development documentation.

.github/workflows/integration-localstack.yml
```diff
@@ -0,0 +1,73 @@
...
+      - uses: ./.github/actions/compile-libraries
+        with:
+          os: ${{ runner.os }}
+
+      - name: Build and install wheel
+        shell: bash
+        run: |
+          uv pip install --system build
+          python -m build --wheel
+          wheel_file=$(ls dist/*.whl | head -n 1)
```
gxugxu: Please see above.

.github/actions/run-test/action.yml
```diff
@@ -52,6 +52,21 @@ runs:
       run: |
         python -m unittest discover -v tests -t .
 
+    - name: Run End-to-End Integration Tests (Linux, moto backend)
+      # tests/integration is opt-in: the "Run Unittests" step above collects it but every test skips
+      # itself unless RUN_INTEGRATION_TESTS=1, and moto is not in [all]. This step is what actually
+      # runs the suite -- install moto, flip the gate, Linux-only to keep the win/mac lanes lean. The
+      # heavier LocalStack and scaling-stress runs are separate manual workflows.
```
rafa-be: Claude is always adding these overly long comments referring to the prompt, or just repeating exactly what the code says.

We should come up with some instruction in `AGENTS.md` to avoid these.

.github/actions/run-test/action.yml
```diff
@@ -52,6 +52,21 @@ runs:
...
+      if: inputs.os == 'Linux'
+      shell: bash
+      env:
+        RUN_INTEGRATION_TESTS: '1'
+      run: |
+        # Install the integration extra (moto) from the built wheel so the pin lives only in pyproject.
+        wheel_file=$(ls dist/*.whl | head -n 1)
+        uv pip install --system "$wheel_file[integration]"
```
rafa-be: I'm surprised this step is required at all. The previous step (*Install Wheel*) does this already:

```
uv pip install --system "$wheel_file[all]"
```

.github/actions/run-test/action.yml
rafa-be: Not that I don't agree, but the existing *Install Wheel* step does this too.

.github/workflows/integration-localstack.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          os: ${{ runner.os }}
+
+      - uses: ./.github/actions/3rd-party-libraries-install
+        with:
+          os: ${{ runner.os }}
+
+      - uses: ./.github/actions/compile-libraries
+        with:
+          os: ${{ runner.os }}
```
rafa-be: These steps are exactly the same as the ones in `build-and-test.ymq`. Any way to share these somehow?

.github/workflows/scaling-stress.yml
```diff
@@ -0,0 +1,92 @@
...
+          os: ${{ runner.os }}
+
+      - name: Build and install wheel
+        shell: bash
+        run: |
+          uv pip install --system build
+          python -m build --wheel
+          wheel_file=$(ls dist/*.whl | head -n 1)
+          echo "Installing wheel: $wheel_file"
+          uv pip install --system "$wheel_file"
```
rafa-be: Duplicate code with `integration-localstack.yml` and `build-and-test.yml`.

tests/integration/test_dynamic_local_e2e.py
rafa-be, line 1: If this one does not require neither Moto nor LocalStack, is it really an integration test?

It seems to be it might be a duplicate with some tests in `tests/entry_points/test_worker_manager.py`.

.github/actions/run-test/action.yml
e117649: agreed, fixed

.github/actions/run-test/action.yml
e117649: I added some instructions that should help with this, along with some other Claude issues

.github/workflows/integration-localstack.yml
e117649: Added to `tests/integration/README.md` with a reference to that readme added to https://finos.github.io/opengris-scaler/tutorials/development/guidelines.html

tests/integration/test_dynamic_local_e2e.py
e117649: yes you're right, it was redundant and only added one small assert I added to the existing test

.github/actions/run-test/action.yml
e117649: `moto` is a dev dependency  so I don't think it belongs under `all`, but I don't like that claude put it under `integration` either so I moved it to --group dev with all the other dev dependencies

.github/workflows/integration-localstack.yml
e117649: Good call, moved it to a build-native-dependencies action they can all share

.github/workflows/build-and-test.yml
```diff
@@ -35,30 +35,53 @@ jobs:
...
 
+      - uses: ./.github/actions/build-and-install-wheel
+
+      - name: Run End-to-End Integration Tests (moto backend)
+        shell: bash
+        env:
+          RUN_INTEGRATION_TESTS: '1'
+        run: |
+          python -m unittest discover -s tests/integration -t . -v
+
```
rafa-be: I think this might not actually work.

[The log just shows a huge list of never ending exceptions](https://github.com/finos/opengris-scaler/actions/runs/29369839991/job/87210280377?pr=872#step:8:95).

.github/actions/run-test/action.yml
```diff
@@ -34,23 +34,22 @@ runs:
...
-    - name: Run Unittests
+    - name: Run End-to-End Integration Tests (Linux, moto backend)
+      # The run above skips tests/integration (gated on RUN_INTEGRATION_TESTS); run it here with the gate on.
+      if: inputs.os == 'Linux'
       shell: bash
+      env:
+        RUN_INTEGRATION_TESTS: '1'
       run: |
-        python -m unittest discover -v tests -t .
+        python -m unittest discover -s tests/integration -t . -v
```
rafa-be: Do we need this as most of the tests will be skipped, and as `integration-py311` will run them anyway?

.github/workflows/build-and-test.yml
```diff
@@ -35,30 +35,53 @@ jobs:
...
         with:
           os: ${{ runner.os }}
 
-      - uses: ./.github/actions/3rd-party-libraries-install
+      - uses: ./.github/actions/run-test
         with:
           os: ${{ runner.os }}
 
-      - uses: ./.github/actions/compile-libraries
+  integration-py311:
```
rafa-be: Just rename it `integration`. We have steps using different versions of Python, and we don't use the `pyXXX` suffix for these.

tests/utility/utility.py
```diff
@@ -1,6 +1,24 @@
 import logging
+import multiprocessing
 import unittest
 
+# Shared timing bounds for tests that spawn real processes / clusters: one poll granularity and one
+# "wait for a spawned process to terminate" bound, so the process-lifecycle tests do not each
+# hard-code their own. Only reached in full on failure, so the bound errs on the side of patience.
+POLL_INTERVAL_SECONDS = 0.05
```
rafa-be: This variable should go into `_harness.py.

tests/scheduler/test_scaling.py
```diff
@@ -96,6 +100,8 @@ def test_scaling_basic(self):
 
         with Client(self.scheduler_address) as client:
             client.map(time.sleep, [0.1] * 100)
```
rafa-be, line 102:
```suggestion
```

Useless with the following line

.github/workflows/container-e2e.yml
```diff
@@ -0,0 +1,72 @@
...
+      - uses: ./.github/actions/build-native-dependencies
+        with:
+          os: ${{ runner.os }}
+
+      - uses: ./.github/actions/build-and-install-wheel
+
+      - name: Run container-scaling e2e
+        shell: bash
+        env:
+          RUN_CONTAINER_E2E: '1'
```
rafa-be: All these Workflows files are exactly identical except for specific `RUN_*` environment variable, and some varying Python version (couldn't all be 3.13 ?):

```ymq
          RUN_CONTAINER_E2E: '1'
```

```yml
          RUN_CROSS_BACKEND_E2E: '1'
```

```yml
          RUN_EC2_E2E: '1'
```

```ymq
          RUN_FLOCI_E2E: '1'
```


This is bad design. 

For example, it could be a single workflow (`e2e-test.yml` ?), and guarded steps:

```yml
steps:
  ...

  floci-e2e:
    if: ${{ github.event.label.name == 'floci-e2e' }}
    env:
        RUN_CONTAINER_E2E: '1'
    ...

  ec2-e2e:
    if: ${{ github.event.label.name == 'ec2-e2e' }}
    env:
      RUN_EC2_E2E: '1'
    ...
    
  cross-backend-e2e:
     if: ${{ github.event.label.name == 'cross-backend-e2e' }}
     env:
          RUN_CROSS_BACKEND_E2E: '1'
```

scripts/e2e/build_cibuildwheel.sh
rafa-be, line 1: Could we move all these scripts into its own directory (`scripts/e2e-testing` ?).

The scripts directory becomes a little bit cramped.

rafa-be (commented):
I spent most of the time reviewing the CI workflows and scripts, and looked a bit at the test code itself.

TBO, I am absolutely not sure the tests do what they pretend to do. Even looking at the CI logs, these might be testing nothing, I'm not sure. The CI log 95% exceptions but the tests somehow pass.

No offense, as I know you did this under thigh time constraints, but the quality LLM-generated tests is mediocre.

I'm not against merging it iff:

- it stays self-contained in its own directory. That includes the scripts that should go into their own directory;
- we can at least validate it does anything useful by looking at the log. [This is clearly current not the case](https://github.com/finos/opengris-scaler/actions/runs/29369839991/job/87210280377?pr=872#step:8:95);
- there is a little bit of cleanup on the Github Action side.

magniloquency:
The PR description mentions moto but I don't think I see it anywhere in this PR?

.github/actions/build-and-install-wheel/action.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Build And Install Wheel'
+description: 'Build the Scaler wheel and install it with all extras and the dev dependency-group'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Build and install wheel
+      shell: bash
+      run: |
+        uv pip install --system build
```
magniloquency, line 10: maybe this is out of scope but why aren't we using a venv?

.github/actions/build-native-dependencies/action.yml
```diff
@@ -0,0 +1,26 @@
...
+      with:
+        os: ${{ inputs.os }}
+
+    - uses: ./.github/actions/3rd-party-libraries-compile
+      with:
+        os: ${{ inputs.os }}
+
+    - uses: ./.github/actions/3rd-party-libraries-install
+      with:
+        os: ${{ inputs.os }}
```
magniloquency, line 22: I like this action but the naming seems off, we should either rename each of these to `native-libraries-*` or *this* action to `build-3rd-party-libraries` to be consistent

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,90 @@
...
+        description: "Which deployment to scale."
+        required: true
+        type: choice
+        options: [ container, container_waterfall, ecs, ec2, ecs_ec2 ]
+      num_tasks:
+        description: "Tasks per burst wave. Blank = the backend's default."
+        required: false
+        type: string
+      task_seconds:
+        description: "Sleep per task, seconds. Blank = the backend's default (0.15)."
```
magniloquency, line 20: does this mean 0.15 seconds?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: 'Checkout Repository'
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Resolve topology
+        shell: bash
+        run: echo "SCALER_E2E=${SCALER_E2E#e2e-}" >> "$GITHUB_ENV"
+
+      - uses: ./.github/actions/setup-env
+        with:
+          os: ${{ runner.os }}
```
magniloquency, line 61: Also off topic but this is something I've been wondering about for a while: Couldn't the action just get `runner.os` internally instead of us passing it in like this?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      # Installs scaler on the host, which is what runs the client, scheduler, and manager processes.
+      - uses: ./.github/actions/build-and-install-wheel
+
+      # The EC2 topologies boot instances that install the CURRENT source, so they need a portable wheel
+      # of it (not a PyPI release). Matches the interpreter above, so it is the same code under test.
+      - name: Build current-source manylinux wheel
+        if: ${{ env.SCALER_E2E == 'ec2' || env.SCALER_E2E == 'ecs_ec2' }}
+        shell: bash
+        run: CIBW_OUTPUT_DIR=dist_manylinux ./scripts/e2e/build_cibuildwheel.sh --only cp313-manylinux_x86_64
```
magniloquency, line 76: python 3.14?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Run e2e
+        shell: bash
+        env:
+          SCALER_IT_NUM_TASKS: ${{ inputs.num_tasks }}
+          SCALER_IT_TASK_SECONDS: ${{ inputs.task_seconds }}
+          SCALER_IT_WATERFALL_POLICY: ${{ inputs.policy_content }}
+        run: |
+          python -m unittest tests.integration.test_scaling -v 2>&1 | tee e2e.log
```
magniloquency, line 85: How does this know which test case(s) to run? It seems to me like that is a parameter of the workflow. I'll probably understand when I get to the test file but I think it would be good to have a comment in here somewhere

scripts/e2e/run.sh
```diff
@@ -0,0 +1,51 @@
...
+# Usage:
+#       ./scripts/e2e/run.sh <topology>            container | container_waterfall | ecs | ec2 | ecs_ec2
+#       DOCKER="sudo docker" ./scripts/e2e/run.sh ecs        # if the docker socket is root-only
+#
+# The harness prints a "web GUI: http://localhost:PORT" line you can open to watch the scaling live.
+#
+# Tunables, same as the workflow_dispatch inputs:
+#       SCALER_IT_NUM_TASKS         tasks per burst wave
+#       SCALER_IT_TASK_SECONDS      sleep per task, seconds
+#       SCALER_IT_WATERFALL_POLICY  raw waterfall_v1 policy for the waterfall topologies, one rule per
```
magniloquency, line 14: Maybe I'm missing something but what does `IT` mean?

tests/integration/backends.py
```diff
@@ -0,0 +1,432 @@
...
+    """A real provisioner driven by the scheduler like the cloud managers, but each "machine" is a container
+    running a fixed ``baremetal_native`` worker -- the local, free analog of a booted instance. Each machine
+    gets its own IP, so spread is real."""
+
+    def __init__(
+        self, worker_scheduler_address: str, workers_per_machine: int, max_machines: int, name_prefix: str
+    ) -> None:
+        self._runtime = DockerRuntime()
+        self._worker_scheduler_address = worker_scheduler_address  # gateway-reachable, for the containers
+        self._workers = workers_per_machine
```
magniloquency, line 65: for clarity we should note that the variable is _per_ unit (machine)

```py
        self._workers_per_unit = workers_per_unit
```

tests/integration/backends.py
```diff
@@ -0,0 +1,432 @@
...
+
+    async def start_units(self, count: int) -> None:
+        for _ in range(count):
+            self._counter += 1
+            name = f"{self._name_prefix}-{self._counter}"
+            command = [
+                "scaler_worker_manager", "baremetal_native", self._worker_scheduler_address,
+                "--worker-manager-id", name,
+                "--mode", "fixed",
+                "--num-of-workers", str(self._workers),
```
magniloquency, line 91: I think this parameter is deprecated in favour of `--max-task-concurrency`

tests/integration/backends.py
```diff
@@ -0,0 +1,432 @@
...
+    max_machines: int,
+    worker_manager_id: str,
+    name_prefix: str,
+) -> None:
+    """Connects to the scheduler over loopback and provisions container machines that reach it via
+    ``worker_scheduler_address`` (the docker-bridge gateway)."""
+    from scaler.config.defaults import DEFAULT_HEARTBEAT_INTERVAL_SECONDS, DEFAULT_IO_THREADS
+    from scaler.config.types.address import AddressConfig
+    from scaler.utility.logging.utility import setup_logger
+    from scaler.worker_manager_adapter.worker_manager_runner import WorkerManagerRunner
```
magniloquency, line 124: Can we move these imports to the top?

tests/integration/backends.py
```diff
@@ -0,0 +1,432 @@
...
+    scale-up divisor, so the pool caps at ``ceil(max_task_concurrency / ecs_task_cpu)`` tasks."""
+    point_boto3_at_floci(endpoint_url)
+
+    from scaler.config.common.worker import WorkerConfig
+    from scaler.config.common.worker_manager import WorkerManagerConfig
+    from scaler.config.section.ecs_worker_manager import ECSWorkerManagerConfig
+    from scaler.config.types.address import AddressConfig
+    from scaler.config.types.worker import WorkerCapabilities
+    from scaler.utility.logging.utility import setup_logger
+    from scaler.worker_manager_adapter.aws_raw.ecs import ECSWorkerManager
```
magniloquency, line 216: Moves these imports to the top?

tests/integration/docker.py
magniloquency, line 1: why not use this? https://docker-py.readthedocs.io/en/stable/

magniloquency (commented):
Partial review, mostly style and a few questions

PR #873, opened 2026-07-07
feat: Add scale-down cooldown to worker manager capacity coordinator

magniloquency:
## Summary
- The scheduler eagerly requests scale-downs as soon as demand dips, which can cause worker managers to flap under intermittent load.
- `CapacityCoordinator` now defers any scale-down until it has been continuously requested for `scale_down_cooldown_seconds`. Further decreases while deferred don't extend the window, but a scale-up (or the desired count returning to/above the live count) cancels it.
- Cooldown timing is handled by a new generic `Cooldown` utility (`src/scaler/utility/cooldown.py`), composed into `CapacityCoordinator`.
- New `WorkerManagerConfig.scale_down_cooldown_seconds` field (default 30s; `None` or `0` disables it), threaded through all worker manager backends.


rafa-be (dismissed):

src/scaler/utility/cooldown.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Call `start_if_not_running` to arm the timer if it is not already running, then poll
+    `remaining_seconds` to check how much time is left. `reset` disarms the timer.
+
+    Args:
+        duration_seconds: How long the timer runs once armed. None disables the timer
+            entirely: it never arms, and `remaining_seconds` always returns None.
+    """
+
+    def __init__(self, duration_seconds: Optional[float]) -> None:
```
gxugxu: Would it be possible to make `duration_seconds` and `_started_at` both of type `float`? I think some sentinel values can be used reasonably here.

gxugxu: why not just use `=-1`?

src/scaler/config/defaults.py
```diff
@@ -58,6 +58,10 @@
 # number of workers, echo worker use 1 process
 DEFAULT_MAX_TASK_CONCURRENCY = os.cpu_count()
 
+# minimum number of seconds after a scale-up before a worker manager will honor a scale-down request,
+# to avoid flapping under intermittent load. 0 disables the cooldown.
+DEFAULT_SCALE_DOWN_COOLDOWN_SECONDS = 30
```
gxugxu: Please name it to `DEFAULT_WORKER_MANAGER_SCALE_DOWN_COOLDOWN_SECONDS` and place it after `DEFAULT_WORKER_MANAGER_TIMEOUT_SECONDS`

Rationale: Easy to find through common prefix. Scale down really is part of worker manager's job.

gxugxu (commented):
Other than that, it's fine.

src/scaler/config/defaults.py
magniloquency: done

src/scaler/utility/cooldown.py
magniloquency: done, treating 0 as disabled and rejecting <0 with value error

RaphaelJ (approved):

rafa-be (dismissed):

gxugxu (dismissed):

src/scaler/config/defaults.py
```diff
@@ -42,6 +42,10 @@
 # if didn't receive heartbeat for following seconds, then scheduler will treat worker manager as dead and disconnect it
 DEFAULT_WORKER_MANAGER_TIMEOUT_SECONDS = 10
 
+# minimum number of seconds after a scale-up before a worker manager will honor a scale-down request,
```
e117649: comment is off, minimum number of seconds after a scale-down request, not scale-up (the code and docstring are correct)

e117649:
Up to you if you want to change it here or not now (since we use it elsewhere), but we should start using time.monotonic instead of time.time, to prevent issues with system clock updates, or running during a time change depending on how the system updates its clock (rare but could happen since this is a long running service).

gxugxu (approved):

e117649 (approved):

rafa-be (approved):

rafa-be (approved):

PR #875, opened 2026-07-07
Cleaning up WebSocket classes in YMQ

rafa-be:
Makes the `WebSocketStream` class use a similar interface as the `SecureSocket`/`SecureServer`, `TCPSocket`/`TCPServer` and `Pipe`/`PipeServer` classes used by other socket types.

Before, the `WebSocketStream` had specific methods such as `upgradeAsServer()` or `upgradeAsClient()` to handle the WebSocket handshake. 

That removes all the WebSocket complexity from the `AcceptServer` and `ConnectClient` classes.


src/cpp/scaler/ymq/internal/websocket_stream.h
```diff
@@ -3,43 +3,34 @@
...
-        scaler::utility::MoveOnlyFunction<void(std::expected<WebSocketStream, scaler::wrapper::uv::Error>)>
-            callback) noexcept;
-
-    // Perform the client-side HTTP/1.1 Upgrade handshake on a connected TCP socket,
-    // then call callback with a ready-to-use WebSocketStream.
-    static void upgradeAsClient(
-        scaler::wrapper::uv::TCPSocket socket,
-        const WebSocketAddress& address,
-        scaler::utility::MoveOnlyFunction<void(std::expected<WebSocketStream, scaler::wrapper::uv::Error>)>
-            callback) noexcept;
```
rafa-be, line 42: Most of the refactoring is about replacing `upgradeAsServer` and `upgradeAsClient` new member methods `accept()` and `connect()`, respectively; similar to other socket types.

The underlying TCP socket is now "hidden" within the WebSocket class, like for `SecureSocket`.

tests/cpp/ymq/internal/test_websocket_stream.cpp
```diff
@@ -27,80 +27,90 @@ void runUntil(scaler::wrapper::uv::Loop& loop, const bool& done)
         loop.run(UV_RUN_ONCE);
 }
```
rafa-be, line 29: Most of the changes in these tests are the `TestWebSocketStreamPair` class now replacing the old `TestTCPPair` and `TestWebSocketStreamPair` classes.

The tests themselves haven't changed much.

src/cpp/scaler/ymq/internal/websocket_server.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+std::expected<WebSocketAddress, scaler::wrapper::uv::Error> WebSocketServer::getSockName() const noexcept
+{
+    std::expected<scaler::wrapper::uv::SocketAddress, scaler::wrapper::uv::Error> tcpAddress = _server.getSockName();
+    if (!tcpAddress.has_value()) {
+        return std::unexpected {tcpAddress.error()};
+    }
+
+    // Reconstruct the WebSocket address with the actual bound port (handles port 0 auto-assignment).
+    WebSocketAddress address = _address.value();
```
magniloquency, line 62: This fn is declared noexcept but i think this could throw?

src/cpp/scaler/ymq/internal/websocket_stream.h
```diff
@@ -62,25 +67,40 @@ class WebSocketStream {
...
-    static void finishServerUpgrade(std::shared_ptr<ServerUpgradeContext> ctx) noexcept;
+    enum class Role { Undefined, Client, Server };
 
     struct State {
         scaler::wrapper::uv::TCPSocket _socket;
-        bool _isServer;
+        Role _role {Role::Undefined};
+
+        // Used only during the HTTP Upgrade handshake, reset once completed.
+        std::optional<scaler::wrapper::uv::ConnectCallback> _upgradeCallback {};
```
magniloquency: Type should be `HandshakeDoneCallback`?

magniloquency (changes requested):

src/cpp/scaler/ymq/internal/websocket_stream.h
rafa-be: Done

src/cpp/scaler/ymq/internal/websocket_server.cpp
rafa-be: Done

magniloquency (dismissed):

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -214,83 +216,98 @@ std::expected<std::optional<DecodedFrame>, scaler::wrapper::uv::Error> tryDecode
...
-    if (acceptIt == headerMap.end() || acceptIt->second != computeWebSocketAccept(ctx->key)) {
-        ctx->callback(std::unexpected(scaler::wrapper::uv::Error {UV_EPROTO}));
+    if (acceptIt == headerMap.end() || acceptIt->second != computeWebSocketAccept(key)) {
+        completeUpgrade(state, std::unexpected(scaler::wrapper::uv::Error {UV_EPROTO}));
         return;
     }
 
     // Preserve any data that arrived after the HTTP headers (should be empty in practice).
-    std::vector<uint8_t> leftover(
-        ctx->recvBuffer.begin() + static_cast<std::ptrdiff_t>(headersEnd + 4), ctx->recvBuffer.end());
```
gxugxu, line 279: Why 4? Is it the length of string 'HTTP'?

Write a line above like this:
```
const size_t STRING_HTTP_LENGTH=4;
```
and use this variable instead.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -316,150 +333,158 @@ void WebSocketStream::finishServerUpgrade(std::shared_ptr<ServerUpgradeContext>
...
     }
 
     // Verify Connection header contains "upgrade" (case-insensitive, may be a token list)
     if (toLower(connectionIt->second).find("upgrade") == std::string::npos) {
-        ctx->callback(std::unexpected(scaler::wrapper::uv::Error {UV_EPROTO}));
+        completeUpgrade(state, std::unexpected(scaler::wrapper::uv::Error {UV_EPROTO}));
         return;
     }
 
     if (versionIt->second != "13") {
```
gxugxu, line 352: The previous `"websocket"` hard coded string is fine, but why "13"? Current version of ws we are using?

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -316,150 +333,158 @@ void WebSocketStream::finishServerUpgrade(std::shared_ptr<ServerUpgradeContext>
...
         });
+}
 
-    if (!readStartResult.has_value()) {
-        auto cb = std::move(ctx->callback);
-        cb(std::unexpected(readStartResult.error()));
+void WebSocketStream::completeUpgrade(
+    const std::shared_ptr<State>& state, std::expected<void, scaler::wrapper::uv::Error> result) noexcept
+{
+    if (!state->_upgradeCallback.has_value()) {
```
gxugxu: Why do user have choice to _not_ submit an `_upgradeCallback`? If user shouldn't be able to do that, then we should force them to pass it in, and put `assert` at the beginning of `completeUpgrade`, instead of doing silent return.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
```diff
@@ -557,7 +582,7 @@ void WebSocketStream::processRecvBuffer(std::shared_ptr<State> state) noexcept
             auto pongPayload = frame.payload;
             if (pongPayload.size() > MAX_CONTROL_FRAME_PAYLOAD)
```
gxugxu, line 581: If `MAX_CONTROL_FRAME_PAYLOAD` is defined as a MACRO, we should change it to `constexpr size_t`; if not, we should not use ALL_CAP_NAME, as ALL_CAP_NAME is reserved for MACRO (suggested by standard [coding guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es9-avoid-all_caps-names)).

gxugxu: Since this is not introduced in this PR, we can add issue for this and supply separated fix.

src/cpp/scaler/ymq/internal/websocket_stream.h
```diff
@@ -62,25 +67,40 @@ class WebSocketStream {
...
 
     struct State {
         scaler::wrapper::uv::TCPSocket _socket;
-        bool _isServer;
+        Role _role {Role::Undefined};
+
+        // Used only during the HTTP Upgrade handshake, reset once completed.
+        std::optional<HandshakeDoneCallback> _upgradeCallback {};
+
         std::vector<uint8_t> _recvBuffer {};
```
gxugxu, line 79: Don't we have an universal `typedef` for buffers? Should we have one?

gxugxu: If we don't have it, then don't bother changing it. If we already have it, then we should use that.

src/cpp/scaler/ymq/internal/websocket_stream.cpp
rafa-be: Done. I replaced it by an `assert()`.

src/cpp/scaler/ymq/internal/websocket_stream.h
rafa-be: Not really.

It's also not really easy to find a common type: depending on use, some are statically allocated (`std::array<uint8_t>`), dynamically allocated (`std::vector`) or just weakly referenced (`std::span`).

gxugxu: good; make sense.

It's okay to have it as is.

gxugxu (approved):

gxugxu (approved):

PR #878, opened 2026-07-08
Disallow empty string to be key name

### LINKED ISSUE #871: Empty string is a valid capability name
The current capability parser allows empty string names/key for capabilities:

```python
In [1]: from scaler.config.types.worker import WorkerCapabilities

In [2]: WorkerCapabilities.from_string('gpu,=1')
Out[2]: WorkerCapabilities(capabilities={'gpu': -1, '': 1})
```

Empty string capabilities names don't make sense and should raise an exception that propagates back to the entry point.

gxugxu:
# READY FOR REVIEW.

Closes https://github.com/finos/opengris-scaler/issues/871


src/scaler/config/types/worker.py
```diff
@@ -44,14 +44,19 @@ def from_string(cls, value: str) -> Self:
         if not value:
             return cls(capabilities)
         for item in value.split(","):
-            name, _, value = item.partition("=")
-            if value != "":
+            name, separator, capability_value = item.partition("=")
```
magniloquency, line 47: str.partition seems to behave a bit weirdly with empty strings, maybe we could use `.split("=", 1)` instead and then check the length of the returned list? or we could use iterable unpacking.

The code as-is will accept `"a="` so we should either do one of those or find a way to guard against that case

magniloquency (changes requested):

src/scaler/config/types/worker.py
gxugxu: > The code as-is will accept "a=" so we should either do one of those or find a way to guard against that case

Accept how? This code will NOT accept `"a="` because the conversion follows it will raise.

rafa-be (approved):

rafa-be (changes requested):
Please add a test case in `tests/config/test_config_types.py`

rafa-be (dismissed):

src/scaler/version.txt
```diff
@@ -1 +1 @@
-2.5.8
+2.6.0
```
magniloquency: should be a patch version bump because this is a bug fix?

magniloquency (changes requested):

src/scaler/version.txt
gxugxu: Depends on how you see it. I see it as an API change, as we didn't disallow it explicitly. Had I have control of this, I would bump the major version.

gxugxu: @rafa-be @e117649 give some comments here. thanks

rafa-be: I agree with @gxugxu and always wanted to follow a [more usual versioning standard](https://semver.org/):

```<Major>.<Minor>.<Patch>```

With `Major` for breaking changes, `Minor` for new features and `Patch` for non-breaking, no-feature changes.

As these changes are technically breaking, it would require a `Major` bump. 

However, we could argue that we never intended to support empty string capabilities, and that these changes should be a patch as it was a bug. We never documented we accepted empty capabilities so it's not technically breaking.

I don't really see how we could justify a `Minor` version bump for these changes though.

magniloquency: I agree with this, I like semver and I think we should use it. In this case I think because this was always a buggy behaviour we can justify a patch bump.

One other question, when do we want to perform bumps? In feature PRs, just before cutting a release, or perhaps another time? In my opinion, it's best to bump once right before release so that we don't create versions that never get released.

rafa-be: > One other question, when do we want to perform bumps? In feature PRs, just before cutting a release, or perhaps another time? In my opinion, it's best to bump once right before release so that we don't create versions that never get released.

The problem with bumping versions on release is that the person making the release will have to look at all commits to determine it's a major, minor or patch release.

Thus I think it might actually be better to bump the versions on commit, as the person pushing the changes is the most able to determine if the changes are breaking or not.

e117649: I agree with @rafa-be that since we never documented this behaviour, and it is fairly minor, it could be argued as a patch. 

I agree with with @magniloquency that it would be better to bump before release. This avoids skipping versions, and having to reapprove PRs on fixing a merge conflict due to version. A couple solutions could be adding [feat] or [breaking] to the title of your PR if the next release should bump minor or major respectively.

We could also leave open a PR for the next release that just bumps the version where any of us can comment if it needs to be escalated from patch to minor or major.

magniloquency: > This avoids [...] having to reapprove PRs on fixing a merge conflict due to version.

This is a really nice benefit.

> A couple solutions could be adding [feat] or [breaking] to the title of your PR if the next release should bump minor or major respectively.

I could get on board with this, or we could even call it [major], [minor], or [patch].

> We could also leave open a PR for the next release that just bumps the version where any of us can comment if it needs to be escalated from patch to minor or major.

This also works.

gxugxu: > I don't really see how we could justify a Minor version bump for these changes though.

Mostly established convention. I was told to bump Minor when Major should be bumped.

We need to have clear definition on when to update Major.

Now, since you guys have consensus of bumping patch, I have revised the code.

magniloquency (approved):

src/scaler/version.txt
rafa-be: > I agree with with @magniloquency that it would be better to bump before release. This avoids skipping versions, and having to reapprove PRs on fixing a merge conflict due to version. A couple solutions could be adding [feat] or [breaking] to the title of your PR if the next release should bump minor or major respectively.

I'm fine with this. Maybe `feat:`, `fix:` and something else for breaking changes. 

> We need to have clear definition on when to update Major.

Easy: we are introducing breaking changes with what a documented and user-facing API. 

Your changes are user-facing, but the empty-string capability hasn't never been documented, and is more of a bug than a regression.

Issue #869, opened 2026-07-03
OpenSSL issue when the cluster is started through `SchedulerClusterCombo`


rafa-be:
There is an issue when the cluster is started through `SchedulerClusterCombo` with a TLS address on Linux.

It seems like OpenSSL has issues when ran inside multiple "spawn" Python processes.

It looks like this might be that `_ymq.so` and object_storage_server.so each statically link their own copy of OpenSSL and export the OpenSSL symbols with default (global) visibility. When both are loaded into one process, ELF symbol interposition makes one module's SSL_* calls resolve into the other module's OpenSSL build/global state, corrupting the server-side TLS state machine → EPROTO on handshake.

The OSS server process always hits this: ObjectStorageServerProcess (used by the combo and the scaler / scaler_object_storage_server entry points) imports scaler.cluster.object_storage_server → scaler.io.network_backends → scaler.io.ymq._ymq, so _ymq.so gets loaded alongside object_storage_server.so.

In `ScalerPythonModule.cmake`
```
@@ -1,3 +1,5 @@
+set(_SCALER_PYMOD_CMAKE_DIR "${CMAKE_CURRENT_LIST_DIR}")
+

@@ -93,6 +95,11 @@ function(scaler_add_python_module)
         target_link_libraries(${PYMOD_TARGET} PRIVATE ${PYMOD_LINK_LIBRARIES})
     endif()
 
+    if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+        target_link_options(${PYMOD_TARGET} PRIVATE
+            "LINKER:--version-script=${_SCALER_PYMOD_CMAKE_DIR}/scaler_pymod_exports.map")
+    endif()
+
...

_Originally posted by @e117649 in https://github.com/finos/opengris-scaler/issues/842#issuecomment-4877547237_

PR #880, opened 2026-07-09
Build OpenSSL from source in launchpad provisioner script

magniloquency:
## Summary
- The launchpad's generated cloud-init script fails to install the scheduler package from a git ref: CMake requires a statically-linked OpenSSL (`OPENSSL_USE_STATIC_LIBS`), but AL2023's `openssl-devel` package only ships `libssl.so`. Its `OpenSSLConfig.cmake` still points `OpenSSL::SSL` at a static `libssl.a` path unconditionally, so ninja fails at link time with `libssl.a ... missing and no known rule to make it`.
- Build and install OpenSSL from source in the provisioner script (same as CI's `.github/actions/3rd-party-libraries-*` steps and `scripts/library_tool.sh`), so a real static `libssl.a` exists under `/usr/local` before the pip build runs.
- Also added `perl` to the `dnf install` line, required by OpenSSL's `Configure` script.


docs/html_extra/launchpad/provisioner.js
```diff
@@ -472,7 +472,8 @@ function buildUserData(cfg, creds) {
       : `git clone --depth 1 ${cloneUrl} /opt/scaler-src`;
 
     gitBuildLines = `# C++ build deps: GCC 14 (required for C++23 <expected>) + Cap'n Proto toolchain
-dnf install -y git gcc14 gcc14-c++ gcc14-libstdc++-devel autoconf automake libtool libuv-devel openssl-devel
+# perl is needed to run OpenSSL's Configure script when building it from source below.
```
gxugxu: remove this comment

docs/html_extra/launchpad/provisioner.js
```diff
@@ -482,6 +483,13 @@ cd /opt/scaler-src
 CC=/usr/bin/gcc14-gcc CXX=/usr/bin/gcc14-g++ bash scripts/library_tool.sh capnp download
 CC=/usr/bin/gcc14-gcc CXX=/usr/bin/gcc14-g++ bash scripts/library_tool.sh capnp compile
 bash scripts/library_tool.sh capnp install
+
+# CMake requires a statically-linked OpenSSL (OPENSSL_USE_STATIC_LIBS); AL2023's
```
gxugxu: Change this three line comment into:
`# compile openssl because we need static library`

gxugxu:
@rafa-be is it established that we use static library as much as possible? It feels to me that dynamic library is the right way to go, specifically to `OpenSSL`. Though, I have no intention on switching to using dynlib here.

@magniloquency my understanding is that all dependencies need to be installed for launchpad? What about `libuv`?

rafa-be:
> @magniloquency my understanding is that all dependencies need to be installed for launchpad? What about libuv?

Looks like it uses `libuv-devel`:

```
dnf install -y git gcc14 gcc14-c++ gcc14-libstdc++-devel autoconf automake libtool libuv-devel openssl-devel perl
```

rafa-be:
> @rafa-be is it established that we use static library as much as possible? It feels to me that dynamic library is the right way to go, specifically to OpenSSL. Though, I have no intention on switching to using dynlib here.

Yeah I'm wondering that too too.

I had/am having a lot of issues with statically linked OpenSSL.

While I never had any issue with static libuv, OpenSSL seems to have some global state that does not like to have multiple instances of the static library loaded at runtime (see #869).

If these changes from @magniloquency work, I'd suggest we keep them for now, until I investigate #869 more and maybe move to dynamic libraries.

rafa-be (dismissed):

gxugxu:
> > @rafa-be is it established that we use static library as much as possible? It feels to me that dynamic library is the right way to go, specifically to OpenSSL. Though, I have no intention on switching to using dynlib here.
> 
> Yeah I'm wondering that too too.
> 
> I had/am having a lot of issues with statically linked OpenSSL.
> 
> While I never had any issue with static libuv, OpenSSL seems to have some global state that does not like to have multiple instances of the static library loaded at runtime (see #869).
> 
> If these changes from @magniloquency work, I'd suggest we keep them for now, until I investigate #869 more and maybe move to dynamic libraries.

I see. I missed the `libuv-devel` here. Yes, I agree. This PR looks good to me now.

docs/html_extra/launchpad/provisioner.js
magniloquency: done

docs/html_extra/launchpad/provisioner.js
magniloquency: done

gxugxu (approved):

PR #881, opened 2026-07-09
fix: launchpad improvements

magniloquency:
Live: https://magniloquency.github.io/scaler/launchpad/

## Summary
- Collapse the deployment sidebar column when there's nothing to show, and add a download button for scheduler logs
- Make the deployment sidebar the clear "where do I connect" answer: hero the scheduler address with a "connect your client to this address" note, and tuck the rest of the connection details (object storage, monitor, SSH, instance id) behind a collapsed "More deployment details" toggle
- Strip the bare public/private IP and VPC/CIDR/subnet lines out of the deploy terminal log (not usable connection strings, just noise) and point users at the sidebar once deployment completes
- Minor polish: fix "openGRIS" -> "OpenGRIS" casing throughout the launchpad UI, and clean up label separators / drop the redundant log path text from the scheduler log terminal header


rafa-be (dismissed):

gxugxu (dismissed):

docs/html_extra/launchpad/provisioner.js
```diff
@@ -1180,6 +1176,7 @@ async function provision(
   addLog("─".repeat(52), "dim");
   addLog("  DEPLOYMENT COMPLETE", "done");
   addLog("─".repeat(52), "dim");
+  addLog("  → Your connection details are in the panel on the right →", "addr");
```
gxugxu: Is this printed directly on screen for user to see? If so, would the visual effect looks bad if we use '->' instead?

gxugxu (approved):

e117649 (approved):

PR #882, opened 2026-07-09
fix: Improve "Installation" section of documentation

gxugxu:
# READY FOR REVIEW.

This PR makes a few improvements to "Installation" section of documentation.

- Commands with dependencies (e.g. `uv pip install opengris-scaler[gui]`) did not work due to the fact that `[]` participates in shell expansion. Fixed.
- The documentation was incomplete due to the fact that it did not instruct user to install `uv`. Fixed by adding a step that instruct the user to do so.
- Added some notes and hints in this section.
- Tell user to restart their shell if they are installing `uv` for the first time.

Corresponding page now looks like this:
image

Reviewers may view it [here](https://gxugxu.github.io/opengris-scaler/tutorials/installation.html).


docs/source/tutorials/installation.rst
```diff
@@ -5,6 +5,54 @@ Installation
 
 The ``opengris-scaler`` package is available on PyPI and can be installed using any compatible package manager. The examples below use `uv <https://docs.astral.sh/uv/getting-started/installation>`_.
 
+Install uv and Create a Virtual Environment
+-------------------------------------------
+
+.. note::
+
+    Skip this section if you already have a uv virtual environment prepared.
```
magniloquency, line 13: I think we can remove this part

```suggestion
```

docs/source/tutorials/installation.rst
```diff
@@ -5,6 +5,54 @@ Installation
...
+
+            curl -LsSf https://astral.sh/uv/install.sh | sh
+
+    .. group-tab:: Windows
+
+        .. code-block:: powershell
+
+            powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
+
+Restart your shell. Then create and activate a virtual environment named ``opengris-scaler-venv``, pinned to Python 3.13, the newest version supported by all optional dependencies (Scaler itself supports Python 3.10 or later). uv downloads this Python version automatically if it is not already installed:
```
magniloquency: Scaler works on Python 3.14 again now that we pinned oci to an old version. I think the "." between "shell" and "then" should be a comma

magniloquency (dismissed):
Overall looks good, just one thing I think we can remove and a minor change

docs/source/tutorials/installation.rst
gxugxu: Well, there's a comment in `pyproject.toml`:
```
    # graphblas need compile on 3.14, and this python condition prevents up from building 3.14 wheel, comment it out for now
    # "opengris-scaler[graphblas]; python_version <= '3.13'",
```
I opened an issue to track this https://github.com/finos/opengris-scaler/issues/886, won't fix in this PR as it's out of scope.

gxugxu: The grammar problem was fixed.

gxugxu: And specifically, if problem in #886 still exists, using python with 3.14 will NOT work when user tries to install `opengris-scaler[graphblas]`.

docs/source/tutorials/installation.rst
gxugxu: No. I follow the instruction on new box and old box. This needs to be pointed out.

rafa-be (approved):

PR #883, opened 2026-07-10
fix: task orphaning when a balance-cancel targets a worker that dies mid-cancel

magniloquency:
## Summary

When the scheduler load-balances a task off a worker, it sends a `TaskCancel` and immediately deregisters the task from that worker in the policy controller, before the worker ever confirms the cancel. If the worker dies while the cancel is in flight (e.g. its EC2 instance is terminated by an autoscaling worker manager mid-rebalance), the task is no longer tracked under any worker, so the normal worker-disconnect timeout sweep (`remove_worker`) can never find it to fail/reroute it. The task is left permanently stuck in `balanceCanceling`.

This was confirmed against a scale-up/scale-down test: after balance-canceling, ~95% of the canceled tasks never recovered and were still stuck after 9+ minutes.

## Fix

- `worker_controller.py`: `on_task_cancel` now does a non-destructive `get_worker_by_task_id` lookup instead of `remove_task`, so the task stays registered to its worker while the cancel is pending.
- `task_controller.py`: on a *confirmed* balance-cancel, we now explicitly deregister the task (`on_task_done`) right before reassigning it to a new worker, so it doesn't short-circuit back onto the worker it was just balanced away from.

With this, if the worker dies before confirming, the existing disconnect-timeout sweep still finds and properly fails/reroutes the task instead of orphaning it.


gxugxu (commented):
Were you able to produce an test case that goes wrong without this fix and goes correct with this fix? If so, consider adding it.

gxugxu (commented):
version number

magniloquency:
> Were you able to produce an test case that goes wrong without this fix and goes correct with this fix? If so, consider adding it.
@gxugxu 

I was able to verify the fix  by running scaler in the launchpad but I didn't create a unit test. I'll investigate that next week

rafa-be (approved):

e117649 (approved):

PR #885, opened 2026-07-10
feat: Scaler supports secure WebSockets (`wss://`)

### LINKED ISSUE #844: YMQ WebSocket don't support OpenSSL
The current `WebSocketStream` class in YMQ does not allow secure WebSockets (`wss://`).

This is annoying as web browsers usually prevent WebSockets to be used on non-local hosts if these don't use TLS.

YMQ already has transport layer support for OpenSSL (`SecureSocket` and `SecureServer`).

There are two possible implementations:

- hard code the use of `SecureSocket` instead of `TCPSocket` within `WebSocketStream` if the `secure` flag is set. `SecureSocket` and `TCPSocket` share a very similar API and behavior, so that shouldn't be hard.
-  a more scalable approach by making `WebSocketStream` accept an abstract socket type, as proposed in #843. But that's more work.

rafa-be:
Adds Scaler support for `wss://` addresses.

Most of the changes are related to making the `WebSocketStream` class support either `TCPSocket` or `SecureSocket` as the underlying transport layer (through a `std::variant`):

```diff
-    // Returns the underlying TCP socket.
-    scaler::wrapper::uv::TCPSocket& transport() noexcept;
+    using Transport = std::variant<scaler::wrapper::uv::TCPSocket, scaler::wrapper::openssl::SecureSocket>;
+    
+    // Returns the underlying transport socket.
+    Transport& transport() noexcept;
```

Basic usage:

```bash
# OSS
$ scaler_object_storage_server wss://127.0.0.1:5264 --tls-cert sample_cert.pem --tls-key sample_private_key.pem

# Scheduler
$ scaler_scheduler wss://127.0.0.1:5263 -osa wss://127.0.0.1:5264 --tls-cert sample_cert.pem --tls-key sample_private_key.pem

# Workers
$ scaler_worker_manager baremetal_native wss://127.0.0.1:5263 -wmi worker-manager-1

# Client
$ ipython
Python 3.12.4 (main, Jun  7 2024, 04:28:19) [Clang 14.0.0 (clang-1400.0.29.202)]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.12.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: Use `F2` or %edit with no arguments to open an empty editor with a temporary file.

In [1]: from scaler import Client

In [2]: c = Client("wss://127.0.0.1:5263")

In [3]: c.submit(round, 3.14).result()
Out[3]: 3
```


src/cpp/scaler/wrapper/openssl/secure_socket.cpp
rafa-be, line 1: Enabling TLS for WebSocket unveiled a re-entrant issue in `SecureSocket`:

The class was calling its `WriteCallback` before `reset()`-ing its value. The registered callback on the upper layer `WebSocket` called `write()` itself during the WS handshake, that would trigger a re-entrant issue with the original callback being called twice.

Now, all callbacks are `move()`-d and `reset()` before being called, ensured there cannot be more than one call.

gxugxu (dismissed):

magniloquency (approved):

gxugxu (approved):

PR #894, opened 2026-07-14
fix: Cap max_task_concurrency with upper limit 2^32-1

### LINKED ISSUE #696: `max_task_concurrency=-1` (unlimited) cannot be serialized over Cap'n Proto
## Summary

`WorkerManagerConfig.max_task_concurrency = -1` is the Python-level sentinel for \"no concurrency limit\". The scaling policies in `VanillaScalingPolicy` and `CapabilityScalingPolicy` guard on `max_concurrency != -1` to skip the capacity check.

However, the `maxTaskConcurrency` field in the `WorkerManagerHeartbeat` Cap'n Proto message is typed as `UInt32`, which cannot represent -1. Attempting to serialize a heartbeat with `max_task_concurrency=-1` raises a `KjException` at the Cap'n Proto layer:

```
capnp.lib.capnp.KjException: failed: expected value >= 0 && T(value) == value;
Value out-of-range for requested type.; value = -1
```

This means:
- Any worker manager configured with `max_task_concurrency=-1` crashes on its first heartbeat.
- The `!= -1` branches in both `VanillaScalingPolicy._create_start_commands` and `CapabilityScalingPolicy._create_start_command` are unreachable via the wire protocol.

## Suggested fix

Pick a wire-safe sentinel and apply it consistently:

- **0** — natural "no limit" value; update the guard from `!= -1` to `!= 0` everywhere
- **`0xFFFFFFFF`** — keeps the `-1`-like semantics; convert on serialise/deserialise in the Python wrapper
- **Separate boolean field** `unlimited: Bool` in the Cap'n Proto schema — most explicit, no magic numbers

The Python config layer can continue to accept `-1` as user-facing input; the conversion just needs to happen before the value is written into a Cap'n Proto message.

gxugxu:
# READY FOR REVIEW.

fix: Cap max_task_concurrency with upper limit 2^32-1.
closes #696.

- Note that this changes the limit of possible `max_task_concurrency`.
- A constant is added to `defaults.py`. This follows existing convention (e.g. `CAPNP_MESSAGE_SIZE_LIMIT`)
- Simplifies calculation in some worker managers with the new limit (in contrast to -1).


src/scaler/worker_manager_adapter/orb_aws_ec2/worker_manager.py
```diff
@@ -423,7 +423,7 @@ def _create_user_data(self) -> str:
 
 """
 
-        # --max-task-concurrency is not passed: scaler_worker_manager defaults to cpu_count - 1 workers,
+        # --max-task-concurrency is not passed: scaler_worker_manager defaults to cpu_count workers,
```
gxugxu, line 431: leftover across changes

docs/source/tutorials/worker_managers/common_parameters.rst
```diff
@@ -12,7 +12,7 @@ Networking
...
+* ``--max-task-concurrency`` (``-mtc``): Maximum total number of workers that can be active across all spawned instances or containers (default: number of CPUs). Must be an integer between ``0`` and ``4294967295`` (``2^32 - 1``): ``0`` means this machine or manager starts no workers at all, and ``2^32 - 1`` is the upper limit, which is large enough to act as an effectively unlimited capacity. Negative values (such as the former ``-1`` "no limit" sentinel) are rejected. For worker managers that spawn multi-worker machines (e.g. ORB AWS EC2, AWS Raw ECS), the number of instances spawned is ``ceil(max_task_concurrency / workers_per_instance)``, so the actual worker count may exceed ``max_task_concurrency`` by up to ``workers_per_instance - 1`` due to rounding up to the nearest whole instance.
```
magniloquency: small change, i think putting `2^32 - 1` is enough, and I don't think we should mention that `-1` was previously unlimited. We could put that in the changelog, but eventually that would make the doc out of date?

```suggestion
* ``--max-task-concurrency`` (``-mtc``): Maximum total number of workers that can be active across all spawned instances or containers (default: number of CPUs). Must be an integer between ``0`` and ``2^32 - 1``: ``0`` means this machine or manager starts no workers at all, and ``2^32 - 1`` is the upper limit, which is large enough to act as an effectively unlimited capacity. Negative values are rejected. For worker managers that spawn multi-worker machines (e.g. ORB AWS EC2, AWS Raw ECS), the number of instances spawned is ``ceil(max_task_concurrency / workers_per_instance)``, so the actual worker count may exceed ``max_task_concurrency`` by up to ``workers_per_instance - 1`` due to rounding up to the nearest whole instance.
```

magniloquency: One more thing about this, I suppose that:

> so the actual worker count may exceed ``max_task_concurrency`` by up to ``workers_per_instance - 1`` due to rounding up to the nearest whole instance.

is not entirely accurate, because if `max_task_concurrency == 2**32 - 1`, then it'll be exactly equal? Might be worth calling out

magniloquency (changes requested):
Two minor docs comments, otherwise looks good

docs/source/tutorials/worker_managers/common_parameters.rst
gxugxu: 4294967295

is what's reported to the user upon raised exception wrt this.

gxugxu: but okay, I changed it

gxugxu: well it may or may not, right? And why would it be exactly equal? launching instances with 64 cores will end up having 63 more workers than 2^32-1.

either way, nobody can launch that much workers.

magniloquency (dismissed):
If we now reject `-1` this is technically a breaking change. Should we maintain backwards compatibility?

PR #900, opened 2026-07-20
fix: Scheduler and worker fault tolerance around abrupt peer death

### LINKED ISSUE #921: Worker agent is destroyed when a processor dies while suspended (nested tasks / graph mode)
## Summary

When a processor dies while it is **suspended**, the worker agent that owns it is destroyed: it
either crashes with an unhandled `OverflowError` out of its heartbeat loop, or its event loop
deadlocks forever inside `ProcessorHolder.resume()`. Either way the worker is gone for good — it is
never restarted, and it frequently does not even exit, leaving a live-but-useless process behind.

A processor is only ever suspended when the task running on it submits **nested tasks**
(`VanillaTaskManager.__suspend_if_priority_is_higher`). That is exactly what a graph node that does
its own parallel work looks like — a `pargraph` node opening a nested `Client` — which is why this
only shows up in graph/nested workloads and never with plain `client.submit`.

In production the killer is the pod's cgroup memory limit: suspended parents are the fattest
processes in the pod (they hold the parent stage's data while their children run), so they are what
the kernel OOM killer picks.

## Environment

- opengris-scaler `main` @ `b777560` (2.6.0), verified with a clean checkout
- Python 3.13.14, Linux (Ubuntu 26.04, cgroup v2)
- YMQ network backend (default), `hard_processor_suspend=False` (default)
- Also reproduced with workers in a memory-limited cgroup and the scheduler/object storage/client
  outside it, mirroring workers-in-an-EKS-pod with everything else off-cluster

## Impact

- A single OOM-killed (or otherwise dying) processor takes down the **whole worker**, including the
  healthy processors next to it.
- The worker is never replaced: `NativeWorkerProvisioner.run_fixed` logs `exited unexpectedly` and
  does not respawn, so the pod silently loses capacity until it is restarted.
- The dead worker's process usually does not exit (see steps 3-4 below), so `ps` shows a worker that
  looks alive, does nothing, holds its memory, and keeps its slot.
- With several workers per pod, one OOM cascades: each dead worker leaves its processors behind,
  which raises pod memory, which triggers the next OOM.

## Reproduction

Self-contained script — needs `opengris-scaler`, `pargraph` and `psutil`, nothing else. It starts a
2-worker `SchedulerClusterCombo`, submits a pargraph graph whose nodes each open a nested client,
and SIGKILLs one processor at the moment the agent logs that it suspended it.

```console
$ python issue_repro.py crash
[repro] variant=crash worker agents=[168939, 168940]
[repro] SIGKILL suspended Processor[168969]
[repro] graph failed after 1.4s: ProcessorDiedError("process_status='zombie'")
[repro] FAIL: a worker agent died of an unhandled exception:
[repro]        ERROR scaler[168940]: WorkerID(Worker|FIX|c9d54a69...): failed with unhandled exception:
[repro]        OverflowError: can't convert negative int to unsigned
[repro] FAIL: a worker agent's teardown aborted (binder never destroyed)
[repro] worker destroyed by the death of a suspended processor (see scaler_repro.log)
$ echo $?
1
```

**Control:** SIGKILL a processor that is *running* a task, with no nesting (plain `client.submit`) —
the agent restarts the processor, reports `ProcessorDiedError` for that one task, and keeps serving.
Nothing is wedged. The difference is purely whether the dead processor was suspended.

<details>
<summary><code>issue_repro.py</code></summary>

```python
"""Minimal reproduction: a scaler worker agent is destroyed when one of its processors dies
while that processor is SUSPENDED.

A processor is only ever suspended when the task running on it submits nested tasks -- which is
what a graph node doing its own parallel work does (pargraph node -> nested Client). Plain
`client.submit` never suspends a processor, so it never hits this.

The death here is a SIGKILL, which is what a pod that hits its memory limit gets from the kernel
OOM killer: the suspended parents are the fattest processes in the pod, because they hold the
parent stage's data while their children run.

    pip install opengris-scaler pargraph psutil
    python issue_repro.py crash      # sub-tasks are slow: the parent stays suspended
    python issue_repro.py deadlock   # sub-tasks are fast: the parent is resumed right after

Exit code 0 = worker survived, 1 = worker destroyed. Scaler's own logs go to scaler_repro.log.
"""

import os
import re
import signal
import sys
import threading
import time
from typing import List, TextIO

import psutil
from pargraph import delayed, graph

from scaler import Client, SchedulerClusterCombo

N_WORKERS = 2
N_NODES = 4
FAN_OUT = 3
LOG_PATH = "scaler_repro.log"
SUSPEND_RE = re.compile(r"suspend Processor\[(\d+)\]")

VARIANT = sys.argv[1] if len(sys.argv) > 1 else "crash"
LEAF_SECONDS = 3.0 if VARIANT == "crash" else 0.0


def leaf(index: int, seconds: float) -> int:
    time.sleep(seconds)
    return index


def nested_stage(index: int, seconds: float) -> int:
    """Runs inside a processor. Submitting nested tasks makes the worker agent suspend this
    processor and start a fresh one to run them."""
    with Client() as client:
        futures = [client.submit(leaf, index * 10 + i, seconds) for i in range(FAN_OUT)]
        return sum(future.result() for future in futures)


@delayed
def stage(index: int, seconds: float) -> int:
    return nested_stage(index, seconds)


@delayed
def reduce_all(*parts: int) -> int:
    return sum(parts)


@graph
def pipeline(seconds: float, n_nodes: int) -> int:
    return reduce_all(*(stage(index, seconds) for index in range(n_nodes)))


def worker_agents() -> List[psutil.Process]:
    """Process tree is: this process -> worker manager -> worker agents -> processors."""
    for child in psutil.Process().children():
        agents = child.children()
        if agents and any(agent.children() for agent in agents):
            return agents
    return []


def kill_first_suspended_processor(console: TextIO, deadline: float) -> None:
    """The agent logs `suspend Processor[<pid>]` the moment it suspends one. Kill that pid."""
    with open(LOG_PATH, "r") as log_file:
        log_file.seek(0, os.SEEK_END)
        while time.time() < deadline:
            line = log_file.readline()
            if not line:
                time.sleep(0.01)
                continue
            match = SUSPEND_RE.search(line)
            if match is None:
                continue
            pid = int(match.group(1))
            try:
                os.kill(pid, signal.SIGKILL)
            except ProcessLookupError:
                continue
            print(f"[repro] SIGKILL suspended Processor[{pid}]", file=console, flush=True)
            return


def running(pid: int) -> bool:
    return psutil.pid_exists(pid) and psutil.Process(pid).status() != psutil.STATUS_ZOMBIE


def cleanup() -> None:
    for process in psutil.Process().children(recursive=True):
        try:
            process.kill()
        except psutil.Error:
            pass


def main() -> int:
    # scaler's processes log to stdout/stderr; point those at a file that this script can read,
    # both to drive the kill and to check the outcome
    log_file = open(LOG_PATH, "w+", buffering=1)
    console = os.fdopen(os.dup(1), "w", buffering=1)
    os.dup2(log_file.fileno(), 1)
    os.dup2(log_file.fileno(), 2)

    def say(message: str) -> None:
        print(message, file=console, flush=True)

    cluster = SchedulerClusterCombo(n_workers=N_WORKERS)

    deadline = time.time() + 30
    agents: List[int] = []
    while time.time() < deadline and len(agents) < N_WORKERS:
        agents = [agent.pid for agent in worker_agents() if agent.children()]
        time.sleep(0.5)

    if len(agents) < N_WORKERS:
        say(f"[repro] cluster did not come up: {agents}")
        cleanup()
        return 2

    say(f"[repro] variant={VARIANT} worker agents={agents}")

    threading.Thread(target=kill_first_suspended_processor, args=(console, time.time() + 90), daemon=True).start()

    graph_dict, keys = pipeline.to_graph(n_nodes=N_NODES).to_dict(seconds=LEAF_SECONDS)
    started = time.time()
    with Client(address=cluster.get_address()) as client:
        try:
            say(f"[repro] graph result: {client.get(graph_dict, keys)}")
        except Exception as error:  # noqa: BLE001 - the failure is the point
            say(f"[repro] graph failed after {time.time() - started:.1f}s: {error!r}")

    # give the agent's heartbeat loop a few more ticks with the stale bookkeeping in place
    time.sleep(15)

    # shutdown runs off-thread: it joins the workers, and a wedged one never returns
    threading.Thread(target=cluster.shutdown, daemon=True).start()

    deadline = time.time() + 30
    wedged = list(agents)
    while time.time() < deadline and wedged:
        wedged = [pid for pid in wedged if running(pid)]
        time.sleep(0.5)

    log_text = open(LOG_PATH).read()
    crashed = "failed with unhandled exception" in log_text
    teardown_aborted = "teardown failed" in log_text

    if crashed:
        say("[repro] FAIL: a worker agent died of an unhandled exception:")
        for line in log_text.splitlines():
            if "OverflowError" in line or "failed with unhandled exception" in line:
                say(f"[repro]        {line.strip()[:140]}")
    if teardown_aborted:
        say("[repro] FAIL: a worker agent's teardown aborted (binder never destroyed)")
    if wedged:
        say(f"[repro] FAIL: worker agent(s) {wedged} ignored shutdown and never exited")
        say("[repro]        inspect with: py-spy dump --pid " + " --pid ".join(str(pid) for pid in wedged))

    cleanup()

    if crashed or teardown_aborted or wedged:
        say(f"[repro] worker destroyed by the death of a suspended processor (see {LOG_PATH})")
        return 1

    say("[repro] PASS: every worker agent survived and shut down")
    return 0


if __name__ == "__main__":
    sys.exit(main())
```

</details>

### 1. The dead processor is left in `_suspended_holders_by_task_id`

`VanillaProcessorManager.on_failing_processor` (`src/scaler/worker/agent/processor_manager.py:151`)
removes the holder from `_holders_by_processor_id` (via `__kill_processor`) but never from
`_suspended_holders_by_task_id`:

```python
async def on_failing_processor(self, processor_id: ProcessorID, process_status: str):
    holder = self._holders_by_processor_id.get(processor_id)
    if holder is None:
        return

    task = holder.task()
    if task is not None:
        profile_result = self.__end_task(holder)   # clears holder.task()
    else:
        profile_result = None

    reason = f"process died {process_status=}"
    if holder == self._current_holder:
        self.__restart_current_processor(reason)
    else:
        self.__kill_processor(reason, holder)      # pops _holders_by_processor_id only
    ...
```

`_suspended_holders_by_task_id` is only ever popped in `on_cancel_task`, `on_resume_task` and
`on_task_result`. A processor that dies while suspended leaves a permanent entry pointing at a dead
process.

### 2. The stale entry makes the heartbeat send a negative count, which kills the agent

`VanillaHeartbeatManager.routine` (`src/scaler/worker/agent/heartbeat_manager.py:102`):

```python
queuedTasks=self._worker_task_manager.get_queued_size() - num_suspended_processors,
```

A suspended task is both *processing* and *queued*, so the two normally cancel out. When the
suspended processor dies, `on_failing_processor` emits its task result, which removes the task from
the queue — but `num_suspended_processors` still counts it. `queuedTasks` goes negative, and the
capnp field is unsigned (`src/protocol/message.capnp:80`, `queuedTasks @3 :UInt32`).

Worker loops are created with `swallow_peer_departed=False`, so this escapes `asyncio.gather` and
takes the whole agent down:

```
ERROR scaler[152726]: WorkerID(Worker|FIX|68d2c546...): failed with unhandled exception:
can't convert negative int to unsigned
  File "src/scaler/worker/worker.py", line 133, in _run
    await self._task
  File "src/scaler/worker/worker.py", line 314, in __main_loop
    await asyncio.gather(
  File "src/scaler/utility/event_loop.py", line 53, in loop
    await routine()
  File "src/scaler/worker/agent/heartbeat_manager.py", line 94, in routine
    await self._connector_external.send(
  File "src/scaler/io/ymq_async_connector.py", line 95, in send
    await self._socket.send_message(Bytes(serialize(message)))
  File "src/scaler/io/utility.py", line 37, in serialize
    return payload.to_bytes()
OverflowError: can't convert negative int to unsigned
```

### 3. Teardown then aborts on the same stale entry

`Worker.__teardown` calls `processor_manager.destroy()`, whose `__kill_all_processors` walks
`_suspended_holders_by_task_id` and calls `__kill_processor` on the already-killed holder →
`VanillaProfilingManager.on_process_end` (`profiling_manager.py:38`) raises:

```python
def on_process_end(self, pid: int):
    if pid not in self._process_profiler_by_pid:
        raise ValueError(f"process {pid=} is not registered.")
```

```
ERROR scaler[152726]: WorkerID(Worker|FIX|68d2c546...): teardown failed: process pid=152729 is not registered.
```

`__teardown` (`src/scaler/worker/worker.py:325`) is a flat sequence, so the raise skips everything
after it:

```python
if self._connector_external is not None:
    self._connector_external.destroy()
if self._processor_manager is not None:
    self._processor_manager.destroy("quit")     # <-- raises
if self._binder_internal is not None:
    self._binder_internal.destroy()             # <-- never runs
if self._connector_storage is not None:
    self._connector_storage.destroy()
```

### 4. The undestroyed binder hangs the process at interpreter exit

`YMQAsyncBinder.__del__` calls `destroy()` → `self._socket.shutdown()`, which blocks forever. The
agent has logged `quit`, every loop has logged `exited`, and the process still never leaves:

```
$ py-spy dump --pid 152726
Thread 152726 (idle)
    shutdown (scaler/io/ymq/sockets.py:48)
    destroy (scaler/io/ymq_async_binder.py:49)
    __del__ (scaler/io/ymq_async_binder.py:30)
```

(This last step is the one that is not 100% deterministic — sometimes the crashed agent does manage
to exit. The worker is destroyed either way.)

### Separate outcome of the same race: the agent deadlocks instead of crashing

If the parent task is resumed *before* the heartbeat notices the death, the agent instead blocks
forever in `ProcessorHolder.resume()` (`src/scaler/worker/agent/processor_holder.py:134`):

```python
self._resume_event.set()          # <-- never returns
```

`multiprocessing.Event.set()` → `Condition.notify_all()` → `notify()` waits on
`self._woken_count.acquire()` for the sleeper to acknowledge. The only sleeper was SIGKILLed inside
`Event.wait()` after registering in `_sleeping_count`, so that acknowledgement never comes:

```
$ py-spy dump --pid 150835
Thread 150835 (idle): "MainThread"
    notify (multiprocessing/synchronize.py:297)
    notify_all (multiprocessing/synchronize.py:304)
    set (multiprocessing/synchronize.py:345)
    resume (scaler/worker/agent/processor_holder.py:134)
    on_resume_task (scaler/worker/agent/processor_manager.py:233)
    __processing_task (scaler/worker/agent/task_manager.py:104)
    routine (scaler/worker/agent/task_manager.py:89)
    loop (scaler/utility/event_loop.py:53)
```

The agent's whole asyncio loop is frozen: no heartbeats, so the scheduler drops the worker after
`worker_timeout_seconds` and reroutes its tasks, while the process lives on forever. The unbounded
`self._resumed_event.wait()` on the next line has the same problem.

## Reproduced with a real OOM killer, no artificial SIGKILL

Same workload, workers in a `MemoryMax=1200M` cgroup (equivalent to a k8s memory limit), scheduler,
object storage and client outside it. Nested parallelization piles up suspended parents until the
limit is hit:

```
kernel: Memory cgroup out of memory: Killed process 50732 (python) anon-rss:274068kB
kernel: Memory cgroup out of memory: Killed process 50848 (python) anon-rss:273804kB

WARNING scaler[50729]: Processor[50732] died: process_status='zombie', task_id=...
WARNING scaler[50729]: Processor[50848] died: process_status='zombie', task_id=...
ERROR   scaler[50729]: failed with unhandled exception: can't convert negative int to unsigned
ERROR   scaler[50729]: teardown failed: process pid=50732 is not registered.
```

Both workers stopped heartbeating, the scheduler dropped them 60s later, and the pod's worker
processes were still there — alive, idle, useless.

This variant is timing-dependent (the kill has to land on a suspended processor rather than a
running one), which is presumably why it looks intermittent in production.

## Suggested fixes

1. **Drop the dead processor from the suspended map** — `processor_manager.py`:

   ```diff
        task = holder.task()
        if task is not None:
            profile_result = self.__end_task(holder)
   +        self._suspended_holders_by_task_id.pop(task.taskId, None)
        else:
            profile_result = None
   ```

2. **Never let bookkeeping produce an invalid heartbeat** — `heartbeat_manager.py`:

   ```diff
   -                queuedTasks=self._worker_task_manager.get_queued_size() - num_suspended_processors,
   +                queuedTasks=max(0, self._worker_task_manager.get_queued_size() - num_suspended_processors),
   ```

3. **Make `on_process_end` idempotent** (or stop treating a double-end as fatal) —
   `profiling_manager.py`:

   ```diff
    def on_process_end(self, pid: int):
   -    if pid not in self._process_profiler_by_pid:
   -        raise ValueError(f"process {pid=} is not registered.")
   -
   -    self._process_profiler_by_pid.pop(pid)
   +    self._process_profiler_by_pid.pop(pid, None)
   ```

4. **Make teardown best-effort** — `worker.py:__teardown`: guard each `destroy()` so one failure
   cannot skip the binder and strand the process in `__del__` at interpreter exit.

5. **Make resume survive a dead processor** — `processor_holder.py:resume`: the shared
   `multiprocessing.Event` handshake cannot be made safe with a liveness check alone, since
   `set()` itself blocks on the killed waiter's semaphore. Options: check the processor is alive and
   treat a dead one as a failed task *before* touching the event (narrows the window), bound
   `_resumed_event.wait()` with a timeout, or replace the shared-semaphore handshake with something
   a `SIGKILL` cannot poison (a pipe/socket, or the `hard_processor_suspend` SIGSTOP path).

Fixes 1-4 are small and independent; 5 is the design question.

## Related observations (not part of this bug)

- The task that was on the dead processor is reported to the client as `failed` carrying
  `ProcessorDiedError`, never retried, even though a whole-worker disconnect *does* reroute
  in-flight tasks. Retrying a worker-death failure (with a per-task cap) would make OOM survivable
  from the client's point of view.
- `NativeWorkerProvisioner` never replaces a worker that died (`run_fixed` only logs), and in
  `dynamic` mode `start_units`/`stop_units` never `join()` the processes they create/stop, so those
  do become literal `<defunct>` entries in the container.
- After the pod's agents wedged in the OOM run, the **scheduler stopped processing new peers**: the
  process stayed alive with its main thread idle in `select`, accepted TCP connections, but never
  registered a freshly started worker manager or a new client (no `worker connected` line ever, and
  clients time out after 60s). Last thing it did was reroute the dead workers' tasks. Happy to open
  that separately if it is not already known.

e117649:
A single fault can no longer take the whole cluster down
- Every scheduler loop runs with swallow_routine_errors=True: a routine that raises is logged and the loop keeps going, instead of escaping into asyncio.gather and exiting the process. Opt-in, so the client and worker agents keep crash-and-restart.
- __routing logs and drops a failed state transition rather than re-raising it into whichever handler or timer loop dispatched it.
- The main loop no longer swallows YMQException/ObjectStorageException in silence — a scheduler that stops responding now says why before it goes.

Crashes fixed at the source, not just caught
- A task in canceling whose worker disconnects reaches __state_canceled, which takes a TaskCancelConfirm, but was handed a worker_id → TypeError → scheduler exits. Still live on main: its own test_dead_worker trips it 4 runs in 5 here, then wedges.
- get_client_id raised for a task no tracked client owns — a TaskLog streamed while the task ran, arriving after it finished — from an unguarded receive path.
- A late graph subtask result or cancel-confirm indexed a graph already cleaned up → KeyError. Reachable whenever a client running a graph dies.

Spins and leaks
- The balancer re-applied balanceTaskCancel every cycle to a task already in balanceCanceling (a saturated worker is slow to confirm), logging cannot apply 8 to current state 3 forever. Now only balance-cancels a running task, once per stable piece of advice.
- A cancelled graph left one entry per node in the subtask id map for the scheduler's lifetime.
- EvenLoadAllocatePolicy.remove_worker left the departed worker in _workers_to_queue_size.
- Graph cancel and abort could busy-spin on a ready node with no task info.

Worker
- A processor that dies while suspended is detected and cleaned up; it used to crash the agent on a negative queued-task count, or deadlock it.
- Teardown destroys every resource even when one fails, and reports which one did.

Observability
- Worker disconnects log why, and the worker is removed atomically; scaling decisions log the numbers behind them; a detached send logs failures at warning unless it's just a peer leaving; the detached contract is documented on the binder/connector interfaces.

Tests: a unit test per fix, a 300-worker mass-eviction test, balance integration tests, a suspended-processor death test.

Closes https://github.com/finos/opengris-scaler/issues/921


src/scaler/scheduler/controllers/task_controller.py
```diff
@@ -450,11 +450,12 @@ async def __routing(self, task_id: TaskID, transition: TaskTransition, **kwargs)
...
-            logger.exception(
-                f"{task_id!r}: exception happened, transition: {transition} path: {state_machine.get_path()}"
-            )
-            raise e
+        except Exception:
+            # A bug in a state function must not crash the scheduler: __routing runs both from message
+            # handlers and from the balance/cleanup timer loops, so re-raising would propagate through
+            # asyncio.gather. Log the transition + state path and drop it -- the task may stall (surfaced by
+            # the client/worker timeouts), but the scheduler stays alive.
+            logger.exception(f"{task_id!r}: transition {transition} failed, path: {state_machine.get_path()}")
```
magniloquency, line 423: Is there a way to recover the state machine in this case? what if it fails repeatedly?

src/scaler/scheduler/controllers/worker_controller.py
magniloquency, line 1: follow up: a lot of these fns seem to be missing return typing

src/scaler/scheduler/controllers/worker_controller.py
```diff
@@ -166,29 +166,34 @@ def get_workers_by_manager_id(self, manager_id: bytes) -> List[WorkerID]:
...
-        workers_set.discard(worker_id)
-        if not workers_set:
-            del self._manager_to_workers[manager_id]
+        manager_id = self._worker_to_manager.pop(worker_id, None)
+        if manager_id is not None:
+            workers_set = self._manager_to_workers.get(manager_id)
+            if workers_set is not None:
+                workers_set.discard(worker_id)
+                if not workers_set:
+                    del self._manager_to_workers[manager_id]
```
magniloquency: style: It's preferable to return early instead of nest

src/scaler/scheduler/controllers/task_controller.py
e117649: good catch, it triggers when workers are CPU starved, addressed in [7255b16](https://github.com/finos/opengris-scaler/pull/900/commits/7255b163a34a617749bb93fde98fcafb8bee7c87)

src/cpp/scaler/ymq/binder_socket.cpp
```diff
@@ -223,18 +223,17 @@ void BinderSocket::onRemoteDisconnect(
...
-    if (reason != internal::MessageConnection::DisconnectReason::Disconnected) {
-        return;
-    }
-
-    // Remember the identity so any sendMessage call that lands *after* this disconnect (because
-    // Python is just catching up on messages libuv already buffered) fails fast instead of
-    // queueing forever in _pendingSendMessages and hanging the asyncio loop. The set is
-    // bounded by a TTL purge driven from here (insert-time) so it cannot grow without bound
-    // even under workloads that churn many short-lived peers (e.g. nested-task clients).
+    // Treat BOTH a graceful close (Disconnected, a clean FIN) and an abrupt drop (Aborted, a reset
```
gxugxu: The assumption was that abrupt drop will reconnect. Is it possible to make evicted pod have their dying wish to do a graceful shutdown of ymq sockets?

If that is possible then do that.

Essentially we assumed no one will do kill -9, otherwise the connection can be cut off in anytime and scheduler will suffer a lot. I think we for now should just keep it as is.

e117649: We do SIGTERM + terminationGracePeriodSeconds on orderly teardown, this is for when we can't signal (grace-expiry SIGKILL (CPU-starved worker), OOM, --grace-period=0, or hard node loss)

rafa-be: Yeah I agree with @gxugxu, this fix is wrong.

We should not return a `ConnectorSocketClosedByRemoteEnd` error on connection abort. Or at least not immediately. Otherwise it will just void the whole auto-reconnect idea. This should be fixed within Scaler, not by changing YMQ's core design.

The comment talks about a hang caused by `Binder::sendMessage()` never finishing. If that's the issue, the proper fix should be to make `sendMessage()` fire-and-forget with [`asyncio.create_task()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task), either within the Python connectors classes, or from the calling site.

src/scaler/io/ymq/utils.py
```diff
@@ -58,6 +61,49 @@ def callback(result: Union[T, BaseException]):
     return future.result(timeout)
 
 
+def run_detached(
+    awaitable: Awaitable[Any], description: str, on_done_callback: Optional[Callable[["asyncio.Task"], None]] = None
+) -> "asyncio.Task":
```
magniloquency: Can we use the type directly instead of through a string?

src/scaler/scheduler/controllers/client_controller.py
```diff
@@ -57,6 +57,12 @@ def has_client_id(self, client_id: ClientID) -> bool:
         return client_id in self._client_last_seen
 
     def get_client_id(self, task_id: TaskID) -> Optional[ClientID]:
+        # get_key raises when the task is not owned by any tracked client -- e.g. a TaskLog streamed
+        # while the task was running arrives after the task finished (mapping removed) or after the
+        # client was disconnected. Guard with has_value so callers get the documented None instead of a
+        # ValueError, which (from the unguarded on_receive_message TaskLog path) would crash the scheduler.
+        if not self._client_to_task_ids.has_value(task_id):
+            return None
         return self._client_to_task_ids.get_key(task_id)
```
magniloquency, line 62: future improvement: typically in python `[]` raises and `.get*()` returns `None` if the value is not found

src/scaler/scheduler/controllers/graph_controller.py
```diff
@@ -116,14 +119,20 @@ async def on_graph_task(self, client_id: ClientID, graph_task: GraphTask):
...
 
         # received any subtask canceling will lead the whole graph canceling
         await self.__cancel_whole_graph(graph_task_id)
 
     async def on_graph_sub_task_result(self, result: TaskResult):
+        graph_info = self.__graph_for_subtask(result.taskId)
+        if graph_info is None:
+            # A late result for a subtask whose graph has already finished/cancelled and been cleaned up.
+            # Indexing the maps directly here would KeyError, and on this path that tears the scheduler down.
+            return
```
magniloquency, line 134: presumably this task id should be removed from the mapping at some point. is that handled?

src/scaler/utility/event_loop.py
```diff
@@ -50,7 +56,13 @@ async def loop():
...
         try:
             while True:
-                await routine()
+                try:
+                    await routine()
+                except Exception:
+                    if not swallow_routine_errors:
+                        raise
+                    routine_owner = routine.__self__.__class__.__name__  # type: ignore[attr-defined]
+                    logger.exception(f"{routine_owner}: routine raised, continuing")
```
magniloquency: can we include the exception details in here?

src/scaler/client/agent/bridge.py
```diff
@@ -676,7 +676,7 @@ def identity(self) -> bytes:
     def address(self) -> Optional[AddressConfig]:
         return self._address
 
-    async def send(self, message: BaseMessage) -> None:
+    async def send(self, message: BaseMessage, *, detached: bool = True) -> None:
```
magniloquency: I would feel better about defaulting this to false

magniloquency (commented):
a few possible improvements but not blocking

src/scaler/client/agent/bridge.py
sharpener6: please don't use default value, change all the places use it and pass them explicitly, same as following

e117649:
The last 4 commits (cd016af) and later will be removed from this PR and put into a separate PR later, just pushed to this branch for now for testing

e117649:
> The last 4 commits ([cd016af](https://github.com/finos/opengris-scaler/commit/cd016af6ef8c17fc5e4198571f896f81904233a5)) and later will be removed from this PR and put into a separate PR later, just pushed to this branch for now for testing

These changes are now in https://github.com/finos/opengris-scaler/pull/927 and https://github.com/finos/opengris-scaler/pull/928

src/scaler/scheduler/controllers/client_controller.py
e117649: get_client_id already follows the .get*()-returns-None convention; what doesn't is OneToManyDict.get_key, which raises, and that's why the guard exists. Changing that class's contract reaches past this PR, agreed for a future improvement.

src/scaler/scheduler/controllers/graph_controller.py
e117649: It wasn't handled, good catch, now e51fbf6f removes a graph's subtask ids at the point the graph itself is cleaned up, which covers all three ways a graph can end. I didn't remove the id here instead, because this line only ever sees the one subtask that happens to get a late message, the rest of that graph's ids would still be left behind. The check stays as a guard, but with the mapping cleaned up properly it should no longer be reachable in normal operation. The regression test cancels a graph and asserts the mapping is empty afterwards.

src/scaler/io/mixins.py
```diff
@@ -98,6 +98,17 @@ def destroy(self):
 
     @abc.abstractmethod
     async def send(self, to: bytes, message: BaseMessage, *, detached: bool):
+        """Sends a message to one peer.
+
+        A detached send returns once the send is scheduled and logs whatever failure follows, so a peer that
+        left -- or one that never connected, which parks a send forever -- can neither raise into nor block a
+        caller that has already moved on. Ask for an undetached send only when the caller has to know the
```
magniloquency: i think one of these might read better

```suggestion
        caller that has already moved on. Ask for a non-detached send only when the caller has to know the
```

```suggestion
        caller that has already moved on. Ask for an attached send only when the caller has to know the
```

src/scaler/scheduler/controllers/policies/simple_policy/scaling/capability_scaling.py
```diff
@@ -31,8 +38,32 @@ def get_scaling_commands(
...
+            frozenset(capabilities.items()): desired for capabilities, desired in desired_per_capset
+        }
+
+        for capabilities, desired in desired_per_capset:
+            if previously_logged.get(frozenset(capabilities.items())) == desired:
+                continue
+
+            task_count = len(tasks_by_capability[frozenset(capabilities.keys())])
+            logger.info(f"scaling {manager_id!r}: tasks={task_count}, capabilities={capabilities} -> desired={desired}")
+
```
magniloquency, line 74: I really like having better visibility into this. Can we create some future work items to see if we can somehow refactor implementations to use a common component (DRY)? It would also be great if we could get this data to the webui somehow so that it can be visualized.

magniloquency, line 74: Do we have some kind of limit on the size of the dict? If it's unbounded could it eventually cause OOM?

src/scaler/scheduler/controllers/task_controller.py
```diff
@@ -180,6 +186,17 @@ async def on_worker_connect(self, worker_id: WorkerID):
...
+        state_machine = self._task_state_manager.get_state_machine(task_id)
+        if state_machine is not None and state_machine.current_state() == TaskState.canceling:
+            # canceling routes workerDisconnect to canceled, whose handler takes a TaskCancelConfirm, not
+            # the worker_id the workerDisconnecting targets take.
+            await self.__routing(
+                task_id,
+                TaskTransition.workerDisconnect,
+                task_cancel_confirm=TaskCancelConfirm(taskId=task_id, cancelConfirmType=TaskCancelConfirmType.canceled),
+            )
+            return
```
magniloquency, line 198: This is a good change, but makes me think that we might need to rethink this particular state machine transition if/when we audit the task state machine.

src/scaler/scheduler/controllers/task_controller.py
magniloquency: i can't find this commit

src/scaler/scheduler/controllers/worker_controller.py
```diff
@@ -167,42 +167,57 @@ def get_worker_ids(self) -> Set[WorkerID]:
...
+
+        workers_set = self._manager_to_workers.get(manager_id)
+        if workers_set is None:
+            return
 
-    async def __disconnect_worker(self, worker_id: WorkerID):
-        """return True if disconnect worker success"""
+        workers_set.discard(worker_id)
+        if not workers_set:
+            del self._manager_to_workers[manager_id]
```
magniloquency: Something feels off about this. Is this function's purpose to forget a worker manager? I see that it gets the manager_id by worker_id, but then i only removes that specific worker from the worker manager's worker set. Perhaps this function should be named differently?

src/scaler/scheduler/controllers/worker_controller.py
magniloquency, line 43: Future improvement: I think we should flatten these into one dict and create some kind of worker control struct: `self._workers: Dict[WorkerID, WorkerControlStruct]`. Usually multiple dicts with the same key is an indication you want something like this, unless you explicitly need a struct of arrays

src/scaler/utility/event_loop.py
```diff
@@ -35,12 +35,18 @@ def register_event_loop(event_loop_type: str):
     logger.info(f"use event loop: {event_loop_type}")
 
 
-def create_async_loop_routine(routine: Callable[[], Awaitable], seconds: int):
+def create_async_loop_routine(routine: Callable[[], Awaitable], seconds: int, swallow_routine_errors: bool = False):
     """create async loop routine,
 
     - if seconds is negative, means disable
     - 0 means looping without any wait, as fast as possible
```
magniloquency: Some more future improvements... Sorry to clutter the PR but I'm just noticing many of these now. `seconds` would be better named as `interval` perhaps. Why do we accept `seconds < 0`?

magniloquency (changes requested):

src/scaler/scheduler/controllers/task_controller.py
e117649: that was from before cleaning up this PR, it is now part of a90ad03 "fix: stop the balancer looping on a task stuck in balanceCanceling"

src/scaler/scheduler/controllers/policies/simple_policy/scaling/capability_scaling.py
e117649: Agreed

e117649: You're right, it was not, fixed in 55b88e9.

src/scaler/scheduler/controllers/task_controller.py
e117649: Yep, we should fix this as part of the audit of the whole state machine and transition table.

src/scaler/scheduler/controllers/worker_controller.py
e117649: You are right, it does not forget the manager: it drops one worker's manager mapping and removes that worker from the manager's set, keeping the manager as long as it has other workers. Renamed to __remove_worker_from_manager (55b88e9).

src/scaler/scheduler/controllers/worker_controller.py
e117649: Yep, agreed

src/scaler/scheduler/scheduler.py
```diff
@@ -256,20 +256,33 @@ async def on_receive_message(self, source: bytes, message: BaseMessage):
     async def get_loops(self):
         await self.__initialize_network()
 
+        # swallow_routine_errors=True: a bug in any single routine must not escape to asyncio.gather and
+        # tear the whole scheduler down, taking every connected client and worker with it.
         loops = [
-            create_async_loop_routine(self._binder.routine, 0),
-            create_async_loop_routine(self._connector_storage.routine, 0),
-            create_async_loop_routine(self._graph_controller.routine, 0),
+            create_async_loop_routine(self._binder.routine, 0, swallow_routine_errors=True),
```
magniloquency, line 262: I think this swallows `ClientShutdownException`?

magniloquency (changes requested):

src/scaler/scheduler/controllers/client_controller.py
```diff
@@ -57,6 +57,10 @@ def has_client_id(self, client_id: ClientID) -> bool:
         return client_id in self._client_last_seen
 
     def get_client_id(self, task_id: TaskID) -> Optional[ClientID]:
+        # get_key raises for a task no tracked client owns, which a TaskLog arriving after its task has
+        # finished hits straight from on_receive_message, killing the scheduler. Return the documented None.
```
rafa-be:
```suggestion
```

src/scaler/scheduler/controllers/graph_controller.py
```diff
@@ -116,14 +119,19 @@ async def on_graph_task(self, client_id: ClientID, graph_task: GraphTask):
         await self._unassigned.put((client_id, graph_task))
 
     async def on_graph_task_cancel(self, task_cancel: TaskCancel):
-        graph_task_id = self._task_id_to_graph_task_id[task_cancel.taskId]
+        graph_task_id = self._task_id_to_graph_task_id.get(task_cancel.taskId)
+        if graph_task_id is None or graph_task_id not in self._graph_task_id_to_graph:
+            return
```
rafa-be, line 125: Maybe we should log these events.

src/scaler/scheduler/controllers/graph_controller.py
rafa-be: Should we log these events?

src/scaler/scheduler/controllers/worker_controller.py
rafa-be: Also, manager's IDs could be improved with a specific identifier type (`WorkerManagerID` instead of `bytes`).

It helps with logging and allows the linter to find some issues.

src/scaler/scheduler/controllers/worker_controller.py
```diff
@@ -167,42 +167,57 @@ def get_worker_ids(self) -> Set[WorkerID]:
...
-        workers_set.discard(worker_id)
-        if not workers_set:
-            del self._manager_to_workers[manager_id]
 
         task_ids = self._policy_controller.remove_worker(worker_id)
         if not task_ids:
             return
 
-        logger.info(f"{len(task_ids)} task(s) failed due to worker {worker_id!r} disconnected")
+        logger.warning(f"{worker_id!r}: rerouting/failing {len(task_ids)} in-flight task(s)")
```
rafa-be: I found the new message a little bit confusing. Would it make sense to merge it with the "worker disconnected" at the top?

```suggestion

        if not task_ids:
            logger.info(f"{worker_id!r} disconnected ({reason})")
            return
        
        logger.warning(f"{worker_id!r} disconnected ({reason}): rerouting/failing {len(task_ids)} task(s)")
```

src/scaler/scheduler/scheduler.py
rafa-be: I think yes.

src/scaler/worker/worker.py
```diff
@@ -345,6 +355,10 @@ async def __teardown(self) -> None:
             # Windows named pipes have no filesystem entry to remove; only unlink Unix-domain-socket paths.
             pathlib.Path(self._address_internal.host).unlink(missing_ok=True)
 
+        if failed_to_destroy:
+            # `_run` turns this into a nonzero exit code, so a failed teardown reaches the supervisor.
```
rafa-be:
```suggestion
```

tests/io/test_ymq_async_binder.py
```diff
@@ -27,27 +27,41 @@ async def asyncSetUp(self) -> None:
         self._binder = YMQAsyncBinder(self._context, identity=b"binder-under-test", callback=self._on_receive)
         await self._binder.bind(AddressConfig.from_string("tcp://127.0.0.1:0"))
 
+    async def asyncTearDown(self) -> None:
+        # Destroy here rather than only in the tests: a binder left alive with a queued send is
+        # destroyed by __del__ at interpreter exit instead, where shutdown() blocks forever. A failing
+        # assertion would otherwise hang the whole run rather than just fail.
+        self._binder.destroy()
```
rafa-be: >  a binder left alive with a queued send is destroyed by __del__ at interpreter exit instead, where shutdown() blocks forever.

Normally, calling `shutdown()` on a YMQ socket should cancel all writes and not block:

https://github.com/finos/opengris-scaler/blob/7327b309253fb863f1079b6f5783ec0bd768307e/src/cpp/scaler/ymq/binder_socket.cpp#L47-L55

I asked Claude and it confirms it should not block, but came up with a GIL dead-lock in the Python wrapper when calling shutdown(): https://github.com/finos/opengris-scaler/issues/945

Calling `.destroy()` explicitly is good though. But the comment might be wrong.

tests/worker/agent/test_suspended_processor.py
```diff
@@ -0,0 +1,111 @@
...
+        parent_pid_path = os.path.join(self.directory, "parent_pid")
+        child_started_path = os.path.join(self.directory, "child_started")
+
+        with Client(self.address) as client:
+            future = client.submit(parent_task, client, parent_pid_path, child_started_path, CHILD_DURATION_SECONDS)
+
+            self.assertEqual(future.result(timeout=RESULT_TIMEOUT_SECONDS), "child done")
+
+            # The resumed processor must still be able to run tasks.
+            self.assertEqual(client.submit(square, 6).result(timeout=RESULT_TIMEOUT_SECONDS), 36)
```
rafa-be: That test might be redudant with tests in [test_nested_task.py](https://github.com/finos/opengris-scaler/blob/7327b309253fb863f1079b6f5783ec0bd768307e/tests/task/test_nested_task.py).

rafa-be (dismissed):

src/scaler/scheduler/scheduler.py
e117649: Yep you're right, `create_async_loop_routine` now re-raises

tests/io/test_ymq_async_binder.py
e117649: You're right, updated

magniloquency (approved):

PR #907, opened 2026-07-22
fix: improve visibility into worker and worker-manager crashes

magniloquency:
## Summary
- Worker's exception handling is lifted from `__get_loops` into `_run()`, spanning init through the main loop: every exception path now logs (no more silent `pass`), and the process exits nonzero only for genuinely unexpected errors instead of always exiting 0. `WorkerProcess` (backing the Symphony, AWS-HPC, and OCI-HPC worker manager adapters) gets the identical fix, since it had the same silently-swallowed-exception, always-exit-0 pattern.
- Exit codes distinguish a deliberate stop from a worker giving up on its own: `ClientShutdownException`/`CancelledError`/expected YMQ teardown races stay exit 0, while `TimeoutError` (self-detected orphan via `death_timeout_seconds`) and `ObjectStorageException` (can't reach the object storage server) now exit 1, since neither represents anyone asking the worker to stop.
- Enable Python's `faulthandler` across every scaler process entry point (worker, processor, scheduler, object storage server, worker managers, gui) via a new `bootstrap_process()` helper, so a SIGSEGV/SIGABRT/SIGBUS/SIGFPE/SIGILL (e.g. an uncaught C++ exception in ymq hitting `std::terminate`/`abort`) dumps a Python-level stack trace instead of leaving zero trace. Correctly recognizes the `-`/`/dev/stdout` logging-path sentinels as "write to stdout" (matching `setup_logger`) instead of trying to open a literal file named `-`. The faulthandler target is derived from the logger's actual configured state (`get_logger_info`) rather than the raw `log_paths` argument, since `log_paths` is meaningless once a `logging_config_file` is supplied.
- Fixed-mode native worker manager (`run_fixed()`) now logs every worker exit (with the exit code translated to a signal name where relevant) instead of blindly joining in sequence with no visibility, and distinguishes a signal-requested stop, a worker that shut down cleanly on its own, and an unexpected death (now that exit code 0 only ever means a deliberate stop). Since this observes the OS-level process exit from the manager side rather than relying on the worker's own cooperation, it also correctly detects and labels a SIGKILLed worker (exitcode -9), not just ordinary crashes. This classification tracks exactly which workers the shutdown signal handler terminated, rather than a single shared flag, so a worker that crashes independently right around a shutdown signal is never mislabeled as a deliberate stop.
- `run_fixed()` no longer bootstraps its own logging/faulthandler - that only mattered for `SchedulerClusterCombo`, which spawns `NativeWorkerManager.run()` directly with no prior setup, since the CLI entry points already bootstrap before reaching it. `SchedulerClusterCombo` now bootstraps at that actual spawn site instead, so the CLI-invoked path no longer bootstraps twice.

Two known gaps remain, not addressed here: `faulthandler` cannot catch SIGKILL under any circumstances (no userspace handler can, by kernel design), and dynamic-mode's `NativeWorkerProvisioner` still has no liveness monitoring at all, so an unexpectedly killed worker there (SIGKILL or otherwise) goes undetected until the scheduler's heartbeat timeout.

### DIFF
--- a/src/scaler/cluster/combo.py
+++ b/src/scaler/cluster/combo.py
@@ -34,11 +34,21 @@ from scaler.config.section.scheduler import PolicyConfig
 from scaler.config.types.address import AddressConfig, SocketType
 from scaler.config.types.worker import WorkerCapabilities
 from scaler.utility.network_util import get_available_tcp_port
+from scaler.utility.process_bootstrap import bootstrap_process
 from scaler.worker_manager_adapter.baremetal.native import NativeWorkerManager
 
 logger = logging.getLogger(__name__)
 
 SCHEDULER_GRACEFUL_SHUTDOWN_TIMEOUT_SECONDS = 10
+
+
+def _run_native_worker_manager(worker_manager: NativeWorkerManager) -> None:
+    # This process is spawned directly, so it must bootstrap its own logging/faulthandler.
+    logging_config = worker_manager.config.logging_config
+    bootstrap_process(
+        logging_config.paths, logging_config.config_file, logging_config.level, process_name="worker_manager_native"
+    )
+    worker_manager.run()
 
 
 class SchedulerClusterCombo:
@@ -130,7 +140,9 @@ class SchedulerClusterCombo:
             )
         )
 
-        self._worker_manager_process = multiprocessing.get_context("spawn").Process(target=self._worker_manager.run)
+        self._worker_manager_process = multiprocessing.get_context("spawn").Process(
+            target=_run_native_worker_manager, args=(self._worker_manager,)
+        )
 
         # Synthesized signal-disposition for the scheduler. multiprocessing.Process.terminate() is
         # TerminateProcess on Windows, which never runs Python signal handlers, so the scheduler would
--- a/src/scaler/cluster/object_storage_server.py
+++ b/src/scaler/cluster/object_storage_server.py
@@ -10,7 +10,8 @@ from scaler.io.ymq import YMQException
 from scaler.object_storage.object_storage_server import ObjectStorageServer
 from scaler.utility.exceptions import ObjectStorageException
 from scaler.utility.identifiers import ClientID, ObjectID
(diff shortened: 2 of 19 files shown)


e117649 (commented):
Looks good to me, claude raised a few nits, I think (1) is worth changing now, your call on the other 3:

> 1. Exit-code asymmetry. ObjectStorageException (storage unreachable) → exit 1, but ConnectorSocketClosedByRemoteEnd → exit 0 — and the old comment notes that also covers "scheduler never reachable, connector exhausted retries." So a worker that can never reach the scheduler at startup exits 0 (looks clean), while an unreachable object store exits 1. Given the PR's goal ("nonzero on unreachable dependency"), that's inconsistent. The counter-argument is real too — a dropped scheduler is normal on scale-down, so you don't want nonzero noise. Worth confirming intent; distinguishing "never connected" from "connected then dropped" would resolve it cleanly.
> 2. _run is ~40 near-identical lines in worker.py and worker_process.py. Two separate classes, so it's defensible, but a shared helper would keep the exit taxonomy from drifting.
> 3. Teardown isn't guarded. finally: await self.__teardown() — if teardown raises, the intended exit_code and the "quit" log are lost and the process dies on the teardown error instead. Wrapping teardown so the chosen exit code always wins would be tidier.
> 4. Nit: _faulthandler_file is overwritten without closing if bootstrap_process were ever called twice (once-per-process today, so harmless).

magniloquency:
@e117649:

1. fixed. we set a flag on the first heartbeat received and if we exit before that it's considered an error (never connected to the scheduler), otherwise it's a graceful exit (this is how it's implemented, apparently, i think we should look into making graceful exits more clear later)
2. leaving as is for now, worker and worker_process code duplication is a known issue, we should explore a larger refactor later
3. fixed. added a separate try except around teardown
4. fixed it anyway, it now closes an existing file handle before opening a new one if it exists

e117649 (dismissed):

e117649 (approved):

darenliang (approved):
LGTM

PR #908, opened 2026-07-22
DO NOT MERGE: diagnose windows CI hang in test_scheduler_disconnect_raises_disconnected_error

magniloquency:
Throwaway diagnostic branch to root-cause a windows-latest CI failure seen on #907 (`test_scheduler_disconnect_raises_disconnected_error` hangs the full 30s on Windows instead of getting `DisconnectedError`/`TimeoutError`).

Adds temporary instrumentation to `ClientAgent`/`ClientHeartbeatManager` to check whether the client agent's asyncio event loop keeps ticking on Windows after the scheduler is killed, or stalls outright. Will be closed once the Windows job finishes.

### DIFF
--- a/src/scaler/client/agent/client_agent.py
+++ b/src/scaler/client/agent/client_agent.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 import sys
 import threading
+import time
 from concurrent.futures import Future
 from typing import Callable, Optional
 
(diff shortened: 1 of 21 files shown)


magniloquency:
Root-caused via this diagnostic (see #909) -- closing now that we have the answer.

Issue #912, opened 2026-07-22
Windows CI Hangs


magniloquency:
The Windows CI is hanging following the merge of #906. I think that this is likely not a bug introduced in that PR, but rather a latent bug exposed by it. I hypothesize that the root cause of this bug is related to socket shutdown handling on Windows, but more investigation is needed to be sure.

To fix this in the interim `test_scheduler_crash` and `test_scheduler_disconnect_raises_disconnected_error` have been disabled in PR #913

## The following analysis was generated by claude and is unverified

---

1. Socket teardown has no bound on Windows. MessageConnection only detects a dead peer via its read path (EOF/ECONNRESET). When a peer is forcibly killed (TerminateProcess, what .kill() does), Windows -- unlike POSIX -- does not synchronously tear down that process's sockets and notify the peer, so nothing ever tells libuv the connection is dead. Client teardown (MessageConnection::shutdownClient()) waits on a uv_shutdown() callback that will never fire in this case, and the Python binding that calls it (PyConnectorSocket_shutdown / PyBinderSocket_shutdown) blocks the calling thread on that wait with no timeout, ever. Since tests run sequentially in one process, one wedged teardown hangs the whole suite behind it.

Empirically confirmed: enabling SO_KEEPALIVE on the YMQ connection (to bound dead-peer detection via the OS's own probing) eliminated the hang for test_scheduler_disconnect_raises_disconnected_error in an isolated CI run -- strong evidence for this being the actual mechanism, not a guess.

2. Client heartbeat death-timeout is coupled to the heartbeat send. Separately, ClientHeartbeatManager.routine() checks the death-timeout and sends the heartbeat in the same loop iteration. If the send itself never resolves (which it won't, per bug 1, once the scheduler is dead), the loop never gets back around to re-check the timeout, so timeout detection can hang too. This is fixed on a branch (splitting the check into its own independent routine) but that fix alone isn't sufficient -- it just moves the hang to bug 1's teardown path instead of preventing it, since detection now succeeds but the subsequent disconnect()/teardown still blocks unboundedly.

PR #913, opened 2026-07-22
test: skip Windows scheduler-crash tests that hang CI

magniloquency:
## Summary
- `test_scheduler_crash` and `test_scheduler_disconnect_raises_disconnected_error` both kill the scheduler mid-test and immediately tear the client down in the same test, which wedges forever on Windows in an unbounded socket-shutdown wait, hanging the whole `Run Unittests` step.
- Skips both on `win32` as an interim mitigation with a link to #912, which tracks the underlying teardown bug and the real fix.

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -3,6 +3,7 @@ import logging
 import multiprocessing
 import os
 import random
+import sys
 import tempfile
 import time
 import unittest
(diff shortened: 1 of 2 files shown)


e117649 (approved):

1597463007 (approved):
LGTM

PR #919, opened 2026-07-27
fix: log why a processor exits instead of surfacing it as a zombie

magniloquency:
When a native-worker processor dies on its own (e.g. object storage drops while fetching a task's inputs, or the process is OOM-killed), the only trace was a later, cryptic `process died process_status='zombie'` from the agent's poll -- the real cause was invisible because the processor exited silently and the agent discarded the exit code.

- Route every exit path in `Processor.__run_forever` through a `__log_exit` helper: ERROR (with traceback) for real exceptions, WARNING naming the orphaned `task_id` when a task was in flight, DEBUG for routine shutdowns. `SystemExit` is now logged instead of exiting silently.
- Surface the real exit code: add `scaler.utility.exitcode.describe_exitcode` (naming the signal, e.g. `-9 (SIGKILL)`), expose `ProcessorHolder.exitcode()`, and use it in `on_failing_processor` for the log and `ProcessorDiedError`.

The `describe_exitcode` implementation and its test (`tests/utility/test_exitcode.py`) are taken verbatim from #900, and the shared exit-code pieces mirror it, so the two branches merge without conflict.

### DIFF
--- a/src/scaler/utility/exitcode.py
+++ b/src/scaler/utility/exitcode.py
@@ -1,0 +1,16 @@
+import signal
+from typing import Optional
+
+
+def describe_exitcode(exitcode: Optional[int]) -> str:
+    """Render a process exit code, naming the signal for signal-terminated processes.
+
+    multiprocessing reports a process killed by signal N as exit code -N, so an OOM kill (SIGKILL) shows
+    as -9. Turn that into e.g. "-9 (SIGKILL)" so the cause is legible in logs.
+    """
+    if exitcode is not None and exitcode < 0:
+        try:
+            return f"{exitcode} ({signal.Signals(-exitcode).name})"
+        except ValueError:
+            pass
+    return str(exitcode)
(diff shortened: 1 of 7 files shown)


e117649 (commented):
1. Swallowing SystemExit zeroes the exit code. The new handler catches it and doesn't re-raise, so _bootstrap no longer sees it — I confirmed a process that exited 3 now exits 0. It's reachable (a task calling sys.exit(N) isn't caught by __process_task's except Exception), and it destroys the very exit code this PR surfaces: on_failing_processor would report exitcode=0. One-line fix: raise after logging.

2. The task context is missing in the PR's own headline scenario. __send_result sets self._current_task = None before writing to storage, so a storage failure there logs with no task id and never emits the "no task result will be sent" warning. Captured from the branch — the message has no task id while the traceback shows __send_result → set_object failing.

3. Deliberate kills now log ERROR + traceback. kill() → SIGTERM → __interrupt destroys the connectors on purpose; a processor mid-storage-call then raises ObjectStorageException("connector is closed."), which __log_exit escalates to ERROR because an exception object was passed. In my run that ERROR is immediately followed by stop Processor[...], reason: quit — the agent reporting an error for something it asked for. Routine cancels and scale-downs will now emit tracebacks. Suggested an _interrupted flag set in __interrupt for __log_exit to consult.

e117649 (dismissed):

rafa-be (approved):

Issue #924, opened 2026-07-31
Scheduler fail to shutdown properly if a worker is dying


rafa-be:
`tests.scheduler.test_dead_worker` succeed but shows a very suspicious exception while shutting down the cluster combo:

```bash
$ python -m unittest tests.scheduler.test_dead_worker
[..]
2026-07-31 19:24:11+0200 INFO scaler[2960]: ScalerClient: disconnect from tcp://127.0.0.1:38227
2026-07-31 19:24:11+0200 INFO scaler[2960]: canceling 0 task(s)
2026-07-31 19:24:11+0200 INFO scaler[2960]: ScalerClient: disconnect from tcp://127.0.0.1:38227
2026-07-31 19:24:11+0200 INFO scaler[2960]: canceling 5 task(s)
2026-07-31 19:24:11+0200 INFO scaler[2972]: ClientID(Client|532f8581688345ee8f654f71620ab018) disconnected
2026-07-31 19:24:11+0200 INFO scaler[2960]: ClientAgent: client quitting
2026-07-31 19:24:11+0200 INFO scaler[2960]: YMQAsyncConnector: exited
2026-07-31 19:24:11+0200 INFO scaler[2960]: ClientHeartbeatManager: exited
2026-07-31 19:24:12+0200 INFO scaler[2972]: balancing task: {WorkerID(Worker|FIX|98866a30d79543fa92c1258cd4f4edca|6c732242be8a46418e84d9e67199f5ad): 2}
2026-07-31 19:24:12+0200 ERROR scaler[2972]: b'\x02\xb8N\xa2\xa4\xa3J\xd8\x99*\xcd`\xc2\xbb\xedO': cannot apply 8 to current state 3
2026-07-31 19:24:12+0200 INFO scaler[2972]: b'\x02\xb8N\xa2\xa4\xa3J\xd8\x99*\xcd`\xc2\xbb\xedO': unknown transition: 8
2026-07-31 19:24:12+0200 ERROR scaler[2972]: b'\x1d5\x02b\xe9\x86Ky\x82R\xbf\x0e\rU\x9d\x8e': cannot apply 8 to current state 3
2026-07-31 19:24:12+0200 INFO scaler[2972]: b'\x1d5\x02b\xe9\x86Ky\x82R\xbf\x0e\rU\x9d\x8e': unknown transition: 8
[...]

PR #925, opened 2026-07-31
fix: Detached send() for Scaler connectors

rafa-be:
Allows `send()` calls to be executed asynchronously, preventing the scheduler to be unresponsive in some events (e.g. during a worker being unexpectedly killed).

The unit-test I added seems to surface another issue: https://github.com/finos/opengris-scaler/issues/924 (will be solved in another PR).

### DIFF
--- a/src/scaler/worker/agent/processor_manager.py
+++ b/src/scaler/worker/agent/processor_manager.py
@@ -129,7 +129,7 @@ class VanillaProcessorManager(ProcessorManager):
 
         self._profiling_manager.on_task_start(holder.pid(), task.taskId)
 
-        await self._binder_internal.send(holder.processor_id(), task)
+        await self._binder_internal.send(holder.processor_id(), task, detached=True)
 
         return True
 
@@ -194,7 +194,8 @@ class VanillaProcessorManager(ProcessorManager):
                         objectTypes=(ObjectMetadata.ObjectContentType.object,),
                         objectNames=(b"",),
                     ),
-                )
+                ),
+                detached=True,
             )
 
             await self._task_manager.on_task_result(
@@ -290,13 +291,13 @@ class VanillaProcessorManager(ProcessorManager):
             if processor_id not in self._holders_by_processor_id:
                 continue  # processor got killed while we were iterating over the list
 
-            await self._binder_internal.send(processor_id, instruction)
+            await self._binder_internal.send(processor_id, instruction, detached=True)
 
     async def on_internal_object_instruction(self, processor_id: ProcessorID, instruction: ObjectInstruction):
         if not self.__processor_ready_to_process_object(processor_id):
             return
 
-        await self._connector_external.send(instruction)
+        await self._connector_external.send(instruction, detached=True)
 
     def destroy(self, reason: str):
         if self._connector_storage is not None:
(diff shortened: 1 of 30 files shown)


src/scaler/worker/agent/processor_manager.py
```diff
@@ -290,13 +291,13 @@ async def on_external_object_instruction(self, instruction: ObjectInstruction):
             if processor_id not in self._holders_by_processor_id:
                 continue  # processor got killed while we were iterating over the list
 
-            await self._binder_internal.send(processor_id, instruction)
+            await self._binder_internal.send(processor_id, instruction, detached=True)
```
magniloquency, line 294: Because this socket is internal, it should always be "fast", right? How are we deciding which sends should be attached and which should be detached? Making scheduler sends detached make sense because of the number of workers it manages and how bad it would be if the loop stalled, but does the worker need it too? I'm not saying that the worker shouldn't be detached, but I want to understand why we should have it one way or the other

magniloquency (commented):
one comment, otherwise lgtm

src/scaler/worker/agent/processor_manager.py
rafa-be: I set everything to `detached = True` as that's the same behavior as ZMQ, for which we designed these routines for.

However you're right, it might be better to have `detached = False` for these internal sockets. Performances shouldn't be much different indeed.

The main difference will be that `send()` will raise an exception on a write error, and crash the worker, while a detached `send()` will only log the error. I've to think more if it can break something but maybe it's better.

e117649: My vote is to leave it detached, attaching would raise on a send to a processor that has died but not yet been reaped, and that lands in `_run`'s `ConnectorSocketClosedByRemoteEnd` branch, so the worker exits 0 looking like a normal scheduler disconnect.

e117649 (approved):

src/scaler/worker/agent/processor_manager.py
magniloquency: That error handling seems a bit suspicious to me, if a processor has died that's something we need to handle. Exiting 0 in that case is I think maybe not correct

magniloquency (commented):
I have some more thoughts on this, but nothing really that blocks this from going in. What I think we're finding is that the scheduler has not been very resilient against workers disappearing randomly (which is in effect what worker managers do right now).

These detached sends definitely solves the problem of a blocked send causing the scheduler to hang, but making them detached like this also means that we can't make use of any error that arises from it. If the send returned an error like "remote end disconnected" (or whatever it's actually called), that could be a good and early signal to the scheduler that a worker has disappeared for example.

The only other thing I could imagine is going back to attached sends with some kind of timeout in place, but I'm not sure that's right either because YMQ virtualizes sockets and there's always the possibility that there's transitive network failure which is resolved after a reconnect.

Ultimately our various timeouts should be able to handle these cases that fall through the cracks, and such cases might become much rarer if/when we make worker shutdown more graceful, so maybe this is just fine?

To be honest part of me wonders if this is an indication that the design of the scheduler's concurrency could be improved? I think it wouldn't be unusual for us to have e.g. one asyncio task per worker, rather than putting everything together. But that's a larger architectural discussion that I don't think we should have right now

magniloquency (approved):

PR #927, opened 2026-08-06
fix: Lost task result on transport drop

e117649:
A processor that has finished a task can lose the result if the connection under the hand-off drops — and that work is already paid for, up to hours of it.

Workers in pods behind AWS middleboxes, which reap an idle flow after 350 s, and a long running task that left the storage flow idle for the whole of it. The write was cancelled the moment it was attempted, the exception unwound the processor's main loop, and the agent reported the task as failed with ProcessorDiedError — which for a graph subtask aborts the whole graph. The connectors reconnect on their own, so the very next attempt would have landed.

- Storage server: drop a partially delivered request. A set/duplicate arrives as a header then a payload, keyed on the sender's identity while the server waits. A connection dropping in between leaves a request that can never complete, and a client keeps its identity across a reconnect, so the next request's header was taken for the missing payload: a debug build asserted, a release build stored the header as the object's content and left the client waiting forever. The announced length catches it. First, because it's what makes a retry safe.
- Sync storage connector: report a transport failure as a storage failure. The read side already translated YMQ errors into ObjectStorageException; the write side let them through raw, so a cancelled storage write read as a failure of whatever socket the caller happened to be handling errors for — logged as "agent connector stop requested", pointing the diagnosis at a healthy connection.
- Processor: retry the hand-off instead of dropping it. Store, announce, send — each step retried with a growing delay. Every one is safe to repeat: the agent hasn't been told yet, and each storage attempt uses a fresh object ID, so a retry can never announce an ID the server holds a half-written object under. A teardown the agent asked for is never retried.

Verified against a real storage server behind a TCP proxy that resets mid-write: at 320 MB it reproduces the production error exactly and the retry recovers the result byte-identically. The new C++ test is red without the server fix.

**Not fixed here:** with a payload small enough that the kernel takes the whole write, an aborted send is still reported as successful, so the connector blocks forever on a response that died with the connection.

**Note: The length check is not airtight.** It catches a reconnect whose next header differs in size from the payload the dropped request announced — every size but one. A serialized ObjectRequestHeader is 80 bytes, so a request that announced an 80-byte payload and then lost its connection will still have the next header taken for that payload. Closing that needs something the protocol lacks: a request sequence, or a signal that the peer behind an identity reconnected. Before this change, every size was mis-consumed.


src/cpp/scaler/object_storage/object_storage_server.cpp
```diff
@@ -175,6 +175,25 @@ void ObjectStorageServer::processRequests(std::function<bool()> running)
...
+                    "ObjectStorageServer: dropping a partially delivered request, expected a payload of ",
+                    it->second.first.payloadLength,
+                    " bytes, got a message of ",
+                    headerOrPayload->size(),
+                    " bytes");
+
+                identityToFullRequest.erase(it);
+                it = identityToFullRequest.end();
+            }
+
```
rafa-be, line 196: That looks like a legit issue, but this will only fix the issue if the payload isn't the size of the header.
 
I'd say that's fine for now, but I created [this issue](https://github.com/finos/opengris-scaler/issues/931) to keep track of the problem.

src/scaler/worker/agent/processor/processor.py
```diff
@@ -214,8 +222,10 @@ def __run_forever(self):
 
                 self.__on_connector_receive(message)
 
-        except ymq.SocketStopRequestedError:
-            self.__log_exit("agent connector stop requested")
+        except ymq.SocketStopRequestedError as e:
+            # Only the agent connector can raise this here, now that the storage connector reports its own
+            # failures as ObjectStorageException. Name the error rather than rely on that attribution.
```
rafa-be: That LLM comment refers to another part of the PR without context, I think it should go away.

rafa-be (dismissed):

src/cpp/scaler/object_storage/object_storage_server.cpp
magniloquency: I agree with Raphael and I think this indicates a deeper issue with the OSS. It relies on the ordering of multiple payloads for its message structure. This on its own is not a problem, but it would be more robust if the message parts had metadata to indicate which object they're for, which part of the message it is, etc. At the moment the 'payload' part of OSS messages contain only the data of the object it is transferring, as far as I know.

For now this works though and is unlikely to cause an issue

src/scaler/worker/agent/processor/processor.py
```diff
@@ -364,16 +371,64 @@ def __send_result(self, source: ClientID, task_id: TaskID, task_result_type: Tas
...
+                    raise
+
+                logger.warning(
+                    f"Processor[{self.pid}]: failed to {description} of task_id={task_id.hex()} on attempt "
+                    f"{attempt}/{RESULT_HAND_OFF_MAX_ATTEMPTS} ({e!r}), retrying in {delay} seconds"
+                )
+
+            time.sleep(delay)
+            delay *= 2
+            attempt += 1
```
magniloquency, line 434: I'm a bit curious about this. YMQ provides virtualized sockets, they're meant to be sort of isolated from what's happening at the OS layer. A YMQ socket is meant to live longer than any individual connection. When you send data over a YMQ socket, my understanding is that YMQ's promise is that it will try its hardest to send the data, even if that involves multiple reconnects. The need for this indicates to me that somehow that design is not being honoured here. Do we know exactly what error is raised here?

magniloquency (changes requested):

src/scaler/worker/agent/processor/processor.py
e117649: The error is `SocketStopRequestedError`, from:
https://github.com/finos/opengris-scaler/blob/66d101380aa39dba87049f3e6ab0924bee870a87/src/cpp/scaler/ymq/internal/message_connection.cpp#L286-L291

magniloquency: Hmmmm. To be honest, this kind of reads like a bug to me. I think we should make it so YMQ retries on this, however I'm content to get this fix through for now and defer this cpp-level fix until later. @rafa-be, do you have any thoughts on this?

magniloquency (approved):
Approving this because I think it fixes the bug for now, but I think we should revisit YMQ's behavior in this situation

src/scaler/worker/agent/processor/processor.py
rafa-be: `SocketStopRequestedError` only occurs when the socket is locally closed (i.e. not because the remote dropped).

At first, we only expected it to be triggered only when `socket.close()` is called, but it can also happen if the OS kills the socket itself (e.g. on local TCP timeout).

As there is a guard to check that we haven't requested the shutdown (`self._interrupted` should `False`), I think this function is somewhat valid.

Alternatively YMQ could return another exception when `close()` if initiated by the OS, but it requires more changes.

rafa-be (approved):

src/scaler/worker/agent/processor/processor.py
rafa-be: > I think we should make it so YMQ retries on this

YMQ, like ZMQ, does not guarantee delivery of messages. However, YMQ could be more proactive in these non-initiated `close()` events and retry itself.

magniloquency: > YMQ, like ZMQ, does not guarantee delivery of messages

I feel like I recall some specific conversation around this in the past, but I'm not sure if any of it was recorded, and so I haven nothing to point to.

PR #928, opened 2026-08-06
fix(ymq): enable TCP keep-alive on connections

e117649:
Scaler connections routinely go idle for the length of a task: a processor fetches its arguments, computes for minutes or hours, then writes its result. With no probes on the wire, neither end learns that a peer died or that a middlebox (NAT, firewall, load balancer) dropped the flow until the next write -- which is the write carrying the result, at the point where failing costs the most.

Enable keep-alive with a 30 second idle delay, well under the usual middlebox timeouts, on every TCP-backed connection (plain, TLS and WebSocket). A socket option the platform refuses is logged rather than fatal: the connection still works, it just cannot notice an idle peer going away.


magniloquency (dismissed):

src/cpp/scaler/ymq/internal/client.cpp
```diff
@@ -107,6 +107,23 @@ std::expected<void, scaler::wrapper::uv::Error> Client::setNoDelay(bool enable)
...
+        return tcp->keepalive(enable, delaySeconds);
+    }
+    if (auto* tls = std::get_if<scaler::wrapper::openssl::SecureSocket>(&_socket)) {
+        return tls->keepalive(enable, delaySeconds);
+    }
+    if (auto* ws = std::get_if<WebSocketStream>(&_socket)) {
+        return std::visit(
+            [enable, delaySeconds](auto& transport) { return transport.keepalive(enable, delaySeconds); },
+            ws->transport());
+    }
```
rafa-be, line 120: Why not adding a `keepalive()` method to `WebSocketStream` like you do for `SecureSocket`?

rafa-be (dismissed):

magniloquency (approved):

rafa-be (approved):

PR #929, opened 2026-08-06
test: bound the dead worker test's setup wait and always tear its cluster down

e117649:
The wait for the worker processes to appear had no deadline, so a worker manager that never spawns them hangs the run instead of failing it, and the cluster was only shut down on the path where every assertion held -- a failing one leaves a scheduler, a worker manager and its workers behind for the rest of the suite.

Neither is what makes this test flaky today. It kills a worker, and once the scheduler times that worker out it reroutes its tasks; if one of them is being cancelled at that moment the scheduler dies of an unrelated bug, and the test then wedges in teardown rather than failing. This only makes sure that when it does fail, it fails.

### DIFF
--- a/tests/scheduler/test_dead_worker.py
+++ b/tests/scheduler/test_dead_worker.py
@@ -39,6 +39,7 @@ class TestDeadWorker(unittest.TestCase):
         combo = SchedulerClusterCombo(
             address=address, n_workers=N_WORKERS, worker_timeout_seconds=WORKER_TIMEOUT_SECONDS
         )
+        self.addCleanup(combo.shutdown)
 
         with Client(address=address) as client:
             # Ensures the workers are connected and registered by the scheduler.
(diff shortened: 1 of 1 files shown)


tests/scheduler/test_dead_worker.py
```diff
@@ -39,6 +39,7 @@ def test_scheduler_responsive_after_worker_death(self):
         combo = SchedulerClusterCombo(
             address=address, n_workers=N_WORKERS, worker_timeout_seconds=WORKER_TIMEOUT_SECONDS
         )
+        self.addCleanup(combo.shutdown)
```
magniloquency, line 42: i think there are a few other tests that call combo.shutdown() at the end. can we apply this fix to those too?

rafa-be: Most of them do it in the `tearDown()` method.

But for those that do it in the test method, yes.

rafa-be (approved):

tests/scheduler/test_dead_worker.py
e117649: I think we can do this in a separate PR

magniloquency: sounds good -- separate pr

magniloquency (approved):

PR #938, opened 2026-08-13
fix: let a nested task past the per-worker task queue limit

e117649:
A task that creates a Client and blocks on the tasks it submits holds a worker queue slot until those nested tasks complete. The allocate policies only hand a task to a worker that has a free slot, so once such parents have filled every queue, their nested tasks are refused and wait in the scheduler's unassigned queue for a slot only a blocked parent can release. Nothing in the cluster can make progress: two workers with a queue size of two and four such parents sat idle until the client gave up 60 seconds later, and recovered only because cancelling its tasks freed the slots.

Both allocate policies now exempt a nested task from the queue size. The queue size is admission control for new work, and a nested task is not new work -- the cluster is already committed to its parent. Nor does it add concurrency: the worker suspends the lower-priority parent to run it, exactly as it already does when a queue has room. Client stamps a nested submission with its parent task's priority plus one, so a non-zero priority identifies one.

statistics() now clamps the reported free slots at zero, so an over-committed worker does not report a negative count into the monitor's unsigned field.


rafa-be:
Can this can happen if the nesting hierarchy deepness does not exceed the length of the worker queue?

EDIT:
For example, can a nested call tree with a depth of 4 hang if the queue size is greater than 4?

e117649:
> Can this can happen if the nesting hierarchy deepness does not exceed the length of the worker queue?
> 
> EDIT: For example, can a nested call tree with a depth of 4 hang if the queue size is greater than 4?

Not depth against one worker's queue, it hangs once every slot in the pool is occupied and each worker's running task is blocked on a nested submission, which can be as few as one nesting task per worker. Combined with 5 tasks per worker this can hit when 1/5 of the tasks are nested.

e117649:
Closes https://github.com/finos/opengris-scaler/issues/950

PR #939, opened 2026-08-13
fix: scale to the work, not to a task-to-worker ratio

e117649:
A worker runs one task at a time, so the number of workers is the cluster's parallelism. Both simple scaling policies instead target a fixed tasks-per-worker ratio, which caps a run well below the capacity the fleet was configured to allow and shrinks the fleet while work is still outstanding.

Measured with 15 six-second tasks against a manager permitted 10 workers:

  capability  asks for ceil(tasks / 5), so 3 workers; only 3 ever ran anything, and it scaled down
              to 2 and then to 1 while 9 and then 5 tasks were still queued. 57s instead of 12s.
  vanilla     grows by one worker only while tasks outnumber workers ten to one, then parks: the
              ratio lands inside its [1, 10] band and it re-targets its current count forever. It
              stopped at 2 workers. 50s.

Stopping a worker that still has queued tasks also reroutes them and re-executes anything already in flight, so the shrink is not merely wasteful.

Both now ask for one worker per outstanding task, still bounded by the manager's maxTaskConcurrency -- the knob that exists to cap how much capacity a manager may create. The same run asks for 10 and finishes in 22s (worker startup dominates the rest). Shrinking falls out of the same rule, so a fleet still drains to zero when the work is done, but never below the number of tasks still outstanding.

WaterfallScalingPolicy has the same defect and is left alone: it also distributes a cluster-wide target across managers in priority order, so it needs its own change.

### DIFF
--- a/src/scaler/scheduler/controllers/policies/simple_policy/scaling/capability_scaling.py
+++ b/src/scaler/scheduler/controllers/policies/simple_policy/scaling/capability_scaling.py
@@ -1,5 +1,4 @@
 from collections import defaultdict
-from math import ceil
 from typing import Dict, FrozenSet, List, Tuple
 
 from scaler.protocol.capnp import ScalingManagerStatus, WorkerManagerCommand, WorkerManagerHeartbeat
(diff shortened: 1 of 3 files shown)


rafa-be:
> A worker runs one task at a time, so the number of workers is the cluster's parallelism. Both simple scaling policies instead target a fixed tasks-per-worker ratio, which caps a run well below the capacity the fleet was configured to allow and shrinks the fleet while work is still outstanding.
>
> Measured with 15 six-second tasks against a manager permitted 10 workers:
>
> capability asks for ceil(tasks / 5), so 3 workers; only 3 ever ran anything, and it scaled down
to 2 and then to 1 while 9 and then 5 tasks were still queued. 57s instead of 12s.

Indeed, the current policy is highly sub-optimal.

But I'm not sure scaling to the total number of queued task is really good either. Sounds a bit overkill if the tasks are small.

What about computing a kind of load-average metric (see UNIX's `uptime`), and use that metric for scaling ?

For example, if we have 5 workers, and measure a load-average (= avg. queued tasks per worker over a period) of 3 tasks per worker, we could scale to `5 * 3 = 15` workers. Load-average could be computed on a somewhat moderate duration, e.g. 1 or 3 minutes.

PR #943, opened 2026-08-14
chore: drop the jupyterlite-core override and pin the stack by range

e117649:
jupyterlite-sphinx 0.23.0 raised its cap to jupyterlite-core<0.9, so it no longer conflicts
with jupyterlite-pyodide-kernel and the override is dead. The pins become ranges that are open
to patches and closed to minors: a kernel minor moves the bundled Pyodide series, which is the
wasm wheel's ABI, and a jupyterlite-sphinx minor changes the JupyterLite site layout the docs'
?path= links depend on.


e117649:
Supersedes/closes https://github.com/finos/opengris-scaler/pull/940

rafa-be (approved):

PR #944, opened 2026-08-14
fix: Rewrite the task state machine

magniloquency:
The task state machine encodes a Mealy machine in the shape of a Moore machine. `__routing` selects a handler by the **destination** state but takes its payload from the **caller**, so every caller has to derive the routing rule of the machine in order to know which keywords to pass. Three of the five non-terminal handlers recover the incoming transition with `previous_state()` before they act, and `__state_canceled` cannot, because two paths reach it from `canceling`.

This gives each transition its own event class and its own coroutine. The action receives the source state and the event, does the work, and returns the state that the task lands in. The router matches on the event class, so mypy checks every call against the receiving action's signature, and each action answers every member of `TaskState` with `assert_never`, so a new state becomes a type error in every action instead of a silent rejection.

Moving the actions onto the transitions also moves the commit: the action now runs first and the state is written afterwards, so the state never claims work that has not happened. That reversal removes an exclusion the old order was providing by accident, so this also adds a per-task lock to replace it, and a teardown for the actions that raise.

## Shape of the change

Eight event classes and eight actions replace the twelve members of `TaskTransition`. The three result transitions become one event whose payload carries the difference.

```python
async def __on_worker_disconnect(self, source: TaskState, event: WorkerDisconnected) -> Optional[DisconnectTargetStates]:
    match source:
        case TaskState.running | TaskState.balanceCanceling:
            return await self.__acquire_and_dispatch(event.task_id)
        case TaskState.canceling:
            # remove_worker already released the capacity, so there is no on_task_done here
            ...
            return TaskState.canceled
        case TaskState.inactive | TaskState.success | ...:
            return None
        case _:
            assert_never(source)
```

An arm that returns a state is a legal source; the arm that returns `None` is the ignore list and says why. `None` means "not permitted from this state" -- the router logs it and does nothing, which is what racing duplicates need. The `Literal` return alias is not a second source of truth: mypy checks every `return` against it, so behavior cannot widen without the annotation widening too.

## The order of the commit and the action, and what that costs

Committing first means a failed action leaves the state advanced with none of the work done. `hasCapacity` commits `running` and *then* sends the task to the worker, so a failed send leaves a machine that claims a worker holds a task the worker was never told about; the `canceling` + `workerDisconnect` crash below is the sharpest version, where the task lands in `canceled` without the client ever being told. So the action runs first here.

**The reversal alone is not safe.** Committing first was doing a second job that nothing had written down. On `main`, nothing can interleave between the router reading the source state and writing the new one: `on_transition` is a plain `def` and it is the first statement of `__routing`, and every action awaits only after that, because every action sends something. A second event arriving for the same task while the first is still sending therefore reads the state the first event *already wrote*, and is rejected by it. The commit is what makes the machine a mutex between two events racing for one task.

State that invariant carefully, because the natural phrasing is false. It is not "no `await` sits between the read and the commit" -- there are `await` tokens in between, in `on_task_cancel` and `on_task_cancel_confirm`, which read `current_state()` and then `await self.__routing(...)`. It holds anyway because `await` on a coroutine is not a suspension point: the callee's body starts running immediately on the same stack, and the loop gets control back only at something that actually blocks. That is a fact about what the callee does at runtime, not about the text at the call site, so no reviewer can check it by reading for `await`.

Reverse the order and both events act. Driving a `TaskResult` whose client send parks mid-flight, then a `WorkerDisconnected` for the same task:

```
transitions committed on ONE machine:
    WorkerDisconnected     running -> running     <- reroutes the task to a second worker
    TaskResultReceived     running -> success     <- reports success to the client
```

No exception is raised anywhere. The client is told the task succeeded and the same task is handed to a second worker, which is at-least-once execution reached through a door the scheduler has no fence for. This is not hypothetical: the scheduler gathers its binder dispatch, the balance loop and the worker cleanup loop as separate coroutines, so a heartbeat timeout can call `on_worker_disconnect` for a task whose result is mid-send. Three smaller faults follow from the same cause: the client gets two terminal messages, `on_task_done` releases the same slot twice, and `_statistics[source]` decrements twice against one increment, so a gauge goes negative.

## A per-task lock

Each `TaskStateMachine` gets an `asyncio.Lock`, held for the whole transition, and the source state is read inside it. The action still runs before the commit, and the lock is what stops a second event reading a source state the first has decided to leave. The lock lives on the machine, so it is created and destroyed with it and there is no side table to leak.

`__acquire_machine` folds every reason an event can fail to reach a machine into one contract -- **if it returns `None`, no lock is held** -- and `__locked_machine` releases in a `finally`, so a normal completion, an early return and a raised action all free the lock from one place. That last one matters most: the router swallows action exceptions to keep the scheduler alive, and a swallowed exception that also stranded the lock would deadlock every later event for that task.

The re-check after acquiring compares by identity rather than against `None`, because the property the router depends on is that the machine it locked is still the registered one. It is live, not defensive: a holder whose action raised removes a machine that still reads its original, non-terminal state, so a waiter would find its event perfectly legal and send a cancel or a result for a task the scheduler has just torn down. `remove_state_machine` now requires the caller to hold the lock, and logs if it does not: the re-check covers the window between reading the mapping and taking the lock, and that rule covers the window after it. Neither subsumes the other.

**A bounded failure mode.** The acquire has a 10 second timeout. A transition holds the lock for a few sends and no I/O waits, so this is not a tuning knob balancing throughput against latency; it is the point past which the only remaining explanation is a deadlock. It sits well below the 60 second worker and client timeouts, so a stuck lock reports itself as a logged drop naming the task and its debug path, rather than surfacing later and less legibly as a worker being declared dead. If it ever fires, the answer is to fix the cycle, not to raise the number.

The lock cannot deadlock because the acquisition graph has no cycle. Fan-out only happens on terminal transitions, and everything reached is either a non-terminal transition that fans out to nothing, or a brand-new task with no machine to lock. The one recursive edge is `__cancel_whole_graph` cancelling siblings, and it is closed twice and independently: tasks remove themselves from `running_task_ids` before the fan-out, and `graph_info.status` is set before the `gather`. Either one alone is enough. Both guards now carry comments naming the other and saying that the per-task locks depend on them, since they live in a different file from the lock.

## Failing a faulted transition to its client

Running the action first means a raised action leaves partial work behind with no commit, so this can no longer be deferred. The router still swallows the exception -- it runs from both timer loops and message handlers, and an escape stops the scheduler through `asyncio.gather` -- but it now also fails the task to its client rather than leaving it live in its source state, where the machine and the payload leak and the client hangs until `client_timeout_seconds` drops the result anyway.

It is reported as an ordinary `TaskResult(failed)` carrying a `serialize_failure(SchedulerError(...))` payload written through the storage connector. A new `TaskResultType` was the obvious alternative and is unusable: `future_manager.on_task_result` pops the future *before* its if/elif chain and the final `else` raises, and the client agent's loops catch only `CancelledError` and `KeyboardInterrupt`, so one scheduler-side fault on one task would take down the whole client agent and fail every other outstanding future. The existing type needs no protocol change and every existing client understands it.

Telling the client goes first and each teardown step is guarded on its own. Reusing `__send_task_result_to_client` would not work: that helper releases the worker before it tells the client, so an action that raised inside `on_task_done` would fail the same way again and the client would never hear anything.

The teardown then has to leave the controller's own bookkeeping consistent, which is two things beyond dropping the machine.

**It takes the task out of `_unassigned` as well as out of `_task_id_to_task`.** Those two are removed together or not at all, because `__acquire_workers` reads `_task_id_to_task[_unassigned[0]]` on every drain. An id left queued with no payload raises `KeyError` at the head, and nothing pops the head, so it raises on every drain after that and no queued task is ever dispatched again. It does not stay contained: `__retry_unassignable` runs from inside the action of every terminal transition, so each later task then raises inside its own action and is torn down by this same path. The window is one `await` wide -- `__acquire_and_dispatch` appends to `_unassigned` and then awaits `__send_monitor` -- so a monitor publisher failing during a reroute with no free worker reaches it. `__on_task_cancel` already does this removal before terminating a queued task, so the teardown was the only removal path that did not.

**It commits `failed` before removing the machine.** `commit` is the only place a count moves between states, so removing a machine without one leaves `_statistics` counting a task that no longer exists, and `scaler top` reports a permanently running task per faulted transition. `failed` is a state the schema already has, it is what the client was just told, and it records the fault in the debug path. This is separate from the faulted-transition counter below, which is a new enum member and stays deferred.

## What this fixes

**`canceling` + `workerDisconnect` crashed the scheduler.** The pair routes to `canceled`, whose handler takes a `TaskCancelConfirm`, while `on_worker_disconnect` passes a `worker_id`. Python binds the arguments at the call expression, so the coroutine is never constructed and the handler body never runs. The machine has already moved to `canceled`. The `TypeError` escaped through `__disconnect_worker`, so the other tasks of the dead worker never received their disconnect either, and it then terminated the scheduler through `asyncio.gather`. The path needs a client cancel of a running task whose worker dies before it can confirm.

**A worker dying during a cancel logged a spurious error per task.** `__state_canceled` assumed the task still occupied worker capacity, but `__disconnect_worker` calls `remove_worker` *before* it routes the transition, which drops every `task_id -> worker` mapping. The later `on_task_done` then wrote `Cannot find task in worker queue` at error level. `__state_inactive` guards against this with `previous_state()`; `__state_canceled` could not, because both of its inbound paths come from `canceling`. The disconnect arm now owns that knowledge directly.

**The double transition above**, and the partial work an action leaves behind when it raises.

## Removed

`TaskTransition` (declared in `common.capnp`, but no message field ever had that type), `TRANSITION_MAP`, `previous_state()`, the `_state_functions` registry and the `**kwargs` dispatch, `_task_result_transition_map`, the `assert current_state() == ...` line in every handler, and the two hand-written guards in `on_task_cancel` / `on_task_cancel_confirm` along with the `get_worker_by_task_id` lookup that existed only to fill a parameter its receiver ignored.

Two dead states go with them. `workerDisconnecting` existed only to hold a reroute-or-fail decision that always answers "reroute": `schedulerHasNoTask -> failedWorkerDied` was unreachable, because `__state_inactive` fills `_task_id_to_task` and only the two terminal senders remove from it, each directly after `remove_state_machine` with no `await` between. `balanceCanceled` was in the enum and in `TaskStateManager.TASK_STATES` but was never a target in `TRANSITION_MAP` and had no entry in `_state_functions`, so the scheduler reported a count that was always zero. The uncalled `is_running` / `is_canceling` / `is_finished` / `is_canceled` / `is_done` API goes too.

`failedWorkerDied` stays unreachable -- nothing constructs a `TaskResult(failedWorkerDied)` -- but it is kept, since a reroute limit would give it a successor.

## The graph as a derived artifact

Deleting `TRANSITION_MAP` costs one readable copy of the graph, so `tests/scheduler/controllers/task_state_graph.txt` recovers it. A test drives every (state, event) pair through the real router with stubbed collaborators, records the state that comes back, and compares the result against the checked-in file. A routing change appears as a diff in review and a rejected pair appears as a visible `(rejected)` row rather than an absent key. A table can drift from the code that reads it; a snapshot changes only when behavior changes. It earned its place during this work: it caught a two-line behavior change that would otherwise have shipped silently. The reachability checks -- every state reachable from `inactive`, every live state reaching a terminal state -- read that same file.

Exhaustiveness needs no test: a missing `case` arm is a type error naming the state before the tests run.

`main` has no unit test for `TaskStateMachine` or `TaskStateManager` at all; only `test_dead_worker` and `test_balance` reach them, end to end and incidentally. Alongside the snapshot there are behavior tests that pin the semantics rather than the shape, and concurrency tests, each of which has to force the interleaving by parking an action mid-flight -- the unforced version passes vacuously, because two transitions that never overlap prove nothing about what happens when they do. Those cover the double transition, the lock being released when an action raises, an event arriving while the task reaches a terminal state, an event arriving while the teardown removes the machine (the one test that pins the identity re-check, since the terminal case is refused by the source state on its own), and an action that raises reaching the client as a failed result.

`mypy_path = "src"` is added so that checking a single file under `tests/` resolves `scaler.protocol.capnp` instead of silently degrading it to `Any`, which is what an IDE checking one file does. CI already had this; a reader did not.

## Not in scope

**The `balanceCanceling` dead end stays.** A balance cancel whose worker no longer holds the task has no accepted transition out of `balanceCanceling`, so the task is stranded with no worker, no cancel in flight and no exit path. Per-transition actions surface it as a declared `case` arm with a `FIXME` and a reason, where the table hid it as a missing key, but closing it means choosing between terminating the task towards its client and rescheduling it, and those differ in what the client sees. That is a behavior change and does not belong in a restructure. `test_balance_cancel_confirm_not_found_strands_the_task` and `test_balance_cancel_without_a_worker_strands_the_task` pin the current behavior so the follow-up has to replace them deliberately.

**The faulted-transition gauge.** `TaskManagerStatus` is keyed by `TaskState` and its schema is compiled into the C++ pymod, so a new counter needs a capnp change and a rebuild. The follow-up that splits the gauges from the counters already requires that change, so it belongs there.

**Pinning the two `__cancel_whole_graph` guards** needs an integration test against the real `VanillaGraphTaskController`. The timeout is the backstop in the meantime, and both guards carry comments naming each other, including the `running_task_ids` removals in `__mark_node_done` and `__mark_node_canceled` that a reader editing those two functions would have no reason to look for.

## Notes for review

- One arm departs from today's behavior beyond the fixes above. `canceling` + `CancelConfirmNotFound` now calls `on_task_done`. The worker answered that it does not hold the task, but the scheduler still maps the task to that worker, and the state it lands in is terminal, so without the release that worker's capacity leaks.
- The scheduler still dispatches `TaskResult` and `TaskCancelConfirm` without the identity of the sender, so a task rerouted after a false-positive worker timeout can run on two workers and the scheduler will accept whichever result arrives first. That is a separate change and is untouched here.
- Statistics still mix gauges and counters in one `TaskManagerStatus` map. That is the follow-up named above.
- A client older than the release that adds `SchedulerError` cannot unpickle it and fails inside `deserialize_failure`. That is a degraded failure on one future rather than a fatal one on the whole agent.


src/scaler/scheduler/controllers/mixins.py
```diff
@@ -84,7 +84,7 @@ def on_task_begin(self, client_id: ClientID, task_id: TaskID):
         raise NotImplementedError()
 
     @abc.abstractmethod
-    def on_task_finish(self, task_id: TaskID) -> Optional[bytes]:
+    def on_task_finish(self, task_id: TaskID) -> Optional[ClientID]:
```
magniloquency, line 87: Technically unrelated, but good to fix while we're here

tests/scheduler/controllers/task_state_graph.txt
rafa-be, line 1: Shouldn't this not be pushed to Github and be in .gitignore?

magniloquency: We moved from making the transitions declarative to being defined by functions. It's a good change, but this file is the transition table derived from the those functions. I'm good with either gitignoring this or just deleting it + the test, let me know.

.github/workflows/fork-wheels.yml
rafa-be, line 1: Don't forget to remove this before merge.

src/scaler/scheduler/task/task_state_manager.py
```diff
@@ -1,76 +1,67 @@
...
         """
 
-        task_state_machine = self._task_id_to_state_machine.get(task_id, None)
-        if task_state_machine is None:
-            logger.error(f"{task_id!r}: unknown {transition=} for non-existed state machine")
-            return None
+        state_machine = self._task_id_to_state_machine.get(task_id, None)
+        if state_machine is None:
+            logger.error(f"{task_id!r}: cannot commit {event_type.__name__} for non-existent state machine")
+            return
```
rafa-be, line 54: Might be great to check the `state_machine`'s lock is acquired here:

```python
        if state_machine is not None and not state_machine.lock.locked():
            logger.error(f"{task_id!r}: state machine state changed without its lock held")
```

src/scaler/utility/exceptions.py
```diff
@@ -18,6 +18,16 @@ class ProcessorDiedError(Exception):
     pass
 
 
+class SchedulerError(Exception):
+    """The scheduler failed while it was applying a transition for this task, so the task cannot continue.
+
+    This is a scheduler-side fault rather than a failure of the task's own code. It is reported as an ordinary failed
+    result so that every existing client understands it: a new TaskResultType would reach the client's final `else`
+    and raise, after the future was already popped, taking down the whole client agent."""
```
rafa-be: Not sure if that comment is useful. These exceptions are user-facing and give too much implementation details.

src/scaler/scheduler/controllers/task_controller.py
```diff
@@ -25,19 +27,68 @@
...
 logger = logging.getLogger(__name__)
 
+# A transition holds its machine's lock for a few sends and no I/O waits, so this is not a tuning knob balancing
+# throughput against latency: it is the point past which the only remaining explanation is a deadlock. It sits well
+# below the 60 second worker and client timeouts, so a stuck lock reports itself as a specific error naming the task
+# rather than resurfacing later as a worker or client being declared dead. If it ever fires, fix the cycle.
+LOCK_ACQUIRE_TIMEOUT_SECONDS = 10
+
+# The state that each action can produce. mypy checks every return against these, so the behavior of an action cannot
+# widen without its annotation widening too.
```
rafa-be, line 62: That's nice 👌

src/scaler/utility/exceptions.py
magniloquency: Removed it. Also maybe worth noting that I want to improve this later: https://github.com/finos/opengris-scaler/issues/967

src/scaler/scheduler/task/task_state_manager.py
magniloquency: Done

src/scaler/scheduler/controllers/task_controller.py
```diff
@@ -208,149 +206,413 @@ def get_status(self) -> TaskManagerStatus:
...
+                | TaskState.canceling
+                | TaskState.balanceCanceling
+                | TaskState.success
+                | TaskState.failed
+                | TaskState.failedWorkerDied
+                | TaskState.canceled
+                | TaskState.canceledNotFound
+            ):
+                # the task already holds a worker, or it finished before the acquired worker could be used
+                return None
```
rafa-be: Aren't `TaskState.running`, `TaskState.balanceCanceling`, `TaskState.success` unexpected states for this event, and should be logged as errors?

e117649 (approved):

PR #953, opened 2026-08-21
fix: GIL dead-lock when a ymq socket is torn down at interpreter exit

### LINKED ISSUE #945: GIL contention dead-lock when `__del__()` a YMQ socket.
- At interpreter shutdown, an undestroyed `YMQAsyncBinder.__del__()` calls `YMQAsyncBinder.destroy()`, which calls blocking `BinderSocket.shutdown()`.
- Python waits in `PyBinderSocket_shutdown()` on `future.wait()` until C++ signals shutdown completion.
- But `BinderSocket::shutdown()` drains pending callbacks first; those callbacks re-enter Python and acquire GIL via `AcquireGIL` from a non-Python thread during finalization.
- During finalization that GIL acquisition can never complete, so shutdown completion is never signaled, and the process hangs in `shutdown()`

Very simple reproducer:

```python
import asyncio
from scaler.config.types.address import AddressConfig
from scaler.io.ymq import IOContext
from scaler.io.ymq_async_binder import YMQAsyncBinder

LEAK = {}

async def main():
    b = YMQAsyncBinder(IOContext(), b"x", lambda *_: None)
    await b.bind(AddressConfig.from_string("tcp://127.0.0.1:0"))
    asyncio.create_task(b.routine())  # leaves one pending recv callback
    LEAK["b"] = b                    # force __del__ at interpreter exit

asyncio.run(main())
```

magniloquency:
Fixes #945.

## What happens

When a YMQ socket is destroyed, the C++ tries to fail callbacks that require acquiring the GIL and calling into Python. Unfortunately if Python is shutting down, the GIL cannot actually be acquired which can crash threads and cause hangs.

## The fix

`AcquireGIL` will check if Python is exiting and indicate it, callers must check the flag and perform some reasonable exit behavior, usually this means doing nothing or almost nothing.

The flag is set from an `atexit` callback registered when `_ymq` is imported, rather than by calling `Py_IsFinalizing()` where the GIL is taken. `Py_IsFinalizing()` can flip between the check and the acquire, whereas `_PyAtExit_Call()` runs before `_PyRuntimeState_SetFinalizing()`, making it the safer choice.

## Affected versions

Deterministic on 3.10 through 3.13. Versions 3.14+ changed the rules for this slightly which can cause it to appear to pass, but this is still something we need to fix, and the CI test job pins 3.10. Note that reverting the fix on a 3.14 machine will not reproduce it -- the new tests pass there without proving anything.

## Commits

The first commit adds the regression tests on their own and is expected to fail CI. The second fixes them.

## Notes for review

- `PyBufferBytes::~PyBufferBytes()` leaks a `Py_buffer` on the shutdown path. Acceptable at exit, but
  it is a deliberate choice rather than an oversight.
- Belt-and-braces option not taken: `PyBinderSocket_shutdown` could also skip its `future.wait()` when
  the flag is set. Unnecessary once the callbacks stop blocking on the GIL, but it would remove the
  last route from a bug elsewhere to a hang.
- Does not cover `os._exit()` or embedded interpreters that never run `atexit`. Degraded there, not
  worse than today.


src/cpp/scaler/ymq/pymod/binder_socket.h
```diff
@@ -189,7 +191,9 @@ static PyObject* PyBinderSocket_send_message(PyBinderSocket* self, PyObject* arg
             std::move(messagePayload->bytes),
             [callback_ = OwnedPyObject<>::fromBorrowed(callback), state](
                 std::expected<void, scaler::ymq::Error> result, std::unique_ptr<scaler::ymq::Bytes> payload) {
-                AcquireGIL _;
+                AcquireGIL gil;
+                if (!gil.acquired())
+                    return;  // interpreter is shutting down, do not re-enter Python
```
rafa-be, line 197: Does this and Python killing threads acquiring GIL on shutdown mean that `SocketStopRequested` errors were/will never be propagated back to Python on shutdown?

Maybe that's fine, but we should be aware of this.

src/cpp/scaler/utility/pymod/gil.h
rafa-be, line 1: I'm not sure embedding the `interpreterIsShuttingDown` logic within `AcquireGIL` is best.

I personally would prefer to use `interpreterIsShuttingDown()` explicitly, and keep `AcquireGIL` as close to `PyGILState_Ensure`/ `PyGILState_Release` as possible.

That is, explicitly do this:

```python
if (interpreterIsShuttingDown)
    return;  // interpreter is shutting down, do not re-enter Python 

AcquireGIL gil;
```

instead of this:

```python
AcquireGIL gil;
if (!gil.acquired())
    return;  // interpreter is shutting down, do not re-enter Python
```

But that's not a big deal anyway.

src/cpp/scaler/utility/pymod/gil.h
```diff
@@ -3,27 +3,53 @@
...
+
+// Acquires the GIL, unless the interpreter is shutting down, in which case acquiring it would kill the calling
+// thread. Callers must check `acquired()` before touching any Python state.
 class AcquireGIL {
 public:
-    AcquireGIL(): _state(PyGILState_Ensure())
+    AcquireGIL(): _acquired {!interpreterIsShuttingDown().load(std::memory_order_acquire)}
     {
+        if (_acquired)
+            _state = PyGILState_Ensure();
```
rafa-be: There might be a very rare event where the interpreter shutdowns during the call to this constructor

```c++
_acquired {!interpreterIsShuttingDown().load(std::memory_order_acquire)}

// <-- Interpreter shutting down from another thread here 

if (_acquired)
    _state = PyGILState_Ensure();  // Python kills this thread (<= 3.13) or suspend it (>= 3.14).
```

Not sure if there exists a workaround.

src/cpp/scaler/ymq/pymod/binder_socket.h
magniloquency: I think yes, even before this change they weren't delivered when the interpreter was shutting down, it's just that the ymq threads were most likely being terminated by python instead of shutting down 'gracefully'.

src/cpp/scaler/utility/pymod/gil.h
magniloquency: I think it's safer this way, it enforces that it must be checked before attempting to acquire the GIL. It's also used in the destructor to avoid releasing the lock. That could also be solved another way, but I think the way it's used here is reasonable. The check is specifically needed exactly when you're acquiring the GIL, there's a semantic basis for keeping them coupled. If you still prefer to decouple them though, I will.

src/cpp/scaler/utility/pymod/gil.h
magniloquency: Ok I looked into this and we can make it more robust. The way it works is we put a semaphore (in reality an atomic) around acquiring and releasing the GIL, and the finalizer code prevents the interpreter from shutting down until we can be sure no threads are in the [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) window. I have created a mini PR to show how this works before we decide if we want to include it here or not. Here's the code, if it looks good to you, I'll polish it up (fix the overly long comments, style, etc.) and add it to this PR.

https://github.com/magniloquency/scaler/pull/35

rafa-be: The fix makes sense.

It adds a lot of overhead though (3 atomic reads-writes at every GIL acquire, not only during shutdown). Not sure what is best. The issue and its fix are legit.

@e117649 opinion on this?

src/cpp/scaler/ymq/pymod/binder_socket.h
rafa-be: That might be fine.

But maybe someday we should double-check we don't rely on the expectation these callbacks are called during the shutdown (e.g. to send a client shutdown message).

rafa-be (dismissed):

src/cpp/scaler/utility/pymod/gil.h
e117649: I agree the fix makes sense, I measured the overhead and it seemed negligible (<3%), a couple suggestions for it:

1. Drop the poll for `std::atomic::wait/notify_all` (C++20), notifying only when the shutdown flag is set. The hot path keeps exactly the two RMWs, the drain wakes immediately instead of up to 200 µs late, and the 200 µs constant goes away, keep the 1 s cap as is  
2. Centralise the flag and the counter

Also I agree with the future work you mentioned to "stop the io threads from the atexit hook"

magniloquency: I spent a lot of time thinking about this and I landed on a solution using semaphores. I think it's pretty good and I've gone ahead and merged it into this PR. You can take a look at this commit https://github.com/finos/opengris-scaler/pull/953/commits/4c39f812c26bfcd9deb6dc92d8f4f0ff02f42559 and let me know what you think.

I'm not sure exactly what you meant by #2 but I did some refactoring so maybe I fixed it?

As for stopping the io threads from the atexit hook. I agree, that would be nice for the future. I can maybe make an issue for that.

e117649 (approved):

rafa-be (approved):

PR #960, opened 2026-08-21
fix: Client uses a non-blocking OSS connector (fixes #948)

### LINKED ISSUE #948: Client might timeout during a large OSS transaction
The current client implementation use a synchronous OSS connector, meaning the client's agent asyncio loop might block for a significant amount of time, and preventing the client to send its heart-beats to the scheduler.

If the OSS transaction is long enough, the client will be disconnected by the scheduler and the task will fail.

rafa-be:
With these changes, the **client and its agent use a non-blocking `AsyncObjectStorageConnector` instead of a `SyncObjectStorageConnector`**.

This prevents the client agent to hang during a long lasting OSS transaction, possibly preventing heart-beats to be transmitted / answered, and ultimately trigerring a client timeout (see #948).


src/scaler/client/agent/bridge.py
```diff
@@ -18,65 +18,34 @@
...
 
 
-class ClientAgentBridge(abc.ABC):
-    """Bridges a synchronous ``Client`` to an asynchronous ``ClientAgent``.
-
-    Implementations encapsulate the lifecycle of the agent (start/stop/wait)
-    and expose a ``SyncConnector``-compatible handle that delivers messages
-    from the ``Client`` to the agent's receive handler.
-    """
-
```
rafa-be, line 49: I had to move the `ClientAgentBridge` abstract-class to `client/agent/mixins.py` to avoid a cycling dependency.

That's because `ScalerFuture` now requires `ClientAgentBridge`, whose implementations requires `ClientAgent`, and ultimately `ScalerFuture`. Moving the ABC to `mixins.py` fixes the issue.

tests/client/test_client_heartbeat.py
```diff
@@ -0,0 +1,92 @@
...
+            SlowSerializer.busy_wait()
+
+        return obj
+
+    @staticmethod
+    def busy_wait() -> None:
+        # Busy loops instead of sleeping, as (de)serializing a large object is CPU bound.
+        deadline = time.monotonic() + SERIALIZATION_DURATION_SECONDS
+        while time.monotonic() < deadline:
+            pass
```
rafa-be, line 58: At first I triggerred the issue using large objects (4 GB), but that caused issues on the CI, and wasn't entirely deterministic on my laptop.

I went for a very slow serializer instead. As the serializer is (currently) being called from the connectors, it simulates the behavior of a very slow network.

rafa-be:
Fixes #948

tests/client/test_client_heartbeat.py
```diff
@@ -0,0 +1,92 @@
...
+            self.address, serializer=SlowSerializer(), heartbeat_interval_seconds=CLIENT_HEARTBEAT_INTERVAL_SECONDS
+        ) as client:
+            # Serializing the argument, and deserializing the result, both take longer than the scheduler's client
+            # timeout.
+            future = client.submit(slow_task, SlowObject())
+            self.assertIsInstance(future.result(timeout=RESULT_TIMEOUT_SECONDS), SlowObject)
+
+            # The client must still be connected to the scheduler, i.e. it kept exchanging heartbeats while fetching the
+            # result object.
+            self.assertEqual(client.submit(round, 3.14).result(timeout=RESULT_TIMEOUT_SECONDS), 3)
```
e117649:
```suggestion
            results = client.map(slow_task, [SlowObject()])
            self.assertIsInstance(results[0], SlowObject)
```

e117649: use a non-delayed future so the fetch is driven from set_result_ready on the agent thread

src/scaler/client/object_buffer.py
e117649, line 85:
```suggestion
            self._cycle_dedup_cache.clear()
            return
```

e117649: pre-existing adjacent issue I noticed while reviewing

src/scaler/client/agent/bridge.py
```diff
@@ -153,6 +122,22 @@ def is_alive(self) -> bool:
     def join(self) -> None:
         self._agent.join()
 
+    def run_in_agent(self, coroutine_factory: Callable[[], Coroutine]) -> concurrent.futures.Future:
```
magniloquency: Why not just have this fn take the coroutine directly?

src/scaler/client/agent/bridge.py
```diff
@@ -18,65 +18,34 @@
...
-    @abc.abstractmethod
-    def is_alive(self) -> bool:
-        """Return True if the agent is still running."""
-
-    @abc.abstractmethod
-    def join(self) -> None:
-        """Wait for the agent to fully stop. Safe to call multiple times."""
-
-
 class IPCAgentBridge(ClientAgentBridge):
```
magniloquency, line 51: Maybe this isn't so important now but this name tripped me up. IPC usually means inter-process communication, but the client agent is actually a thread, not another process. Technically IPC could also stand for intra-process communication which would make it correct, but I don't think that's the way the acronym is typically used. By that same token, the in-process bridge is also kind of named incorrectly, they're both in-process.

src/scaler/client/agent/bridge.py
```diff
@@ -153,6 +122,22 @@ def is_alive(self) -> bool:
     def join(self) -> None:
         self._agent.join()
 
+    def run_in_agent(self, coroutine_factory: Callable[[], Coroutine]) -> concurrent.futures.Future:
+        loop = self._agent.get_loop()
```
magniloquency: I don't really like that we're reaching into the agent's internals here. Could we create a `agent.submit(coroutine)` method?

src/scaler/client/client.py
magniloquency: This is no longer being used. Perhaps it also means the oss addr doesn't have to be a future?

src/scaler/client/agent/future_manager.py
```diff
@@ -70,23 +70,24 @@ def on_task_result(self, result: TaskResult):
...
+            future.set_result_ready(result_object_id, TaskState.success, profile_result)
 
-            else:
-                raise TypeError(f"{result.taskId.hex()}: Unknown task status: {result.resultType}")
+        elif result_type == TaskResultType.failed:
+            result_object_id = self.__get_result_object_id(result)
+            future.set_result_ready(result_object_id, TaskState.failed, profile_result)
+
+        else:
+            raise TypeError(f"{result.taskId.hex()}: Unknown task status: {result.resultType}")
```
magniloquency: Minor: Style: An exhaustive `match` would provide better type-safety

src/scaler/client/object_buffer.py
```diff
@@ -134,6 +154,24 @@ def clear(self):
     def is_valid_object_id(self, object_id: ObjectID) -> bool:
         return object_id in self._valid_object_ids
 
+    async def __set_object(self, object_id: ObjectID, payload: bytes) -> None:
+        connector_storage = await self._bridge.object_storage_connector()
+        await connector_storage.set_object(object_id, payload)
+
+    async def __get_object(self, object_id: ObjectID, is_exception: bool, delete_after_fetch: bool) -> Any:
```
magniloquency, line 170: Please document the meaning of `is_exception`

src/scaler/client/agent/bridge.py
magniloquency: Typing: Can we make this return `Future[T]` for some `T`? (likely comes from the input coroutine / factory?)

src/scaler/client/future.py
```diff
@@ -248,36 +245,103 @@ def _on_waiters_updated(self, waiters: EventList):
...
+                return  # the object is already being fetched
 
             if self._task_state == TaskState.success:
-                self.set_result(self._serializer.deserialize(object_bytes))
+                is_exception = False
             elif self._task_state == TaskState.failed:
-                self.set_exception(deserialize_failure(object_bytes))
+                is_exception = True
             else:
                 raise ValueError(f"unexpected task status: {self._task_state}")
```
magniloquency: Minor: Style: An exhaustive match here would provide better type safety

src/scaler/client/agent/client_agent.py
```diff
@@ -68,11 +68,16 @@ def __init__(
         self._client_agent_address = client_agent_address
         self._scheduler_address = scheduler_address
         self._network_backend = network_backend
+
         self._object_storage_address: Future[AddressConfig] = Future()
```
magniloquency: Can we fully qualify this future type? We use asyncio futures and concurrent futures in different places, so I think it's good to be clear.

src/scaler/client/future.py
```diff
@@ -248,36 +245,103 @@ def _on_waiters_updated(self, waiters: EventList):
...
+
+            self._result_object_future = self._object_buffer.fetch_object(
+                self._result_object_id, is_exception=is_exception, delete_after_fetch=delete_after_fetch
+            )
+
+        self._result_object_future.add_done_callback(self.__on_result_object_fetched)
+
+    def __on_result_object_fetched(self, _: concurrent.futures.Future) -> None:
+        assert self._result_object_future is not None and self._result_object_future.done()
+        is_exception = self._task_state == TaskState.failed
```
magniloquency: This seems to be derived once when creating the future, then again in the future callback? I think typically you would solve this by reusing the local in straight-line async/await code, but I guess that's not what we have here.

src/scaler/client/future.py
```diff
@@ -248,36 +245,103 @@ def _on_waiters_updated(self, waiters: EventList):
...
+
+        try:
+            result_object = self._result_object_future.result()
```
magniloquency, line 298: Question... is `_result_object_future` the same as `_` in the parameters?

src/scaler/client/future.py
```diff
@@ -248,36 +245,103 @@ def _on_waiters_updated(self, waiters: EventList):
...
+        if self._result_object_id is None:
+            return  # umbrella graph tasks do not have a result object
+
+        while not self._result_received and not self.cancelled():
+            if deadline is None:
+                remaining_seconds = None
+            else:
+                remaining_seconds = deadline - time.monotonic()
+                if remaining_seconds <= 0:
+                    raise concurrent.futures.TimeoutError()
```
magniloquency:
```suggestion
                    raise concurrent.futures.TimeoutError
```

src/scaler/client/future.py
```diff
@@ -248,36 +245,103 @@ def _on_waiters_updated(self, waiters: EventList):
...
+        try:
+            if is_exception:
+                self.set_exception(result_object)
+            else:
+                self.set_result(result_object)
+        except concurrent.futures.InvalidStateError:
+            # The future got canceled while its result object was being fetched, e.g. by `Client.disconnect()`.
+            pass
+
+    def _wait_result_object(self, timeout: Optional[float] = None):
```
magniloquency: Return typing

src/scaler/client/future.py
```diff
@@ -248,36 +245,103 @@ def _on_waiters_updated(self, waiters: EventList):
...
             # if it's delayed future, get the result when waiter gets added
             if self._is_delayed and len(self._waiters) > 0:
-                self._get_result_object()
+                self._start_result_object_fetch()
 
     def _has_result_listeners(self) -> bool:
         return len(self._done_callbacks) > 0 or len(self._waiters) > 0  # type: ignore[attr-defined]
 
-    def _get_result_object(self):
+    def _start_result_object_fetch(self):
```
magniloquency: Return typing

src/scaler/client/object_buffer.py
```diff
@@ -95,15 +96,34 @@ def commit_send_objects(self):
...
+        # All of the uploads are started before waiting for any of them, so that these pipeline instead of being
+        # uploaded one round-trip at a time.
+
+        pending_uploads = [
+            self._bridge.run_in_agent(partial(self.__set_object, obj_cache.object_id, obj_cache.object_payload))
+            for obj_cache in self._pending_objects
+        ]
+
+        for pending_upload in pending_uploads:
+            pending_upload.result()
```
magniloquency: Makes ignoring the result explicit and should satisfy some of the more strict type checkers

```suggestion
            _ = pending_upload.result()
```

src/scaler/client/object_buffer.py
```diff
@@ -95,15 +96,34 @@ def commit_send_objects(self):
...
         self._cycle_dedup_cache.clear()
 
+    def fetch_object(self, object_id: ObjectID, is_exception: bool, delete_after_fetch: bool) -> Future:
+        """
+        Starts fetching and deserializing an object from the object storage server.
+        """
+
+        # Do the actual object download within the agent's asyncio loop
+
+        return self._bridge.run_in_agent(partial(self.__get_object, object_id, is_exception, delete_after_fetch))
```
magniloquency: Coordinating this comment with my other one: Why not just call `self.___get_object` here?

magniloquency (changes requested):

src/scaler/client/future.py
```diff
@@ -248,36 +245,103 @@ def _on_waiters_updated(self, waiters: EventList):
...
+
+        while not self._result_received and not self.cancelled():
+            if deadline is None:
+                remaining_seconds = None
+            else:
+                remaining_seconds = deadline - time.monotonic()
+                if remaining_seconds <= 0:
+                    raise concurrent.futures.TimeoutError()
+
+            self._condition.wait(remaining_seconds)  # type: ignore[attr-defined]
```
e117649, line 354:
```suggestion
            if sys.platform == "emscripten":
                # The agent runs on this very thread under Pyodide, so ``Condition.wait`` would
                # block the only thread that can complete the fetch. Suspend the wasm stack on the
                # fetch itself instead, releasing the condition so its callback can set the result.
                fetch = self._result_object_future
                self._condition.release()  # type: ignore[attr-defined]
                try:
                    if fetch is not None:
                        fetch.exception(remaining_seconds)
                except concurrent.futures.CancelledError:
                    pass  # the cancellation is reported by super().result() / super().exception()
                finally:
                    self._condition.acquire()  # type: ignore[attr-defined]
                continue

            self._condition.wait(remaining_seconds)  # type: ignore[attr-defined]
```

tests/client/test_client_heartbeat.py
```diff
@@ -0,0 +1,92 @@
...
+
+        RESULT_TIMEOUT_SECONDS = 60
+
+        with Client(
+            self.address, serializer=SlowSerializer(), heartbeat_interval_seconds=CLIENT_HEARTBEAT_INTERVAL_SECONDS
+        ) as client:
+            # Serializing the argument, and deserializing the result, both take longer than the scheduler's client
+            # timeout.
+            future = client.submit(slow_task, SlowObject())
+            self.assertIsInstance(future.result(timeout=RESULT_TIMEOUT_SECONDS), SlowObject)
```
e117649, line 89:
```suggestion
            results = client.map(slow_task, [SlowObject()])
            self.assertIsInstance(results[0], SlowObject)
```

src/scaler/client/future.py
rafa-be: ~Actually, it has to be derived twice, as an exception might have occurred while fetching the object, even if the original task succeeded.~

Done

src/scaler/client/future.py
rafa-be: Yes. It's a little bit annoying, Python callbacks usually don't have this kind of signature and rely on lambda capture.

src/scaler/client/agent/bridge.py
rafa-be: The issue is that it's a Python error to create a co-routine but never await on it.

I rewrote run_in_agent() in such a way that it does not use a factory but destroy the coroutine if the loop is closing.

src/scaler/client/agent/bridge.py
rafa-be: Yeah I kind of agree, but out of scope of this PR. Let's keep this in mind for #975

src/scaler/client/agent/bridge.py
rafa-be: Done

src/scaler/client/client.py
rafa-be: Done.

src/scaler/client/object_buffer.py
rafa-be: Added the docstring to `fetch_object()`.

src/scaler/client/object_buffer.py
rafa-be: Done

src/scaler/client/future.py
rafa-be: Done

src/scaler/client/future.py
rafa-be: Done

src/scaler/client/future.py
rafa-be: Done

src/scaler/client/object_buffer.py
rafa-be: Done

src/scaler/client/future.py
rafa-be: I solved it slightly differently by reusing a method from `bridge.py`

tests/client/test_client_heartbeat.py
rafa-be: Done, although I solved it slightly differently.

src/scaler/client/future.py
rafa-be: Done

src/scaler/client/agent/future_manager.py
```diff
@@ -75,19 +75,20 @@ def on_task_result(self, result: TaskResult):
...
-        elif result_type == TaskResultType.failed:
-            result_object_id = self.__get_result_object_id(result)
-            future.set_result_ready(result_object_id, TaskState.failed, profile_result)
+            case TaskResultType.failed:
+                future.set_result_ready(self.__get_result_object_id(result), TaskState.failed, profile_result)
 
-        else:
-            raise TypeError(f"{result.taskId.hex()}: Unknown task status: {result.resultType}")
+            case _:
+                raise TypeError(f"{result.taskId.hex()}: Unknown task status: {result.resultType}")
```
magniloquency: Can we use `assert_never` here? It will mean that mypy can check the match for exhaustiveness.

src/scaler/client/future.py
```diff
@@ -264,12 +264,13 @@ def _start_result_object_fetch(self):
...
-                is_exception = True
-            else:
-                raise ValueError(f"unexpected task status: {self._task_state}")
+            match self._task_state:
+                case TaskState.success:
+                    is_exception = False
+                case TaskState.failed:
+                    is_exception = True
+                case _:
+                    raise ValueError(f"unexpected task status: {self._task_state}")
```
magniloquency: `assert_never`

magniloquency (changes requested):

src/scaler/client/agent/future_manager.py
magniloquency: https://github.com/finos/opengris-scaler/blob/afcec811322f869dc3e2273850fd678d5fb21421/src/scaler/scheduler/controllers/task_controller.py#L48-L51

magniloquency (approved):

src/scaler/client/agent/future_manager.py
rafa-be: Done

src/scaler/client/future.py
rafa-be: Done

PR #961, opened 2026-08-24
fix: pickle round trips for PyYMQ exceptions

### LINKED ISSUE #366: PyYMQ exceptions are not serializable
`YMQException` objects are not pickleable:

```ipython
In [1]: from scaler.io.ymq.ymq import YMQException, ErrorCode

In [2]: import pickle

In [3]: e = YMQException(ErrorCode.CoreBug, "Hello!")

In [4]: pickle.dumps(e)
---------------------------------------------------------------------------
PicklingError                             Traceback (most recent call last)
Cell In[4], line 1
----> 1 pickle.dumps(e)

PicklingError: Can't pickle <class 'ymq.YMQException'>: import of module 'ymq' failed
```

However, Cloudpickle works.

This is a problem because the worker might, in some cases, use Pickle to serialize these objects instead of using the client's serializer (that usually is Cloudpickle).

HarperZ9:
Closes #366.

### Summary

- expose `ErrorCode` and PyYMQ exception types through their public
  `scaler.io.ymq` module identity so the standard-library pickle loader can
  resolve them;
- add a pickle round-trip regression for the base `YMQException` and a generated
  subtype, preserving exact type, `args`, `code`, and `message`;
- bump the package patch version from 2.6.0 to 2.6.1 as required by the
  contribution guide.

### Verification

- RED: the new focused test failed for both representative exception types with
  `PicklingError` because `_ymq` was not importable; after normalizing the
  exception module, `ErrorCode` exposed a second invalid module identity.
- GREEN: `python -m unittest tests.io.ymq.test_types`: 8 tests passed.
- GREEN after a fresh native rebuild: `python -m unittest -v
  tests.io.ymq.test_types`: 8 tests passed in 0.005 seconds.
- `isort --profile black --line-length 120 --check-only` on changed Python
  files: passed.
- `black -l 120 -C --check --target-version py312` on changed Python files:
  passed.
- `flake8 --max-line-length 120 --extend-ignore=E203` on changed Python files:
  passed.
- `git diff --check HEAD^ HEAD`: passed.

The corrected full Python discovery run executed 719 tests in 755.726 seconds
with one unrelated scheduler-disconnect timing error and six skips. That exact
test had passed in the prior full run and passed immediately in isolation in
4.937 seconds. This draft does not claim a clean full-Python-suite pass.

The Linux Debug native build completed at 100 percent. The final CTest preset
passed all 6 top-level targets in 44.73 seconds, including the 156-case YMQ test
binary. Four MITM cases that initially exceeded their WSL helper-startup timeout
passed together in isolation after the dependency-bearing interpreter was
warmed.

### AI assistance disclosure

OpenAI Codex assisted with repository inspection, test design, implementation,
verification commands, and pull-request drafting. I reviewed the complete diff
and remain responsible for this contribution.


tests/io/ymq/test_types.py
```diff
@@ -22,6 +23,18 @@ def test_exception(self):
...
+            with self.subTest(exception_type=exception_type):
+                original = exception_type(ErrorCode.SysCallError, "oh no")
+
+                restored = pickle.loads(pickle.dumps(original))
+
+                self.assertIs(type(restored), exception_type)
+                self.assertEqual(restored.args, (ErrorCode.SysCallError, "oh no"))
+                self.assertEqual(restored.code, ErrorCode.SysCallError)
+                self.assertEqual(restored.message, "oh no")
+
```
rafa-be, line 51: Could we add one more test with `cloudpickle`?

HarperZ9: Added a separate `cloudpickle` round-trip test in `9500711` for both `YMQException` and `SysCallError`. It verifies the exact restored type, `args`, `code`, and `message`. The focused module passes locally: 9 tests. Black, isort, and flake8 also pass.

PR #984, opened 2026-08-28
fix: do not hold the GIL while joining the YMQ event loop threads

e117649:
Ctrl-C during startup leaves the whole `scaler` process tree alive with its ports still bound. `test_sigint_during_startup_does_not_hang` already covers this and fails 15 of 20 runs on main.

An event loop thread runs Python callbacks, so it cannot be joined by a thread that holds the GIL:

    main thread                              YMQ event loop thread
      PyIOContext_dealloc, GIL HELD
      -> ~IOContext -> ~EventLoopThread
      -> std::thread::join()  ... blocks ... ConnectorSocket::onClientConnected
                                               -> Python callback -> take_gil

Captured with `py-spy dump --native` and with faulthandler after SIGABRT. The two socket shutdowns already release the GIL for the shutdown future's wait, but take it back one line too early, before the resets that do the joining. `PyIOContext_dealloc` never releases it at all.

A `ReleaseGIL` guard now covers the whole blocking teardown in all three places. RAII rather than `Py_BEGIN_ALLOW_THREADS`, because the macros leave the thread state saved if the code between them throws, and both shutdown functions have a `catch` that calls `PyErr_SetString`.


rafa-be (dismissed):

src/cpp/scaler/ymq/pymod/binder_socket.h
rafa-be: Could there be an issue if one of these callbacks (or another Python thread) call shutdown on the same socket?

e117649: Good catch, fixed in [6d8bd3d](https://github.com/finos/opengris-scaler/pull/984/commits/6d8bd3d2ded54dd4902867eec1dd7a097d232704)

rafa-be (approved):

PR #985, opened 2026-08-28
feat: Web UI pages for clients, workers, objects, task sorting and filtering

### LINKED ISSUE #952: WebUI Improvements
- show task currently being worked on by each worker
- show all tasks (not most recent 500)
- sortable tasks
- memory usage graph should record live total memory usage, instead of summing it from completed tasks
- should also have a CPU usage graph
- show all connected clients and show which client submitted a task

e117649:
closes https://github.com/finos/opengris-scaler/issues/952


src/cpp/scaler/object_storage/object_storage_server.cpp
```diff
@@ -332,17 +332,35 @@ void ObjectStorageServer::processDuplicateRequest(
...
+    for (const auto& [objectID, requests]: pendingRequests) {
+        numPending += requests.size();
+        for (const auto& request: requests) {
+            oldest = std::min(oldest, request.waitingSince);
+        }
+    }
+    const uint64_t oldestPendingSeconds =
+        numPending == 0 ?
+            0 :
+            std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - oldest).count();
```
rafa-be: Could we have a `std::optional<std::chrono::steady_clock::time_point> oldestPending` member field instead? 

That would remove this O(n) loop.

src/scaler/scheduler/controllers/object_controller.py
```diff
@@ -95,6 +106,41 @@ def clean_client(self, client_id: ClientID):
...
+    def object_count(self) -> int:
+        return self._object_tracker.object_count()
+
+    def get_largest_objects(self, limit: int) -> List[ObjectDetail]:
+        """The `limit` biggest tracked objects, biggest first, which is what a full store is made of.
+
+        A heap, because this runs on the scheduler's event loop.
+        Sorting a few hundred thousand objects every status frame costs seconds.
+        """
+        largest = heapq.nlargest(limit, self._object_tracker.items(), key=lambda item: self.get_object_size(item[0]))
```
rafa-be: This has a very big algorithmic complexity

src/scaler/scheduler/controllers/object_controller.py
```diff
@@ -38,10 +44,15 @@ def __init__(self, config_controller: VanillaConfigController):
         self._binder: Optional[AsyncBinder] = None
         self._binder_monitor: Optional[AsyncPublisher] = None
         self._connector_storage: Optional[AsyncObjectStorageConnector] = None
+        # keyed by raw id bytes: ids arrive off the wire, and an ObjectID would validate a length nobody needs
+        self._object_sizes: Dict[bytes, int] = {}
```
rafa-be: Shouldn't that be data-structure handled by `ObjectTracker`?

src/cpp/scaler/object_storage/object_storage_server.cpp
```diff
@@ -325,24 +325,34 @@ void ObjectStorageServer::processDuplicateRequest(
...
     alignas(sizeof(uint64_t)) unsigned char serializedPayload[payloadLength];
 
     const uint64_t numIDs    = objectManager.size();
     const uint64_t numObjs   = objectManager.sizeUnique();
     const uint64_t totalSize = objectManager.totalObjectsSize();
+
+    const uint64_t numPending           = _pendingRequestCount;
+    const uint64_t numPendingIDs        = pendingRequests.size();
+    const auto now                      = std::chrono::steady_clock::now();
+    const auto oldestPending            = _pendingObjectsByAge.empty() ? now : _pendingObjectsByAge.begin()->first;
```
rafa-be: Shouldn't the default no pending object be 0 instead of now ?

src/cpp/scaler/object_storage/object_storage_server.cpp
```diff
@@ -394,6 +422,8 @@ void ObjectStorageServer::optionallySendPendingRequests(
     // Immediately remove the object's pending requests, or else another coroutine might process them too.
     auto requests = std::move(it->second);
     pendingRequests.erase(it);
+    _pendingRequestCount -= requests.size();
+    _pendingObjectsByAge.erase({requests.front().waitingSince, objectID});
```
rafa-be: These 3 lines should go into a new `removePendingRequest()` method, to be symmetrical with `addPendingRequest()` and `clearPendingRequests()`

src/protocol/status.capnp
```diff
@@ -9,15 +9,29 @@ struct Resource {
...
+    numberOfObjects @0 :UInt32;      # objects the scheduler is tracking
+
+    # What the object storage server itself reports, from its infoGetTotal request. All zero until the
+    # first answer arrives.
+    storageObjectCount @1 :UInt64;   # object IDs the server holds
+    storageUniqueCount @2 :UInt64;   # distinct payloads behind them
+    storageTotalBytes @3 :UInt64;    # bytes those payloads occupy
+    storagePendingRequests @4 :UInt64;  # requests waiting for an object that does not exist yet
+    storagePendingObjects @5 :UInt64;   # distinct objects those requests wait for
+    storageOldestPendingS @6 :UInt64;   # how long the oldest of them has waited
```
rafa-be: These `storage*` might be better as a sub-`struct`, what do you think?

src/protocol/status.capnp
```diff
@@ -9,15 +9,29 @@ struct Resource {
...
 
-    struct Pair {
-        client @0 :Data;
+    struct ClientStatus {
+        clientId @0 :Data;
         numTask @1 :UInt32;
+        resource @2 :Resource;   # the client process itself, from its own heartbeat
+        latencyUS @3 :UInt32;    # round trip the client last measured to the scheduler
+        lastSeenS @4 :UInt16;    # seconds since that heartbeat arrived
+        connectedS @5 :UInt32;   # seconds since the client's first heartbeat
```
rafa-be: It's weird that some of these fields have a `S` or `US` postfix, while other have a `Seconds` postfix (see `ProcessorStatus`).

Looks like the old code uses `S` (e.g. in `WorkerStatus`).

src/scaler/scheduler/controllers/worker_manager_utilties.py
```diff
@@ -34,7 +34,7 @@ def build_scaling_manager_status(
             capnp.ScalingManagerStatus.WorkerManagerDetail(
                 workerManagerID=d["worker_manager_id"],
                 identity=d["identity"],
-                lastSeenS=d["last_seen_s"],
+                lastSeenSeconds=d["last_seen_s"],
```
rafa-be: Might be better if the `last_seen_s` would be renamed `last_seen_seconds`, for consistency.

src/scaler/scheduler/object_usage/object_tracker.py
rafa-be, line 1: The way the objects are sorted/grouped by log2(size) is so weird, and might have a `O(log(n) * n )` worst-case complexity if objects have similar but slightly different sizes.

We already use `sortedcontainers` in the project, what about using a [`SortedSet`](https://grantjenks.com/docs/sortedcontainers/sortedset.html) instead?:

```python
def objects_by_size_sort_key(obj: ObjectType) -> Tuple:
    """Sort objects by size, then by key (to allow retrieval)"""
    return obj.get_object_size(), obj.get_object_key()

self._objects_by_size = SortedSet([], key=objects_by_size_sort_key)
```

src/scaler/worker/agent/heartbeat_manager.py
```diff
@@ -103,11 +120,14 @@ async def routine(self):
                 memLimit=mem_limit,
                 queueSize=self._task_queue_size,
                 queuedTasks=queued_tasks,
-                latencyUS=self._latency_us,
+                latencyMicroseconds=self._latency_us,
```
rafa-be: Same here, inconsistent naming.

rafa-be (dismissed):

rafa-be (approved):

PR #989, opened 2026-08-31
fix: Only accept task messages from the worker that owns the task

### LINKED ISSUE #963: Scheduler accepts `TaskResult` and `TaskCancelConfirm` from any worker, so a rerouted task can run twice
`TaskResult` and `TaskCancelConfirm` are dispatched without the identity of the sender:

https://github.com/finos/opengris-scaler/blob/e2289887cb1e83f4083cafd2f9f8bd10604d4391/src/scaler/scheduler/scheduler.py#L212-L218

while `WorkerHeartbeat` and `DisconnectRequest`, a few lines below, do get `WorkerID(source)`:

https://github.com/finos/opengris-scaler/blob/e2289887cb1e83f4083cafd2f9f8bd10604d4391/src/scaler/scheduler/scheduler.py#L228-L235

PR #944 rewrites the task state machine and leaves this untouched, noting it as a separate change.

This means if a task is rerouted from an apparently-dead but not actually dead worker for instance, that the task actually executes twice. In our typical use functions we submit to the scaler are pure and free from side effects, but this isn't a requirement not a guarantee.

The fix is to pass `WorkerID(source)` into `on_task_result` and `on_task_cancel_confirm`, drop any
message whose sender is not the task's current owner, and log a warning.

magniloquency:
`TaskResult` and `TaskCancelConfirm` were dispatched without the identity of the sender, so a worker that the scheduler declared dead could still answer for a task that was already rerouted: the client was told an outcome for a task that is still running, and the capacity of the worker that now holds it was released.

The scheduler now passes `WorkerID(source)` into `on_task_result` and `on_task_cancel_confirm`, matching what `WorkerHeartbeat` and `WorkerDisconnectNotification` already do. The events those become carry the worker id, and the router drops any whose sender has been superseded, with a warning.

Only a different, valid holder proves that. An unheld task does not: `remove_worker` drops the assignments of every task a dead worker held, and it runs from the heartbeat timer without the task machine's lock, before the `WorkerDisconnected` it raises reaches the router. A report arriving in that window is from the rightful holder, so an unheld task falls through to the state machine, which refuses every worker-reported event from `inactive` and honors the rest. Dropping it instead would discard a finished task and re-run it, which is the duplicate execution the check exists to prevent.

Events the scheduler raises itself are not checked. `WorkerDisconnected` in particular arrives after `remove_worker` has already dropped the assignment, so checking it would refuse every reroute.

This stops the scheduler acting on the duplicate, it does not stop the duplicate execution: the old worker still ran the function. Preventing that needs the scheduler to tell a reconnecting worker to abandon tasks it no longer owns, which is a separate change. A worker id also cannot express multi-hop staleness, two consecutive timeouts or a reroute back to the same worker, which needs a per-dispatch epoch rather than an identity comparison.


rafa-be (dismissed):

src/scaler/scheduler/controllers/task_controller.py
```diff
@@ -300,6 +311,30 @@ async def __route(self, event: TaskEvent) -> None:
                 self._task_state_manager.remove_state_machine(event.task_id)
                 self._task_id_to_task.pop(event.task_id, None)
 
+    def __is_from_a_superseded_worker(self, event: WorkerReportedTaskEvent) -> bool:
```
rafa-be: This is valid, but an odd method name. `__is_task_owned_by_worker()` (with a reversed return value) would be more meaningful IMO.

rafa-be (dismissed):

e117649 (dismissed):

src/scaler/scheduler/controllers/task_controller.py
magniloquency: done

rafa-be (approved):

PR #990, opened 2026-09-01
fix: Recover tasks stranded in balanceCanceling (fixes #962)

### LINKED ISSUE #962: Tasks stranded in `balanceCanceling` permanently
This issue covers the following two FIXMEs:
 - https://github.com/finos/opengris-scaler/blob/61d1805c5a16acc44e8b0d1fe38b1ea24ca82f09/src/scaler/scheduler/controllers/task_controller.py#L432-L439
 - https://github.com/finos/opengris-scaler/blob/61d1805c5a16acc44e8b0d1fe38b1ea24ca82f09/src/scaler/scheduler/controllers/task_controller.py#L533-L538

PR #944 rewrites the task state machine, however in an effort to avoid too many changes in behavior
some existing buggy behavior was preserved. This issue documents one such bug.

There is a bug in the way task canceling works. If a balance task cancel event happens while a task
is running, the state is always transitioned to the balance canceling state, ignoring the
possibility that no worker holds the task, in which case no cancel is actually sent. The other case
is when a "cancel confirm not found" occurs while the task is in a state of balance canceling; the
state machine ignores it. Either way the task has no worker and no cancel in flight, so nothing that
`balanceCanceling` accepts can ever arrive: the state machine and the `Task` payload leak for the
lifetime of the scheduler and the client's future never completes.

`balanceCanceling` + "cancel confirm canceled" already handles the same situation by releasing the
worker and rescheduling the task, so both cases should probably do the same. Note that #944 pins the
current behavior in two `test_balance_cancel_*_strands_the_task` tests, which would need replacing.

magniloquency:
Resolves the two FIXMEs left by #944.

`balanceCanceling` means a cancel is in flight to a worker that holds the task, but there were two ways for it to get stuck:

- `__on_balance_task_cancel` discarded the `bool` from `__send_task_cancel_to_worker` and entered `balanceCanceling` even when no worker held the task, so no cancel was sent.
- `__on_cancel_confirm_not_found` ignored `cancelNotFound` from `balanceCanceling`.

Nothing the state accepts could then arrive. The state machine and the `Task` payload leaked for the lifetime of the scheduler, the second case also leaked the worker slot permanently, and the client's future never completed. Silent: no log line, only a growing `balanceCanceling` count in the status report.

The two paths are fixed differently, because only one of them owns the recovery.

`cancelConfirm(notFound)` from `balanceCanceling` now releases the worker and reschedules, which is what `balanceCanceling` already does for `cancelConfirm(canceled)` and for `WorkerDisconnected`. The scheduler still maps the task to that worker, the mapping is stale, and no other event is coming to clear it. The task is not terminated towards the client: a balance move is scheduler-initiated and the client asked for nothing.

The failed-send path drops the balance move instead and leaves the task `running`. It must not reschedule: the only way to reach it is `remove_worker`, which drops every task mapping of a departing worker at once and only then drains the tasks one `await` at a time, so a task can read `running` here with its `WorkerDisconnected` already queued behind the balance cancel. Placing it again would race that event into a second dispatch, running the task on two workers and leaking the queue slot of the first. The queued event does the reroute, so the strand is gone either way.

The two `test_balance_cancel_*_strands_the_task` tests are replaced: reschedule and no-capacity-queue variants for the confirm path, a dropped-and-stays-running test plus a follow-up disconnect test for the send path. The graph harness gains the `CancelConfirmNotFound (no capacity)` scenario. Three rows change in the state graph snapshot.

Note: `test_every_state_is_reachable_from_inactive` fails on `main` today because `TaskState` carries `balanceCanceled` and `workerDisconnecting`, which the scheduler never produces. Unrelated to this change and left alone.


src/scaler/scheduler/controllers/task_controller.py
```diff
@@ -431,12 +431,13 @@ async def __on_balance_task_cancel(
...
-                # terminate the task towards the client or reschedule it, so it needs its own review
-                await self.__send_task_cancel_to_worker(task_cancel, TaskState.balanceCanceling)
-                return TaskState.balanceCanceling
+                if await self.__send_task_cancel_to_worker(task_cancel, TaskState.balanceCanceling):
+                    return TaskState.balanceCanceling
+
+                # no worker holds the task, so no cancel is in flight and no confirm can ever arrive. the balance
+                # move is moot, so place the task again instead of waiting in balanceCanceling forever. there is no
+                # on_task_done here, the send failed because the worker controller already has no mapping to release
+                return await self.__acquire_and_dispatch(event.task_id)
```
rafa-be, line 440: In which case do we send a balance request to a dead worker? 

Is it because the balance manager does send all its advice to the task manager at once? Thus the the worker can die between to calls to `TaskManager.on_task_balance_cancel()`.

https://github.com/finos/opengris-scaler/blob/c8782810a2699e52ff395e8e8b1dd841f12b5674/src/scaler/scheduler/controllers/balance_controller.py#L44-L47

src/scaler/scheduler/controllers/task_controller.py
```diff
@@ -531,11 +532,11 @@ async def __on_cancel_confirm_not_found(
...
-                # FIXME: strands the task, it has no worker, no cancel in flight and no exit path. this preserves the
-                # behavior of the transition table, which never accepted this transition from balanceCanceling. the
-                # recovery is a behavior change, either terminate the task towards the client or reschedule it, so it
-                # needs its own review
-                return None
+                # the worker does not hold the task, but the scheduler still maps it there. nobody asked the client
+                # for this cancel, so the task must not be terminated: release the stale mapping and place it again,
+                # which is what a balance cancel confirmed as canceled already does
+                await self._worker_controller.on_task_done(event.task_id)
+                return await self.__acquire_and_dispatch(event.task_id)
```
rafa-be, line 539: Is there an expected case a worker returns `CancelConfirmNotFound` on a task that is in the `BalanceCancelling` state ?

It feels to me that if we get these events, there is a bug occurring, and we might want to behave slightly differently. e.g. raise `SchedulerException` or at least log the error.

src/scaler/scheduler/controllers/task_controller.py
magniloquency: Here the task id worker mappings are removed all at once, but the state machines are updated afterwards, sequentially. this creates a window where a task is running but no mapping exists. so this is a real edge case

https://github.com/finos/opengris-scaler/blob/cd5dc53f2fbe290f351008cd1e01bdf4d1bace77/src/scaler/scheduler/controllers/worker_controller.py#L210-L217

src/scaler/scheduler/controllers/task_controller.py
magniloquency: I agree, I've added a error log. I also added some logging for your previous comment. I think logging is better here over raising, the scheduler can simply re-dispatch the task in this case i think, and it's not something worth bringing the scheduler down over?

rafa-be (dismissed):

src/scaler/scheduler/controllers/task_controller.py
```diff
@@ -431,12 +431,14 @@ async def __on_balance_task_cancel(
...
-                await self.__send_task_cancel_to_worker(task_cancel, TaskState.balanceCanceling)
-                return TaskState.balanceCanceling
+                if await self.__send_task_cancel_to_worker(task_cancel, TaskState.balanceCanceling):
+                    return TaskState.balanceCanceling
+
+                # no worker holds the task, so no cancel is in flight and no confirm can ever arrive. the balance
+                # move is moot, so place the task again instead of waiting in balanceCanceling forever. there is no
+                # on_task_done here, the send failed because the worker controller already has no mapping to release
+                logger.warning(f"{event.task_id!r}: balance cancel found no worker holding the task, placing it again")
+                return await self.__acquire_and_dispatch(event.task_id)
```
e117649: I think we should have `return None` here otherwise the balance arm reroutes a task in the window where a reroute is already on its way, so the task gets placed twice

magniloquency: done

e117649 (approved):

rafa-be (approved):

PR #996, opened 2026-09-01
refactor: reorganize worker manager adapters by provisioner shape

magniloquency:
`worker_manager_adapter/` mixed three layers at its top level: the shared control plane, a proxy-worker runtime used by only three of the seven adapters, and the adapter directories themselves.

This splits it along the provisioner shapes already described in the worker manager redesign (direct / nested / proxy).

```
src/scaler/worker_manager/
├── mixins.py                     # DeclarativeWorkerProvisioner - implemented by all 7
├── runner.py                     # WorkerManagerRunner
├── capacity_coordinator.py
├── desired_concurrency.py
├── native/                       # unit is one local worker process
├── nested/                       # unit is a resource running a child manager
│   ├── child_command.py
│   ├── aws_ecs/
│   ├── orb_aws_ec2/
│   └── oci/
│       └── remote/               # ships in the image, does not import scaler
└── proxy/                        # unit is a local proxy worker fronting an external service
    ├── worker_process.py  task_manager.py  heartbeat_manager.py  mixins.py
    ├── symphony/
    ├── aws_batch/
    │   ├── tools/                # operator CLIs
    │   └── remote/
    └── oci/
        ├── tools/
        └── remote/
```

- `worker_manager_adapter/` -> `worker_manager/`. The runner and capacity coordinator are the worker manager, not adapters to it.
- Proxy runtime moves into `proxy/`, next to its only three users.
- `common.py` splits into `desired_concurrency.py` (used by all seven) and `nested/child_command.py` (only ever used to build the child command line). Drops `CapacityExceededError` and `WorkerNotFoundError`, which were unreferenced.
- Adapter modules renamed to match contents: a `heartbeat_manager.py` holding a `ProcessorStatusProvider` becomes `processor_status.py`; a `task_manager.py` holding an `ExecutionBackend` becomes `execution_backend.py`.
- Each adapter's `utility/` splits into `tools/` (operator CLIs) and `remote/` (code that ships inside the container image and does not import scaler).
- Class names drop the stale `Raw`/`HPC` words.

CLI subcommands and TOML `type =` values are unchanged. Tests mirror the new tree.

Includes one behaviour fix: the OCI provisioner computed its docker build context as `Path(__file__).parent x5 / "src"`, which resolved to `src/src`. Both provisioners now use `parents[5]`.

`nested/oci` and `proxy/oci` both export `OCIWorkerManager`; the two lazy imports share a function scope in the entry points, so they are aliased there.

## Compatibility

Unchanged: the `scaler` package exports (`Client`, `ScalerFuture`, `Serializer`, `SchedulerClusterCombo`, `Scheduler`), all six console scripts, every CLI subcommand and TOML `type =` value, all config keys and `scaler.config.section.*` paths, the Cap'n Proto protocol, and the worker/manager identity strings that show up in `scaler_top`.

The command line the nested managers inject into provisioned resources is byte-identical, so a new parent with an old child (or the reverse) still works, and container images built before this change keep working.

Breaking, for anyone reaching past the CLI:

- `from scaler.worker_manager_adapter...` imports; the path is gone.
- Renamed classes: `OCIRawWorkerManager`, `OCIHPCWorkerManager`, `OCIHPCExecutionBackend`, `AWSHPCWorkerManager`, `BatchWorkerProvisioner`, `AWSProcessorStatusProvider`, `create_oci_hpc_worker`, `AWSBatchProvisioner`, `OCIProvisioner`.
- The two provisioner commands published in the tutorials, `python -m scaler.worker_manager_adapter.{aws_hpc,oci_hpc}.utility.provisioner`, are now `python -m scaler.worker_manager.proxy.{aws_batch,oci}.tools.provision_infrastructure`. The tutorials are updated in this PR, but anyone with the old command in a runbook gets a `No module named` error.
- The `docker build -f src/scaler/.../Dockerfile.container_instance` line in the oci_raw tutorial; the Dockerfiles now live at `.../remote/Dockerfile`.

**Open question for review: do we want a compatibility shim?** A `scaler/worker_manager_adapter/__init__.py` with a module-level `__getattr__` could re-export the old module paths and class names, raise a `DeprecationWarning`, and be deleted after one release. That would cover every item above except the two `python -m` commands, which need a shim module of their own if we want them to keep working.

Arguments against: the project has no CHANGELOG and no deprecation machinery anywhere in `src/`, so there is no convention to extend; `worker_manager_adapter` was never in `scaler.__all__`; and the Python breakage announces itself loudly at import time rather than failing silently. Release notes may be enough.

Arguments for: the two provisioner commands are copy-pasted out of the published tutorials, so they are public in practice whatever the export list says.

Happy to add the shim to this PR if reviewers want it.


e117649:
Looks good, some missed things (all small), I'll let you know later if we need a shim (currently leaning no).

1. `src/scaler/worker_manager/nested/oci/worker_manager.py:240` — rename `OCIWorkerManager` to `OCIInstancesWorkerManager`, and `:36` `OCIWorkerProvisioner` to `OCIInstancesWorkerProvisioner` (call site `:245`).
2. `src/scaler/worker_manager/proxy/oci/worker_manager.py:99` — rename `OCIWorkerManager` to `OCIJobsWorkerManager`, and `:21` `OCIWorkerProvisioner` to `OCIJobsWorkerProvisioner` (call site `:115`); update `proxy/oci/__init__.py:8,25,29,30`.

   Two classes sharing one name is the only new name this PR invents, and it is the one thing here worth getting right the first time. `Instances` and `Jobs` say what each provisions: Container Instances that host workers, versus one Container Instance per task. `NestedOCIWorkerManager` / `ProxyOCIWorkerManager` mirroring the packages would work too.
3. `src/scaler/entry_points/scaler.py:81-87` and `src/scaler/entry_points/worker_manager.py:122-128` — drop the `as NestedOCIWorkerManager` / `as ProxyOCIWorkerManager` aliases, which exist only because of that collision.
4. `tests/entry_points/test_worker_manager.py:745` and `:1072` — the two `patch(...)` targets follow the rename.
5. `docs/source/tutorials/worker_managers/aws_hpc_batch.rst:95` — still says the image is built from ``Dockerfile.batch``; it is now ``proxy/aws_batch/remote/Dockerfile``.
6. `docs/source/tutorials/worker_managers/oci_raw.rst:141` — the documented build context is wrong. The Dockerfile COPYs `scaler/...`, so from the repository root the command fails with `COPY failed: file not found in build context: stat scaler/worker_manager/nested/oci/remote/entrypoint.sh: file does not exist`. With `src` as the context it builds. The command still runs from the repository root, only the context argument changes:

   ```bash
   docker build \
       -f src/scaler/worker_manager/nested/oci/remote/Dockerfile \
       -t us-ashburn-1.ocir.io/<namespace>/<repo>:latest src

   This was already broken before the move, but the PR edits those exact lines.

7. tests/worker_manager/proxy/aws_batch/test_worker_manager.py:22 — TestBatchWorkerProvisionerConcurrencyConversion did not follow BatchWorkerProvisioner to AWSBatchWorkerProvisioner.
8. src/scaler/worker_manager/__init__.py is empty while native, nested and proxy each carry their taxonomy docstring. The three-shape summary belongs where a reader lands first.
9. .scaler_aws_hpc.env keeps the old word in the new tree (provision_infrastructure.py:178,786 plus four doc lines), next to a TOML and a harness that were both renamed. It is a generated file, so renaming costs a regenerate.

PR #998, opened 2026-09-04
feat: Add Kubernetes worker manager

### LINKED ISSUE #958: Add a Kubernetes worker manager for Scaler
Allow Scaler to deploy workers on Kubernetes.

kincl:
Add a worker manager that provisions Scaler workers as bare Kubernetes Pods, owning the pod lifecycle directly (no Deployment or Job controller). A background watch detects externally deleted pods and signals the capacity coordinator to create replacements.

Other changes:
- Add CapacityCoordinator.request_reconcile() for external state changes
- Dockerfile: bake scaler into the image; make PYTHON_VERSION and PYTHON_REQUIREMENTS env vars optional in entrypoint.sh
- Add KinD integration test CI job to GitHub Actions workflow behind SCALER_KIND_TESTS=1 gate, with kind-test.sh script for cluster setup

Closes #958


.github/workflows/build-and-test.yml
```diff
@@ -136,3 +136,85 @@ jobs:
...
+          uv pip install --system build
+          python -m build --wheel
+          wheel_file=$(ls dist/*.whl | head -n 1)
+          uv pip install --system "${wheel_file}[kubernetes]"
+
+      - name: Install KinD
+        run: |
+          mkdir -p "${RUNNER_TEMP}/bin"
+          curl -Lo "${RUNNER_TEMP}/bin/kind" \
+            https://kind.sigs.k8s.io/dl/v0.27.0/kind-linux-amd64
```
magniloquency, line 186: Can we lift this version number into an env var?

docker/Dockerfile
```diff
@@ -1,12 +1,35 @@
...
+    capnproto capnproto-dev \
+    libuv-dev openssl-dev
+
 COPY --from=uv /uv /usr/local/bin/uv
 
+ENV UV_CACHE_DIR="/tmp/uv-cache" \
+    UV_PYTHON_INSTALL_DIR="/opt/python" \
+    PATH="/opt/opengris-scaler/bin:${PATH}"
+
+RUN uv python install 3.12 && \
```
magniloquency, line 15: Why not install a newer version?

magniloquency, line 15: I think we should make a decision about how python is installed. We install 3.12 here, but the entrypoint.sh script will install python based on an environment variable. We should decide if we're baking the Python version into the image or not and adjust accordingly

docker/Dockerfile
```diff
@@ -1,12 +1,35 @@
...
+RUN uv venv --python 3.12 /opt/opengris-scaler && \
+    uv pip install --no-cache -q --python /opt/opengris-scaler /tmp/scaler-src && \
+    rm -rf /tmp/scaler-src /tmp/uv-cache
+
 COPY docker/entrypoint.sh /app/entrypoint.sh
 RUN chmod +x /app/entrypoint.sh
 
+RUN adduser -D -u 1001 -G root scaler && \
+    mkdir -p /app && \
+    chmod g+rwx /app /opt/opengris-scaler
```
magniloquency, line 29: Maybe this user should own /opt/python and /opt/opengris-scaler?

docker/Dockerfile
```diff
@@ -1,12 +1,35 @@
 FROM ghcr.io/astral-sh/uv:latest AS uv
 FROM alpine:3.23
 
+RUN apk add --no-cache \
+    ca-certificates cmake gcc g++ make pkgconf \
+    capnproto capnproto-dev \
+    libuv-dev openssl-dev
```
magniloquency, line 7: We could probably run a two-stage build so that the final image doesn't need these installed. The caveat is that we're installing deps in the entrypoint. On the other hand, we could bake the deps into the image. That would help with startup times

docs/source/tutorials/worker_managers/k8s_raw.rst
```diff
@@ -0,0 +1,516 @@
...
+
+     pip install opengris-scaler[kubernetes]
+
+* RBAC permissions granting the service account ``get``, ``list``, ``watch``,
+  ``create``, and ``delete`` on ``pods`` in the target namespace.
+  See `RBAC Example`_ below for a ready-to-apply manifest.
+
+* A container image pushed to a registry reachable from the cluster.  The image must
+  include ``uv`` or ``pip`` so that Python requirements can be installed at container
+  start-up (the same mechanism used by other raw worker managers).
```
magniloquency, line 30: I think we can be specific, we provide the images so they should just use those. There's extra steps involved like installing build tools for the deps, rather than explaining how to build your own image, we'll just provide ours and users can reverse-engineer it if they really want to

docs/source/tutorials/worker_managers/k8s_raw.rst
```diff
@@ -0,0 +1,516 @@
...
+     docker push myregistry.example.com/scaler:latest
+
+Quick Start
+-----------
+
+Create a virtual environment and install Scaler with the Kubernetes extra:
+
+.. code-block:: bash
+
+   python -m venv .venv
```
magniloquency, line 50: we use `uv` here. you can check the other docs to see how we usually advise installing scaler and python

docs/source/tutorials/worker_managers/k8s_raw.rst
```diff
@@ -0,0 +1,516 @@
...
+         object_storage_address = "tcp://scheduler-service:8517"
+
+         [[worker_manager]]
+         type = "k8s_raw"
+         scheduler_address = "tcp://scheduler-service:8516"
+         worker_manager_id = "wm-k8s-01"
+         max_task_concurrency = 80
+
+         namespace = "scaler"
+         pod_image = "myregistry.example.com/scaler:latest"
```
magniloquency, line 84: what would it take to make this a real registry? can we publish our images on ghcr?

docs/source/tutorials/worker_managers/k8s_raw.rst
```diff
@@ -0,0 +1,516 @@
...
+   * - Field
+     - Type
+     - Default
+     - Description
+   * - ``workers_per_pod``
+     - ``int``
+     - ``1``
+     - Number of Scaler worker processes launched inside each Pod.
+       Acts as the divisor when converting ``max_task_concurrency`` to a pod count.
+       Must be >= 1.
```
magniloquency, line 226: I think we should consider bumping this. each pod / native wm has some overhead, so it's better to have #workers > #wms

docs/source/tutorials/worker_managers/k8s_raw.rst
```diff
@@ -0,0 +1,516 @@
...
+
+   [worker_manager.resource_requests]
+   cpu = "4"
+   memory = "16Gi"
+
+   [worker_manager.resource_limits]
+   memory = "32Gi"
+
+   # pod_template for everything else:
+   pod_template = """
```
magniloquency, line 324: It may be nice to allow this field to also be a filename. we already do this for the `requirements_txt` field so you can refer to it

docs/source/tutorials/worker_managers/k8s_raw.rst
```diff
@@ -0,0 +1,516 @@
...
+       securityContext:
+         allowPrivilegeEscalation: false
+         capabilities:
+           drop: ["ALL"]
+   """
+
+.. warning::
+   If you specify ``env`` in the YAML template's container spec, it **replaces**
+   Scaler's injected env vars (``COMMAND``, ``PYTHON_REQUIREMENTS``,
+   ``PYTHON_VERSION``) entirely. You must re-include them if you need them.
```
magniloquency, line 352: Is it possible for us to merge them?

scripts/kind-test.sh
magniloquency, line 1: maybe this should go in tests/?

src/scaler/worker_manager_adapter/kubernetes/worker_manager.py
```diff
@@ -0,0 +1,301 @@
...
+    Merge rules: nested dicts are merged recursively; lists replace the base list
+    entirely (no element-level merge); scalars use the override value.
+    """
+    result: Dict[str, Any] = dict(base)
+    for key, val in override.items():
+        if key in result and isinstance(result[key], dict) and isinstance(val, dict):
+            result[key] = _deep_merge(result[key], val)
+        else:
+            result[key] = val
+    return result
```
magniloquency, line 44: Put in a util file?

src/scaler/worker_manager_adapter/kubernetes/worker_manager.py
```diff
@@ -0,0 +1,301 @@
...
+
+                # Config fields override the template.
+                if config.node_selector:
+                    pod_dict["spec"]["nodeSelector"] = config.node_selector
+
+                if config.service_account_name:
+                    pod_dict["spec"]["serviceAccountName"] = config.service_account_name
+
+                if config.image_pull_policy:
+                    pod_dict["spec"]["containers"][0]["imagePullPolicy"] = config.image_pull_policy
```
magniloquency, line 228: Is it possible any of these (or adjacent lines) could throw a key error?

src/scaler/worker_manager_adapter/kubernetes/worker_manager.py
```diff
@@ -0,0 +1,301 @@
...
+                    if config.resource_limits:
+                        resources["limits"] = config.resource_limits
+
+                # Invariant: restartPolicy must always be Never.
+                pod_dict["spec"]["restartPolicy"] = "Never"
+
+                await loop.run_in_executor(
+                    None, functools.partial(self._core_v1.create_namespaced_pod, config.namespace, pod_dict)
+                )
+                self._pods.append(pod_name)
```
magniloquency, line 243: Can this race? It seems like we add the pod to the list before it's necessarily launched, so what if we needed to scale down quickly and we tried to delete this pod but it doesn't exit?

magniloquency (changes requested):

.github/workflows/build-and-test.yml
kincl: Moved it to env for the job

docs/source/tutorials/worker_managers/k8s_raw.rst
kincl: Great catch, fixed this in the docs to use uv

src/scaler/worker_manager_adapter/kubernetes/worker_manager.py
kincl: Moved to `src/scaler/utility/dict_utils.py` and moved the unit tests over there as well

docker/Dockerfile
kincl: Done!

docs/source/tutorials/worker_managers/k8s_raw.rst
kincl: Done

docs/source/tutorials/worker_managers/k8s_raw.rst
kincl: Yes! Done, it now is the same behavior as requirements_txt config

docker/Dockerfile
kincl: I made this 3.14, but more questions need to be answered about how opinionated we want to be about python version in the image

kincl: I tried to do both but I agree we should just make a decision here and stick with it. We can have two container images one thats "dynamic" and lets you do on-the-fly provisioning of scaler and one that is "pre-baked"? Or I think it is reasonable to choose some defaults and just ship an image that works at least with respect to the Python version. We need to keep requirements.txt but should we also have some way to specify dev packages that need to be installed in container before pip runs?

scripts/kind-test.sh
kincl: Do you mean convert this into a python test instead of a bash script?

src/scaler/worker_manager_adapter/kubernetes/worker_manager.py
kincl: Great call out, fixed this to be more defensive

kincl:
Switched to not squashing commits for this PR to make reviews easier

docker/Dockerfile
kincl: I decided to take a more minimal change approach to the existing container image. It might make more sense to evaluate these decisions in a future PR. For this change I moved to building the whl into an image and using that image as the base for the final image. This way we can use existing requirements.txt mechanism for installing the code we are testing in CI.

docs/source/tutorials/worker_managers/k8s_raw.rst
kincl: Fixed, removed that from docs.

docs/source/tutorials/worker_managers/k8s_raw.rst
kincl: We should be able to add a image publishing pipeline for the image, should we tackle that in a separate PR or this one? We would need access to publish packages to the finos org.

src/scaler/worker_manager_adapter/kubernetes/worker_manager.py
kincl: We should be good here, since `start_units()` is called on the event loop thread. I did fix a different race on the watch thread that was mutating `self._pods` to make sure it runs on the event thread

docker/Dockerfile
kincl: Opted to revert these changes to the image for now

docs/source/tutorials/worker_managers/k8s_raw.rst
kincl: Fixed, good call!

kincl:
@magniloquency no rush but ptal

PR #1001, opened 2026-09-10
fix: a client inside a worker reaches object storage the way its worker does

### LINKED ISSUE #1003: Nested clients should get object storage address from their worker
We currently do this for the scheduler, but not for object storage

e117649:
closes https://github.com/finos/opengris-scaler/issues/1003


src/scaler/client/client.py
```diff
@@ -83,7 +83,8 @@ def __init__(
         :param stream_output: If True, stdout/stderr will be streamed to client during task execution
         :type stream_output: bool
         :param object_storage_address: Override object storage address (e.g., for Docker/Kubernetes port mapping).
-                                       If None, will use address received from scheduler.
+                                       If None, a client inside a worker uses its worker's address. Any
+                                       other client uses the address the scheduler advertises.
```
rafa-be: I'm afraid this might cause some issues if the `scheduler_address` is different from the worker's `scheduler_address`, e.g. if a user want to connect to a 2nd cluster from the worker itself (we did that when we had one cluster for Pargraph, and another for Parfun).

Instead, we could use the worker's OSS address only if that address *and* the scheduler address are not specified. This would behave like this:

```python
with Client(scheduler_address=..., object_storage_address=...) as c: # uses the user-specified OSS address
    ...

with Client(scheduler_address=...) as c: # uses the OSS address as advertised by the scheduler
    ...

with Client() as c: # uses the OSS address from the worker
    ...
```

rafa-be (approved):

magniloquency (approved):

PR #1002, opened 2026-09-10
fix: two cancel races that crash the proxy worker manager

magniloquency:
Two ways a task cancel kills a proxy worker manager's worker process, taking every other task on it. Both are reproducible against IBM Spectrum Symphony 7.3.2, and both predate the worker manager reorganisation in #996.

**`on_task_new` KeyError.** `process_task` left a picked-up task in `_queued_task_ids`, so an in-flight task looked both queued and processing. A force cancel took the queued branch too and popped `_task_id_to_task` out from under the pending future; `resolve_tasks` then found no task and continued without dropping the id from `_acquiring_task_ids` or releasing the semaphore. The next task to arrive looked that id up and raised. The missing-task path also leaked one unit of concurrency for the life of the worker even where it did not crash.

**`on_cancel_task` KeyError.** A task counts as processing from before `execute()` returns, so a force cancel landing during that await found no future recorded yet.

Cleanup now runs through `_release_task` in a `finally`, so no exit path can strand an id or leak a permit, and both lookups tolerate the gap they cannot rule out. The body of the resolve loop moves into `_report_task_outcome` and `_send_task_result` unchanged.

The eight new tests all fail against the current code, four of them with the exact `KeyError`.

### DIFF
--- a/src/scaler/worker_manager_adapter/task_manager.py
+++ b/src/scaler/worker_manager_adapter/task_manager.py
@@ -82,7 +82,13 @@ class TaskManager(Looper, TaskManagerMixin):
 
         if self._executor_semaphore.locked():
             for acquired_task_id in self._acquiring_task_ids:
-                acquired_task = self._task_id_to_task[acquired_task_id]
+                acquired_task = self._task_id_to_task.get(acquired_task_id)
+                if acquired_task is None:
+                    # Only a bookkeeping bug puts an id here without its task. Skipping it costs one
+                    # priority comparison, where raising would take down the worker and every task on it.
+                    logger.error(f"Acquired task is not in the worker queue: task_id={acquired_task_id.hex()}")
+                    continue
+
                 acquired_task_priority = self._get_task_priority(acquired_task)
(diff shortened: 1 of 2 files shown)


rafa-be:
Isn't this covering up a more complex concurrency bug caused by `_acquiring_task_ids`, `_processing_task_ids` and `_task_id_to_future` not being protected by some kind of locking mechanism, caused by `_execution_backend.execute()` being an asyncio call? 

That's is, `_acquiring_task_ids` and `_processing_task_ids` might hold the task ID while it's not yet inside `_task_id_to_future` as [the call to `_execution_backend.execute` is blocking](https://github.com/magniloquency/scaler/blob/9e0e5e5ef34e0a339de828b7e323fc0abc8c295e/src/scaler/worker_manager_adapter/task_manager.py#L225-L228).

I'm also wondering if the change in `on_task_new()` will create some undesirable behavior for which a lower priority task could be prioritized over an higher priority one:

```python
                    # Stop instead, so the task queues the way it would have if the
                    # id had belonged to a task of at least its priority
```

